PR #108, opened 2020-09-04
Break down resource usage and request throughput in pod Live View by container name.

nserrino:
We want to be able to show the proportional contribution of each container on a pod's resource usage and request throughput.  This change adds that support to do so in the pod Live View.


nserrino:
Screen Shot 2020-09-03 at 9 10 17 PM
Screen Shot 2020-09-03 at 9 10 23 PM

zasgar (approved):
Looks good! Thanks.

PR #113, opened 2020-09-12
Fix denominators in process_stats scripts and remove pod_stats

philkuz:
Erroneously displayed windowed data that were supposed to be rates. Divided by the window size to fix the value.

Also removed pod_stats because it's been superseded by pod.


nserrino (approved):
Thanks for fixing this!

ishanmkh:
@philkuz  merging as @nserrino has reviewed it. i tested it as well.

ishanmkh (approved):
👍

PR #114, opened 2020-09-15
Service graphs fallback on nslookup

philkuz:
A service graph change that relies on nslookup as a fallback if an IP can't be resolved via the Pixie Metadata. This usually means the IP does not correspond to a k8s object and nslookup can cover the case where the IP is external.


zasgar (changes requested):
nslookup is not very fast. I am concerned this is going to cause performance issues for the script. We might want to add support for predicated execution before we make this change.

philkuz:
Got it

PR #120, opened 2020-09-18
Rename connections_by_ip to net_flow_graph and move options to arguments

philkuz:
Rename script
1. Add arguments for ip and start_time
2. Remove the namespace commented out line and instead pass in a metadata string as the metadata.


zasgar (approved):

htroisi:
@philkuz - Is there a reason we don't want to add grouping_entity as an argument as well? (with a pre-populated argument menu that includes pod, namespace, deployment)

Issue #133, opened 2020-09-28
Pod Live View is missing expected column (container) for CPU and memory widgets


nserrino:
**Describe the bug**
The vis.json expects a column called 'container' but it doesn't exist for the resource_timeseries table, causing the container to be omitted from the timeseries chart label.

**To Reproduce**
1. Go to the px/pod view and look at the CPU chart.

**Expected behavior**
The container name should show up in the legend.

htroisi:
This bug was fixed by PR #134.

PR #134, opened 2020-09-28
Add container to px/pod resource_timeseries output table

nserrino:
Issue #133


philkuz (approved):
Only thing is remove "Issue #133" from title, put in desc.

zasgar (approved):

PR #137, opened 2020-09-30
Adding memory to px/namespace stats

philkuz:
Needed this while we were debugging, might as well add it in.
image

### DIFF
--- a/pxl_scripts/px/namespace/namespace.pxl
+++ b/pxl_scripts/px/namespace/namespace.pxl
@@ -33,8 +33,8 @@ def pods_for_namespace(start_time: str, namespace: px.Namespace):
     df = df[df.ctx['namespace'] == namespace]
     df.pod = df.ctx['pod_name']
     df = df.groupby(['pod']).agg(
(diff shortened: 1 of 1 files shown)


pxl_scripts/px/namespace/namespace.pxl
```diff
@@ -32,7 +32,11 @@ def pods_for_namespace(start_time: str, namespace: px.Namespace):
     df = px.DataFrame(table='process_stats', start_time=start_time)
     df = df[df.ctx['namespace'] == namespace]
     df.pod = df.ctx['pod_name']
-    df = df.groupby(['pod']).agg()
+    df = df.groupby(['pod']).agg(
+        rss_bytes=('rss_bytes', px.mean),
+        vsize_bytes=('vsize_bytes', px.mean),
```
zasgar: Since we show bytes now, we don't need to actually have it in the name.

philkuz: image
Updated

zasgar (approved):

PR #145, opened 2020-10-06
Add valid values to px/node and px/net_flow_graph

philkuz:
New UI changes allow users to restrict the set of valid values for vis spec variables. This updates the two scripts that need this restriction to reduce user error.


ishanmkh (approved):
👍

PR #154, opened 2020-10-14
Limit px/service execution to only the nodes containing a service

philkuz:
px/service was an expensive operation for a few reasons. One we had an unnecessary number of aggregates in the summary tables + an unnecessary join. I combined this graph into a single aggregate, although I had to drop a column that contained average rps. Need a way to convert the `start_time` param into a duration, which I've filed as an internal Vizier issue. 

On top of that, outbound_requests needs reworking. Currently we aggregated all of the services and all of the ips those services talk to in a cluster. Big cluster meant huge exec time. Removed that until we have a more efficient way to avoid this. We are still discussing solutions.

### DIFF
--- a/pxl_scripts/px/service/service.pxl
+++ b/pxl_scripts/px/service/service.pxl
@@ -96,38 +96,18 @@ def let_summary_helper(start_time: str):
     '''
     df = let_helper(start_time)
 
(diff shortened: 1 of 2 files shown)


ishanmkh:
@zasgar  -- assigning to you to verify optimization

philkuz:
This got merged in w/ #155.

PR #155, opened 2020-10-14
Remove latency filters in scripts

philkuz:
These filters hide results. We had them originally because we had erroneous latencies picked up by the collector. Those bugs were fixed so the hacks are unnecessary and hide real high latency scenarios.


ishanmkh (approved):
👍

PR #157, opened 2020-10-15
Update px/tcp_drops edgeWeight/Color/Hover column

aimichelle:
tcp_drops was modified from tcp_retransmits, and the edge weight/color/hover column wasn't updated to use the drops column.


ishanmkh (approved):
👍

PR #158, opened 2020-10-16
[px/pod] add widget to display node the pod runs on

htroisi:
image

### DIFF
--- a/pxl_scripts/px/pod/pod.pxl
+++ b/pxl_scripts/px/pod/pod.pxl
@@ -50,7 +50,7 @@ def node(start_time: str, pod: px.Pod):
     df = px.DataFrame(table='process_stats', start_time=start_time)
     df = df[df.ctx['pod'] == pod]
     df.name = df.ctx['node_name']
-    df = df.groupby(['name']).agg()
+    df = df.groupby('name').agg()
     return df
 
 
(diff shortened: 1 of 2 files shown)


pxl_scripts/px/pod/pod.pxl
```diff
@@ -40,6 +39,21 @@ def containers(start_time: str, pod: px.Pod):
...
+
+    Args:
+    @start_time: The timestamp of data to start at.
+    @pod: The name of the pod to filter on.
+
+    '''
+    df = px.DataFrame(table='process_stats', start_time=start_time)
+    df = df[df.ctx['pod'] == pod]
+    df.name = df.ctx['node_name']
+    df = df.groupby(['name']).agg()
```
nserrino, line 53: nit: you can omit the bracket and just say 'name' when the list only has one element

nserrino (approved):

philkuz:
Change the git title to px/pod rather than px/node.


Also I would also add info abt the service that runs the pod. Maybe something like Pod Metadata?

philkuz (changes requested):
rename pr and add more metadata to the table. We can also discuss over slack

htroisi:
image

zasgar:
I think this is ok for now, but we should probably have a custom widget type to make the formatting better.

zasgar (approved):

PR #190, opened 2020-12-19
Update Travis Build to enforce README.md updates if px/* files change. 

htroisi:
Goal: maintain an up-to-date list of our PxL scripts + descriptions in the `pxl_scripts/README.md` file.

 `pxl_scripts/update_readme.py` automatically generates the `pxl_scripts/README.md` file from the PxL script foldernames + yaml file descriptions. 

If a PR includes includes any `px/` files, the build checks that the appropriate updates have been made to the README.md file. If not, it prints an error directing the user to run the `update_readme.py` script and check in the results.


pxl_scripts/update_readme.py
```diff
@@ -1,12 +1,16 @@
+# Run using python3.
+
 import os
 import yaml
 
+if os.getcwd().split("/").pop() == "pixie":
+	os.chdir("./pxl_scripts")
```
zasgar, line 8: For this to be a little cleaner you can probably have it take in a path to look for the scripts as an argument to the python script.

pxl_scripts/update_readme.py
```diff
@@ -1,12 +1,16 @@
...
 
 def script_list_format(folder, script, desc):
 	github_url = "https://github.com/pixie-labs/pixie/tree/main/pxl_scripts"
 	return f"- {folder}/[{script}]({github_url}/{folder}/{script}): {desc}"
 
-def get_script_list(directory="px"):
+def get_script_list(directory="./px"):
```
zasgar, line 13: similarly, the directory to scan can probably be a path as well. 

Alternatively, you can use a glob to find all the **/manifest.yaml files and then use that to determine what the active scripts are.

zasgar (changes requested):
To be consistent with the rest of the repo commands, you can probably add an update command to the Makefile.

pxl_scripts/update_readme.py
htroisi: Rather than take in an argument, I look for the root directory and append the path to the scripts.

zasgar: I know, it’s just a bit weird to have this hard coded in the script when it’s not necessary.

zasgar (approved):

PR #200, opened 2021-02-01
'Slack Alerts using the Pixie API' tutorial code

htroisi:


demos/slack-alert-app/README.md
```diff
@@ -0,0 +1,5 @@
+# Slack Alerts using the Pixie API
+
+Code for the ["Slack Alerts using the Pixie API"](https://docs.pixielabs.ai/tutorials/slackbot-alert) tutorial.
+
+If you have any questions, please reach out on our [Pixie Community Slack](https://slackin.withpixie.ai/) or via a GitHub issue.
```
philkuz, line 5: `file a Github issue`

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,116 @@
...
+        service_stats_msg.append(formatMessage(row["service"],
+                                               row["total_requests"],
+                                               row["error_count"]))
+
+    return "\n\n".join(service_stats_msg)
+
+
+def formatMessage(service, request_count, error_count):
+    # Format Pixie API table row data.
+    print(service, " ---> ", error_count,
```
philkuz, line 67: use logging here if you're gonna use it package wide

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,116 @@
...
+    total_requests=('http_resp_status', px.count)
+)
+
+px.display(df, "status")
+"""
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+def getPixieData(api_key, cluster_id):
```
philkuz, line 39: you'll want to switch over to snake case instead of camel case. Our python standard is to use snake case. 

`getPixieData` would become `get_pixie_data`

demos/slack-alert-app/slackbot.py
philkuz: actually seems unnecessary here. If you need to log this, log it in whatever consumes `formatMessage`

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,116 @@
+import time
+import logging
+import os
```
philkuz, line 3: you should add a requirements.txt file that includes 
`schedule`, `slack`, and `pxapi`.

```
pip freeze
```
and copy the versions for slack, schedule, and pxapi.

philkuz (changes requested):
Address some of the changes

demos/slack-alert-app/slackbot.py
htroisi: This print line wasn't supposed to still be included.

philkuz (approved):

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,128 @@
...
+import schedule
+import pxapi
+from slack.web.client import WebClient
+from slack.errors import SlackApiError
+
+
+# Pixie PxL script.
```
nserrino, line 24: might be nice to say what this script is doing in the comment

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,128 @@
...
+)
+
+px.display(df, "status")
+"""
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+def get_pixie_data(cluster_conn):
+    # Get data from the Pixie API.
```
nserrino, line 53: i would move this line above the function definition

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,128 @@
...
+
+    except SlackApiError as e:
+        logging.error('Request to Slack API Failed: {}.'.format(e.response.status_code))
+        logging.error(e.response)
+
+
+if __name__ == "__main__":
+
+    # Get Pixie API key.
+    PIXIE_API_KEY = os.environ['PIXIE_API_KEY']
```
nserrino, line 94: might be good to put a check to see if the api key is defined in the environment

if 'PIXIE_API_KEY' not in os.environ:
     < some nice error msg >
     
same for the other env vars

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # Create a Pixie client.
+    logging.debug("Authorizing Pixie client.")
+    px_client = pxapi.Client(token=PIXIE_API_KEY)
+
+    # Get Pixie cluster ID.
+    if "PIXIE_CLUSTER_ID" not in os.environ:
+        logging.error("Missing `PIXIE_CLUSTER_ID` environment variable.")
+    PIXIE_CLUSTER_ID = os.environ['PIXIE_CLUSTER_ID']
```
zasgar, line 108: we should hoist all these config variables out to the top of the file. If the intention is not for these to be config constants don't make them all caps.

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,137 @@
...
+        logging.info(f"Sending {msg!r} to {channel!r}")
+        slack_client.chat_postMessage(channel=channel, text=msg)
+
+    except SlackApiError as e:
+        logging.error('Request to Slack API Failed: {}.'.format(e.response.status_code))
+        logging.error(e.response)
+
+
+if __name__ == "__main__":
+
```
zasgar, line 95: have a main function that is called here. Usually the if statement is used to select what happens when this package is imported. Having the entire main function here is not a good idea.

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,137 @@
...
+    # Create a Pixie client.
+    logging.debug("Authorizing Pixie client.")
+    px_client = pxapi.Client(token=PIXIE_API_KEY)
+
+    # Get Pixie cluster ID.
+    if "PIXIE_CLUSTER_ID" not in os.environ:
+        logging.error("Missing `PIXIE_CLUSTER_ID` environment variable.")
+    PIXIE_CLUSTER_ID = os.environ['PIXIE_CLUSTER_ID']
+
+    # Connect to cluster.
```
zasgar, line 110: I usually find comments that just refer to the code not that useful. For example:
    # Connect to cluster.

   connect_to_cluster(...)

It's better to have more descriptive comments than define one for each line. 


For example, 

\# The slackbot needs the following configs: cluser_id, slack_bot_token ...
\# We allow these to be specified using the environment variables...

Then the code to fetch all the tokens.

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,137 @@
...
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+# Get data from the Pixie API.
+def get_pixie_data(cluster_conn):
+
+    # Execute the PxL script.
+    script = cluster_conn.prepare_script(PXL_SCRIPT)
+    logging.debug("Pixie cluster executed script.")
```
zasgar, line 59: I don't think the script has been executed yet.

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,137 @@
...
+                                                      SLACK_CHANNEL,
+                                                      cluster_conn))
+
+    logging.info("Message scheduled for %s Slack channel.", SLACK_CHANNEL)
+
+    while True:
+        schedule.run_pending()
+
+        # Sleep for 5 seconds between checks on the scheduler.
+        time.sleep(5)
```
zasgar, line 137: I'm a bit confused by this. This sleeps every 5 seconds but the messages only send every 5 mins right?

demos/slack-alert-app/slackbot.py
```diff
@@ -0,0 +1,137 @@
...
+# Group HTTP events by service, counting errors and total HTTP events.
+df = df.groupby(['service']).agg(
+    error_count=('error', px.sum),
+    total_requests=('http_resp_status', px.count)
+)
+
+px.display(df, "status")
+"""
+
+logging.basicConfig(level=logging.DEBUG)
```
zasgar, line 51: I'm not sure you want to set the default log level to debug.

zasgar (changes requested):

demos/slack-alert-app/slackbot.py
htroisi: schedule() is used to send the http errors message every 5 minutes, no matter how long it takes to query and post the data. I believe it's standard to use a few second sleep time, because otherwise the message sending could be 5 min +/- sleep_time.

zasgar (approved):

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,166 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"context"
```
nserrino, line 20: nit, can you alphabetize these within 2 categories. 1. built in go packages like ctx, 2. github or go.withpixie.dev ones

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,166 @@
...
+	"go.withpixie.dev/pixie/src/api/go/pxapi/types"
+)
+
+var (
+	// Slack channel for Slackbot to post in.
+	// Slack App must be a member of this channel.
+	slackChannel = "#pixie-alerts"
+
+	// This PxL script ouputs a table of the HTTP total requests count and
+	// HTTP error (>4xxx) count for each service in the `px-sock-shop` namespace.
```
nserrino, line 39: You might want to add to your README that you want to deploy the sock shop demo

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,166 @@
...
+		panic(err)
+	}
+	vz, err := pixieClient.NewVizierClient(ctx, pixieClusterID)
+	if err != nil {
+		panic(err)
+	}
+
+	slackClient := slack.New(slackToken)
+
+	// Send tick every 5 minutes.
```
nserrino, line 97: "Run the script every 5 min"?

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,166 @@
...
+	ticker := time.NewTicker(5*time.Minute)
+	for {
+		tm := &tableMux{tables: make(map[string]*tableCollector)}
+		log.Println("Executing PxL script.")
+		resultSet, err := vz.ExecuteScript(ctx, pxlScript, tm)
+		if err != nil && err != io.EOF {
```
nserrino, line 103: do we send io.EOF in our API? Will resultSet be nil if err is set to io.EOF? If resultSet is nil when err is io.EOF, then this check for it won't matter because we will hit a fatal error on line 107 when we call resultSet.Stream()

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,166 @@
...
+			panic(err)
+		}
+
+		log.Println("Stream PxL script results.")
+		if err := resultSet.Stream(); err != nil {
+			fmt.Printf("Got error : %+v, while streaming.\n", err)
```
nserrino, line 109: nit: remove space before :

nserrino: You switch between log and fmt.Printf, is there a reason for that or should we just pick one?

nserrino, line 109: one more thing: If we want to proceed even when we hit these errors rather than panicking, you may want to put a comment explaining why we are just printing the errors rather than bailing

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,166 @@
...
+		// Get slack message constructed from table data.
+		table := tm.GetTable("http_table").GetTableDataSync()
+		log.Println("Sending slack message.")
+		_, _, err = slackClient.PostMessage(slackChannel, slack.MsgOptionText(table, false), slack.MsgOptionAsUser(true))
+		if err != nil {
+			log.Println("Error sending to slack: " + err.Error())
+		}
+
+		// wait for next tick
+		<-ticker.C
```
nserrino, line 121: Have you checked if ctrl+c or ctrl+z will successfully exist this program? If not you may need to listen to those signals (instructions like https://golangcode.com/handle-ctrl-c-exit-in-terminal/) and do something like this (adapted from https://stackoverflow.com/questions/16466320/is-there-a-way-to-do-repetitive-tasks-at-intervals)

```
quitCh := make(chan struct{})
setupSignalHandler(quitCh)

executeTimer := time.NewTicker(5 * time.Second)
go func() {
    for {
       select {
        case <- ticker.C:
            handleTable(....)
        case <- quitCh:
            ticker.Stop()
            return
        }
    }
 }()
```

then have the ctrl+c/ctrl+z signal listener be responsible for calling the quit channel

htroisi: Yes, ctrl+c successfully exists.

nserrino: okay ignore the above remark then

demos/slack-alert-app/go/slackbot.go
htroisi: // Using a ticker (rather than a for loop + sleep call) means that the slackbot
// will message every 5 minutes, regardless of how long it takes to call the 
// Pixie API (as long as it takes less than 5 minutes).

nserrino: Maybe just 
// Send result every 5 min 
?

demos/slack-alert-app/go/slackbot.go
vihangm: Don't we expect an EOF when streaming not executing?

Anyway EOF is treated as not an error:
https://github.com/pixie-labs/pixie/blob/main/src/api/go/pxapi/results.go#L115-L117

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,171 @@
...
+
+# Group HTTP events by service, counting errors and total HTTP events.
+df = df.groupby(['service']).agg(
+    error_count=('error', px.sum),
+    total_requests=('http_resp_status', px.count)
+)
+
+px.display(df, "http_table")
+`
+)
```
vihangm, line 65: This script is reused in multiple files and inlining it here is a bit ugly.
How about we write it to a separate file and read it in?

Then we could share it between the python and go examples.

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (t *tableCollector) HandleInit(ctx context.Context, metadata types.TableMetadata) error {
+	t.msg.WriteString("*Recent 4xx+ Spikes in last 5 minutes:*\n")
+	return nil
+}
+
+func (t *tableCollector) HandleRecord(ctx context.Context, r *types.Record) error {
+	fmt.Fprintf(&t.msg, "%s \t ---> %s  (>4xx) errors out of %s requests.\n", r.GetDatum("service"),
+																			r.GetDatum("error_count"),
+																			r.GetDatum("total_requests"))
```
vihangm, line 144: nit rm

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,171 @@
...
+
+		// wait for next tick
+		<-ticker.C
+	}
+}
+
+// Implement the TableRecordHandler interface to processes the PxL script table record-wise.
+type tableCollector struct {
+	msg strings.Builder
+	// Channel to communicate to tableMux that tableCollector is done collecting table rows.
```
vihangm, line 132: We aren't really communicating to tableMux. We are just holding that info and using it to block on a call to GetTableDataSync

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		// wait for next tick
+		<-ticker.C
+	}
+}
+
+// Implement the TableRecordHandler interface to processes the PxL script table record-wise.
+type tableCollector struct {
+	msg strings.Builder
```
vihangm, line 131: nit: `msg` is vague. Perhaps `tableDataBuilder`?

demos/slack-alert-app/go/slackbot.go
vihangm: @htroisi I think we should drop the io.EOF check here.
```suggestion
		if err != nil {
```

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (t *tableCollector) HandleInit(ctx context.Context, metadata types.TableMetadata) error {
+	t.tableDataBuilder.WriteString("*Recent 4xx+ Spikes in last 5 minutes:*\n")
+	return nil
+}
+
+func (t *tableCollector) HandleRecord(ctx context.Context, r *types.Record) error {
+	fmt.Fprintf(&t.tableDataBuilder, "%s \t ---> %s  (>4xx) errors out of %s requests.\n", r.GetDatum("service"),
+																							r.GetDatum("error_count"),
+																							r.GetDatum("total_requests"))
```
vihangm, line 119: the tabbing here looks incorrect/weird?
how about the following?
```suggestion
	fmt.Fprintf(&t.tableDataBuilder, "%s \t ---> %s  (>4xx) errors out of %s requests.\n",
		r.GetDatum("service"), r.GetDatum("error_count"), r.GetDatum("total_requests"))```

**Review by vihangm — APPROVED** [review]

File: demos/slack-alert-app/go/slackbot.go [hunk 25]
```diff
@@ -0,0 +1,144 @@
...
+	// Slack channel for Slackbot to post in.
+	// Slack App must be a member of this channel.
+	slackChannel := "#pixie-alerts"
+
+	// This PxL script ouputs a table of the HTTP total requests count and
+	// HTTP error (>4xxx) count for each service in the `px-sock-shop` namespace.
+	// To deploy the px-sock-shop demo, see:
+	// https://docs.pixielabs.ai/tutorials/slackbot-alert for how to
+	b, err := ioutil.ReadFile("http_errors.pxl")
+	pxlScript := string(b)
```
zasgar, line 44: need to check err

demos/slack-alert-app/go/slackbot.go
```diff
@@ -0,0 +1,144 @@
...
+		panic(err)
+	}
+	vz, err := pixieClient.NewVizierClient(ctx, pixieClusterID)
+	if err != nil {
+		panic(err)
+	}
+
+	slackClient := slack.New(slackToken)
+
+	ticker := time.NewTicker(5 * time.Minute)
```
zasgar, line 76: defer ticker.Stop()

zasgar (approved):

PR #201, opened 2021-02-01
Adding gitignore for python

philkuz:


philkuz:
Moved .gitignore up a level because it's necessary for src/api/public as well

zasgar (approved):

PR #211, opened 2021-02-25
Update PxL script Vis Specs to use required arguments.

htroisi:
This PR should cause no visible changes. Instead, it is preparation for @NickLanam's planned UI changes that will alert users to the fact that some scripts have required arguments without which no Live View data will appear:
- Required args will be denoted with an asterisk next to the argument name.
- An error message when the user tries to run a script missing required argument values.

**Vis Spec Arguments:**
- To make a Vis Spec's argument **required**, omit the `defaultValue` field. To run the script, a value must be provided in the Live UI / CLI / API. 
- To make a Vis Spec's argument **optional**, provide a `defaultValue` field containing a string or empty string. An empty string means it's not pre-filled argument, but it's optional anyway.


htroisi:
Updates to the README.md are to fix the linter check

NickLanam:
The invalid vis spec issue turned out to be a false alarm. It is safe to do this with the UI code as it is today. Thanks for the quick update!

NickLanam (approved):

nserrino (approved):

zasgar (approved):

PR #214, opened 2021-03-02
Update Redis PxL scripts to use source / destination columns.

htroisi:
Currently, users need to look at `trace_role`, `pod` and `remote_addr` to determine the "Source" and "Destination" entities for a Redis request. This PR adds `source` and `destination` columns to all of the Redis scripts. 

**Testing**: Tested with off-cluster (`dev-htroisi-memorystore`) and on-cluster redis instances (`dev-cluster-htroisi`). 

**Specific changes**:
- Calculating the source / destinations columns and adding the script references (manual deeplinking) to those columns  are separated into 2 functions. Adding the script reference linking is done at the end of the PxL script, because we want users to be add their own DataFrame operations (filter, groupby, etc) to manipulate the source / destination columns, but that cannot be done on a script reference column. To avoid duplication of work, several columns calculated in `add_source_dest_columns()` are used in `add_source_dest_links()`, this means that any intermediate groupby(), etc functions will need to propagate these columns. 

-  `px/redis_flow_graph`: `edgetThresholds` are educated guesses to match the automatic coloring of the latency columns in the summary table. 

-  `px/redis_stats`: The pod filter filters both the source and destination columns. Because of that change, it also works for remote IP addresses. 

-  `px/redis_stats`: Previously, the timeseries charts used `pod` as the series selector. I changed this to be `destination.` 

**Remaining bugs:** 

- Note that you cannot currently sort the `source` / `destination` columns due to the script_reference() sorting bug (PC-793)

**Next Steps:** 

- Once this PR is approved, these changes will be applied to the rest of the protocol scripts.


oazizi000 (approved):

pxl_scripts/px/redis_data/redis_data.pxl
```diff
@@ -0,0 +1,109 @@
...
+
+
+def redis_data(start: str, num_head: int):
+
+    # redis_events.beta is a beta table, which is subject to change.
+    df = px.DataFrame(table='redis_events.beta', start_time=start)
+    df = add_source_dest_columns(df)
+
+    # Filter source column.
+    # df = df[df.source == 'default/cartservice']
```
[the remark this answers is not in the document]
htroisi: What do you think about leaving this so that it's obvious where to add additional filtering (before the `df.head()` and `add_source_dest_links()` calls)?

pxl_scripts/px/redis_data/redis_data.pxl
```diff
@@ -0,0 +1,109 @@
...
+        'start_time': start,
+        'namespace': df.namespace,
+        'from_entity_filter': '',
+        'to_entity_filter': df.remote_addr,
+        'throughput_filter': '0.0'
+    })
+
+    # If remote_addr is a pod, get its name. If not, use IP address.
+    df.ra_name_link_source = px.select(df.is_ra_pod, df.ra_pod_link, df.ra_link_source)
+    df.ra_name_link_dest = px.select(df.is_ra_pod, df.ra_pod_link, df.ra_link_dest)
```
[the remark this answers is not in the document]
htroisi: Correct. I switched the strategy to your suggestion to below.

pxl_scripts/px/redis_flow_graph/redis_flow_graph.pxl
```diff
@@ -15,52 +16,134 @@ window_ns = px.DurationNanos(10 * ns_per_s)
...
-    df.namespace = df.ctx['namespace']
-    df = df.drop(['upid'])
+    df = add_source_dest_columns(df)
 
     # Filter on namespace as specified by the user.
     df = df[df.namespace == ns]
 
+    # Filter out messages with empty source / destination.
+    df = df[df.source != '']
+    df = df[df.destination != '']
```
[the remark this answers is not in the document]
htroisi: I think `add_source_dest_columns()` should be identical between all protocol scripts. I'm assuming we want to show requests with empty source / destination in `redis_data()` and `redis_stats()`.

pxl_scripts/px/redis_data/redis_data.pxl
```diff
@@ -0,0 +1,108 @@
+# Copyright (c) Pixie Labs, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+''' Redis Data Tracer
+
+This script traces all Redis data on the cluster.
```
nserrino, line 6: nit but the script itself doesn't trace the events. this script displays the most recent events

pxl_scripts/px/redis_data/redis_data.pxl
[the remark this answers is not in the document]
nserrino: I think it at least needs a comment explaining why it's there like, "you can add a filter here", but I think that it's still probably overkill

pxl_scripts/px/redis_flow_graph/redis_flow_graph.pxl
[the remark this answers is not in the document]
oazizi000: Agree with keeping them identical. I'd actually argue just filtering out all these empty sources in all cases. When one of them is empty, it means the process traced was not a K8s pod.

pxl_scripts/px/redis_data/redis_data.pxl
```diff
@@ -0,0 +1,108 @@
...
+    df = add_source_dest_columns(df)
+
+    # Filter source column.
+    # df = df[df.source == 'default/cartservice']
+
+    # Restrict number of results.
+    df = df.head(num_head)
+
+    df = add_source_dest_links(df, start)
+    df = df['time_', 'source', 'destination', 'remote_port', 'cmd',
```
nserrino, line 25: this currently works (we have an issue filed) but is technically not valid pandas, can you change to double [] in case we fix it?

pxl_scripts/px/redis_data/redis_data.pxl
```diff
@@ -0,0 +1,108 @@
...
+            'cmd_args', 'resp', 'latency_ns']
+
+    return df
+
+
+def add_source_dest_columns(df):
+    ''' Adds source, destination columns to the dataframe.
```
nserrino, line 32: I would either put the docstring or not, some of your funcs have it some don't. Also I think it would be more understandable if instead of a docstring, it instead explained what source and destination columns are in this context and what this logic is doing (in plain language)

pxl_scripts/px/redis_data/redis_data.pxl
```diff
@@ -0,0 +1,108 @@
...
+
+    @df: A DataFrame containing trace_role, upid, remote_addr columns.
+
+    Returns: The input DataFrame with source, destination, pod, namespace,
+    is_source_pod_type, is_dest_pod_type columns added.
+    '''
+    df.pod = df.ctx['pod']
+    df.namespace = df.ctx['namespace']
+
+    # TODO: remove hack to get around PP-2445.
```
nserrino, line 42: this should be fixed soon :)

pxl_scripts/px/redis_data/redis_data.pxl
```diff
@@ -0,0 +1,108 @@
...
+
+    @df: A DataFrame containing source, destination, is_source_pod_type,
+    is_dest_pod_type, namespace columns.
+    @start: The timestamp of data to start at.
+
+    Returns: The input DataFrame with script references added to the source,
+    destinaton columns.
+    '''
+
+    # If source is a pod, link to px/pod. If an IP addr, link to px/net_flow_graph.
```
nserrino, line 77: nit: If source is an IP address

pxl_scripts/px/redis_data/redis_data.pxl
```diff
@@ -0,0 +1,108 @@
...
+    df.source = px.select(df.is_server_tracing, df.ra_name, df.pod)
+    df.destination = px.select(df.is_server_tracing, df.pod, df.ra_name)
+
+    df = df.drop(['ra_pod', 'is_ra_pod', 'ra_name', 'is_server_tracing'])
+
+    return df
+
+
+def add_source_dest_links(df, start: str):
+    ''' Adds source, destination columns with script reference links to the DataFrame.
```
nserrino, line 67: same feedback here, I don't think the user understands what source, destination, script reference are so I would consider replacing the docstring you have with some high level explanation about the logic

pxl_scripts/px/redis_flow_graph/redis_flow_graph.pxl
```diff
@@ -15,52 +16,131 @@ window_ns = px.DurationNanos(10 * ns_per_s)
...
     df = px.DataFrame('redis_events.beta', start_time=start)
-    df.pod = df.ctx['pod']
-    df.namespace = df.ctx['namespace']
-    df = df.drop(['upid'])
+    df = add_source_dest_columns(df)
 
     # Filter on namespace as specified by the user.
     df = df[df.namespace == ns]
 
+    # Filter out messages with empty source / destination.
```
nserrino, line 26: Does this happen in the data we have?

pxl_scripts/px/redis_data/redis_data.pxl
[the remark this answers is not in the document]
oazizi000: I agree that hint is useful. In that case, we should add some script arguments for the source/destination filtering. That'll be useful across all scripts, and shows where filtering should be done.

And it can also be followed by a comment like: `# Add additional filters here`.

pxl_scripts/px/redis_flow_graph/redis_flow_graph.pxl
```diff
@@ -15,52 +16,131 @@ window_ns = px.DurationNanos(10 * ns_per_s)
...
-
-    # Specify to and from entities.
-    df = df[df.remote_addr != '']
-    df.to_entity = df.pod
-    df.from_entity = px.pod_id_to_pod_name(px.ip_to_pod_id(df.remote_addr))
-
     df.request_throughput = df.throughput_total / window_ns
-    df = df.groupby(['from_entity', 'to_entity']).agg(
+
+    df = df.groupby(['source', 'destination', 'is_source_pod_type', 'is_dest_pod_type', 'namespace']).agg(
```
nserrino, line 48: I have an idea for how you can get around the "Mean of percentiles" problem but since it affects other diffs besides this one it's probably best done as a separate change

pxl_scripts/px/redis_flow_graph/redis_flow_graph.pxl
```diff
@@ -15,52 +16,131 @@ window_ns = px.DurationNanos(10 * ns_per_s)
...
-    df = df[px.contains(df.from_entity, from_entity_filter)]
-    df = df[px.contains(df.to_entity, to_entity_filter)]
+    return df
+
+
+def redis_summary_with_links(start: str, ns: px.Namespace, from_entity_filter: str, to_entity_filter: str):
+
+    df = redis_flow_graph(start, ns, from_entity_filter, to_entity_filter)
+    df = add_source_dest_links(df, start)
+    df = df['source', 'destination', 'latency_p50', 'latency_p90',
```
nserrino, line 63: same thing about [[]] rather than []

pxl_scripts/px/redis_stats/redis_stats.pxl
```diff
@@ -51,9 +40,10 @@ def pod_redis_let(start: str, pod: px.Pod):
     Returns: Returns the DataFrame containing LET time-series for Redis
         traffic to a Redis database pod.
     '''
-    df = redis_let_per_pod(start, pod, ['timestamp', k8s_object])
-    return df['time_', split_series_name, 'latency_p50',
-              'latency_p90', 'latency_p99', 'request_throughput']
+    df = redis_let_per_pod(start, pod, ['timestamp', 'destination'])
+
+    return df['time_', 'destination', 'latency_p50', 'latency_p90',
```
nserrino, line 45: [[]]

pxl_scripts/px/redis_flow_graph/redis_flow_graph.pxl
htroisi: Since this problem exists in many scripts, it'll be in a different diff. Is your idea something different than the UDF in PP-2491?

pxl_scripts/px/redis_flow_graph/redis_flow_graph.pxl
htroisi: I moved this logic to add_source_dest_columns(), so it'll be applied in all of the scripts. I haven't seen this issue on my Redis test clusters. According to Omid, this happens when the process traced was not a K8s pod.

pxl_scripts/px/redis_flow_graph/redis_flow_graph.pxl
nserrino: yeah. We can add a UDF that does conversion from "-5m" to a time. (we've talked about it before). This removes the need to do 2 aggregates in a row. The reason we do 2 aggregates is basically trying to work around the fact that we don't have a way of converting the total time range to a time. (easier to explain over zoom if you have further questions)

another solution that doesn't require any UDFs is splitting the throughput calculation and the latency calculation into 2 separate tables. throughput will be the result of 2 different aggregates. Latency will be the result of 1 aggregate (since we don't need 2 aggregates for it like we do with throughput right now). Then, we join them together. This solution could work today without any compiler changes but it is a little confusing to the reader.

pxl_scripts/px/redis_data/redis_data.pxl
htroisi: Check out the new comment.

pxl_scripts/px/redis_data/redis_data.pxl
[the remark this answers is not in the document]
htroisi: - `redis_data`: I added `source_filter` and `destination_filter` which use px.contains() instead of exact string matching.
- `redis_flow_graph`: I renamed the to/from entity filtering to source/dest filtering. 
- `redis_stats` has a pod filter which checks that either the source or destination contains the filter name. It will filter IP addresses too.

oazizi000 (approved):

nserrino (approved):

zasgar (approved):

PR #222, opened 2021-03-23
Flamegraph scripts

oazizi000:


pxl_scripts/px/perf_flamegraph/trace.pxl
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) Pixie Labs, Inc.
+# Licensed under the Apache License, Version 2.0 (the "License")
+
+'''PEM Flamegraph
```
htroisi, line 4: CPU Flamegraph?

pxl_scripts/px/perf_flamegraph/vis.json
```diff
@@ -0,0 +1,82 @@
...
+    }
+  ],
+  "globalFuncs": [],
+  "widgets": [
+    {
+      "name": "Flamegraph",
+      "position": {
+        "x": 0,
+        "y": 0,
+        "w": 10,
```
htroisi, line 46: Why not full width (12)?

pxl_scripts/px/perf_flamegraph/vis.json
```diff
@@ -0,0 +1,82 @@
...
+      "defaultValue": "-5m"
+    },
+    {
+      "name": "node",
+      "type": "PX_NODE",
+      "description": "The node to filter on.",
+      "defaultValue": ""
+    },
+    {
+      "name": "namespace",
```
htroisi, line 16: I'd switch the order of the arguments: namespace > node > pod makes more sense.

pxl_scripts/px/perf_flamegraph/trace.pxl
```diff
@@ -0,0 +1,47 @@
...
+    node_agg = df.groupby([pct_grouping_entity]).agg(
+        count=('count', px.sum)
+    )
+
+    # Apply filters.
+    df = df[px.contains(df.node, node)]
+    df = df[px.contains(df.pod, pod)]
+    df = df[px.contains(df.namespace, namespace)]
+
+    # Aggregate across time windows.
```
htroisi, line 32: IIUC, there aren't really time windows, just samples with timestamps. Maybe "Aggregate across all samples?"

pxl_scripts/px/perf_flamegraph/vis.json
oazizi000: Acutally, node is independent of namespace/pod/container/pid.

For example, a node can contain multiple namespaces. Also, a namespace can span multiple nodes. So it's really an any-to-any relationship, not a parent-child relationship.

So we need to keep the namespace and pod filters adjacent to each other. Those two do have a parent-child relationship.

pxl_scripts/px/perf_flamegraph/trace.pxl
oazizi000: These are actually aggregating different flamegraphs into each other. For example, we dump a flamegraph every 30 seconds for the last 30-second time window. But if we query for the last five minutes, it'll take the ten flamegraphs and merge them into one larger flamegraph that covers the entire ten minute time window.

I'll update the comment to make that a bit clearer.

pxl_scripts/px/perf_flamegraph/vis.json
htroisi: Oh, whoops. I'm not sure what I was thinking. :)

htroisi (approved):

zasgar (approved):

PR #245, opened 2021-05-05
Fix 'idz' is not defined

Negashev:
Fix 'idz' is not defined


zasgar:
@Negashev thanks for contributing this! Looks good to me. 

We are in the process of figuring out our CLAs/commit process before we merge this. Let me circle back soon with the next steps.

htroisi:
Our two-way syncing isn't set up yet, so I ended up fixing this upstream after another person ran into this bug.

PR #297, opened 2021-07-28
Update AugmentedTokenValidDuration to 90 min

robertprast:
Session timeouts related to the augmented token present a less than ideal UXP flow. To mitigate the need to constantly grab a refresh token on long lived use of Pixie, we can up the duration of the token to 90 minutes.


zasgar (approved):

vihangm:
Merged as 354fc69233a809bc4e0d6a3b7649d6928d72d1d7

PR #329, opened 2021-10-13
fixup validation errors for kube 1.22 deprecations

winmillwill:
Was just taking the self-hosted instructions for a spin and followed
this guide

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22

... very naively to get the validation errors to stop. My editor also
auto-formatted so the reader may want to use the `w=1` query param.

I ended up just dropping all the pre-v1 versions of the crd's to get
something working quickly. I understand there are probably a number of
reasons why this isn't usable as is, but I figured I'd at least put it
up in case anyone could use it for prior art.


zasgar:
Thanks for the contribution! Merged.

PR #330, opened 2021-10-14
Add additional docs about how to sign commits

hmstepanek:
Add additional documentation explaining how to configure git commit signing and link to the appropriate github docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,20 +118,18 @@ Then you just add a line to every git commit message:
 
 Use your real name (sorry, no pseudonyms or anonymous contributions.)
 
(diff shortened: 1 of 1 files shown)


vihangm:
This is actually conflating 2 kinds of signing. One of which is the GPG signature on commits, and the latter is just the DCO commit signoff. (`git commit -S` for GPG vs `git commit -s` for the DCO signoff).

Since our external contribution process currently requires a Pixie maintainer to patch and submit the commit, only the DCO signing is required for external github PRs. The commit will be GPG signed by the actual committer which is copybara, for eg:

Screen Shot 2021-10-17 at 6 32 14 PM

vihangm:
I think the reformatting makes sense, but I'd remove the GPG signing instructions to avoid conflating the two.

PR #334, opened 2021-10-19
Add MatchRegexRule UDF

elaguerta:
Add MatchRegexRule UDF for use in security PxL scripts. 
* Add a UDF, MatchRegexRule, that takes an encoded json of regular expression rules, and string. It returns the first rule that matches the string or empty string if no match.
* Add unit tests for MatchRegexRule.

### DIFF
--- a/src/carnot/funcs/builtins/regex_ops.cc
+++ b/src/carnot/funcs/builtins/regex_ops.cc
@@ -32,7 +32,7 @@ void RegisterRegexOpsOrDie(udf::Registry* registry) {
   // String contains.
   registry->RegisterOrDie<RegexMatchUDF>("regex_match");
   registry->RegisterOrDie<RegexReplaceUDF>("replace");
-  registry->RegisterOrDie<MatchRegexRule>("match_regex_rule");
+  registry->RegisterOrDie<MatchRegexRule>("_match_regex_rule");
   /*****************************************
    * Aggregate UDFs.
    *****************************************/
--- a/src/carnot/funcs/builtins/regex_ops.cc
+++ b/src/carnot/funcs/builtins/regex_ops.cc
@@ -32,6 +32,7 @@ void RegisterRegexOpsOrDie(udf::Registry* registry) {
   // String contains.
   registry->RegisterOrDie<RegexMatchUDF>("regex_match");
   registry->RegisterOrDie<RegexReplaceUDF>("replace");
+  registry->RegisterOrDie<MatchRegexRule>("match_regex_rule");
   /*****************************************
    * Aggregate UDFs.
    *****************************************/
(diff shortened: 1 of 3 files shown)


hmstepanek:
This is ready for re-review.

src/carnot/funcs/builtins/regex_ops.h
```diff
@@ -110,6 +112,55 @@ class RegexReplaceUDF : public udf::ScalarUDF {
...
+   return Status::OK();
+  }
+  
+  types::StringValue Exec(FunctionContext *ctx, StringValue value) {
+    for (auto& [rule_name, regex_match_udf] : regex_rules_map_) {
+      if (regex_match_udf.Exec(ctx, value).val) {
+        return rule_name;
+      }
+    }
+  return "";
```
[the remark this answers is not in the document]
hmstepanek: I've changed the internal data structure from a map to a vector pair. I also added a length variable to keep track of the size/length of the vector since I was having trouble using .size() to automatically determine it (this was returning a long instead of an int for comparison). Let me know if there's a better way to do this!

src/carnot/funcs/builtins/regex_ops.h
```diff
@@ -110,6 +112,63 @@ class RegexReplaceUDF : public udf::ScalarUDF {
...
+      return Status(statuspb::Code::INVALID_ARGUMENT, "unable to parse string as json");
+    }
+    // Populate the parse regular expressions into self::regex_rules.
+    for (rapidjson::Value::ConstMemberIterator itr = regex_rules_json.MemberBegin();
+         itr != regex_rules_json.MemberEnd(); ++itr) {
+      RegexMatchUDF regex_match_udf;
+      std::string name = itr->name.GetString();
+      std::string regex_pattern = itr->value.GetString();
+      PL_RETURN_IF_ERROR(regex_match_udf.Init(ctx, regex_pattern));
+      regex_rules.push_back(make_pair(name, std::move(regex_match_udf)));
```
zasgar, line 131: emplace_back instead of push_back?

hmstepanek:
@zasgar Changes addressed!

src/carnot/funcs/builtins/regex_ops.cc
```diff
@@ -32,6 +32,7 @@ void RegisterRegexOpsOrDie(udf::Registry* registry) {
   // String contains.
   registry->RegisterOrDie<RegexMatchUDF>("regex_match");
   registry->RegisterOrDie<RegexReplaceUDF>("replace");
+  registry->RegisterOrDie<MatchRegexRule>("match_regex_rule");
```
zasgar, line 35: Can we mark this as an internal/unsupported function? Renaming it to `_match_regex_rule` will do.

hmstepanek: Done!

vihangm:
Looks like this was merged in as 983f23c

PR #336, opened 2021-10-21
Add reflected XSS PxL script

hmstepanek:
Detect reflected XSS attacks by looking for an attack string that appears both in the request and the response.

### DIFF
--- a/src/carnot/funcs/builtins/regex_ops.h
+++ b/src/carnot/funcs/builtins/regex_ops.h
@@ -22,6 +22,8 @@
 #include <algorithm>
 #include <memory>
 #include <string>
+#include <regex>
+#include <rapidjson/document.h>
 #include "re2/re2.h"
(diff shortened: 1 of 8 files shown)


nserrino (commented):
I think it might be cleaner if we just embed the rules inside a XSS UDF, rather than expose it in the PxL script. Then we could replace match_regex_rule with that XSS UDF.  It seems like a cleaner interface for the user. Thoughts?

hmstepanek:
@nserrino The intention here is that users have control over what regular expressions they use. This will:
1. Enable users to control what gets flagged in case one of the regexes produces a false positive on their specific application.
2. Prevent hackers from using a posted list of regexes to avoid detection.
3. Make it easier to add to this list overtime since it is not a comprehensive list of all possible injections but a starting point from which to improve upon.

PR #359, opened 2021-11-17
helm - add support for pod nodeSelector and affinity

smcavallo:
vizier only runs on linux nodes.  
add support for `nodeSelector` so consumers can restrict the pods where vizier can be run.


aimichelle:
Hi @smcavallo ! Thanks for sending out this change! Unfortunately, some more work would need to be done to actually propagate this to the downstream vizier resources. This change only adds the these fields to the CRD, but not to the vizier resources that get deployed by the operator. 
Luckily, I don't think it should be particularly difficult to add in. We can follow the `resourceRequests` as an example: https://github.com/pixie-io/pixie/blob/ea6070e6484c7c0c1dde517187d58cd35a0eac49/src/operator/controllers/vizier_controller.go#L756 

Let me know if you have any questions!

smcavallo:
Hi @aimichelle - thank you so much!  I *think* I did the right thing but I have no idea how to test.  Hoping you can review.

smcavallo:
See - https://github.com/pixie-io/pixie/pull/362

PR #351, opened 2021-11-08
[Seeking additional feedback] Integrate mux parser into stirling

ddelnano:
This is a PR for integrating the mux protocol parser into sterling. It is currently in an incomplete state, but I'm opening this up to more easily track the outstanding work

## Todo
- [x] Add a `socket_tracer/mux_trace_bpf_test.cc` test
- [ ] Determine why the `mux_trace_bpf_test` is unable to see other mux messages
- [ ] Add more verification to the protocol_inference test
- [x] Update the mux table to include relevant mux details
- [x] Remove test cases to `socket_trace_connector_test.cc` (these are legacy per the discussion in this PR)
- [x] Create thriftmux docker image in `socket_tracer/testing/containers`
- [ ] Update to use pixie gcr docker registry thriftmux image

PR #366, opened 2021-11-27
Add the base for thriftmux scala docker images (layers) used for end to end bpf testing

ddelnano:
This PR pulls out the thriftmux container changes necessary for #351. It also switches the build process to use bazel entirely rather than using docker directly.

## Testing
- [x] Used this change in #367 to verify that it works properly

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -100,6 +100,58 @@ load("@vizier_api_python_deps//:requirements.bzl", "install_deps")
 
 install_deps()
 
+# java_external
+load("@rules_jvm_external//:defs.bzl", "maven_install")
+
+finagle_version = "21.4.0"
+scala_version = "2.13.6"
+scala_minor_version = ".".join(scala_version.split(".")[:2])
+maven_install(
+    artifacts = [
+        "com.twitter:finagle-thriftmux_%s:%s" % (scala_minor_version, finagle_version),
+        "com.twitter:finagle-mux_%s:%s" % (scala_minor_version, finagle_version),
+        "com.twitter:finagle-core_%s:%s" % (scala_minor_version, finagle_version),
+        "com.twitter:scrooge-core_%s:%s" % (scala_minor_version, finagle_version),
+        "com.twitter:finagle-http_%s:%s" % (scala_minor_version, finagle_version),
+        "org.apache.thrift:libthrift:0.10.0",
+    ],
+    repositories = ["https://repo1.maven.org/maven2"],
+)
+
+# Setup scala_image
+load("@io_bazel_rules_scala//:scala_config.bzl", "scala_config")
+scala_config(scala_version = scala_version)
+
+load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repositories")
+scala_repositories()
+
+load("@io_bazel_rules_scala//scala:toolchains.bzl", "scala_register_toolchains")
+scala_register_toolchains()
+
+load(
+    "@io_bazel_rules_docker//scala:image.bzl",
+    scala_image_repos = "repositories",
+)
+
+scala_image_repos()
+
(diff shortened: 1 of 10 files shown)


WORKSPACE
```diff
@@ -100,6 +100,58 @@ load("@vizier_api_python_deps//:requirements.bzl", "install_deps")
 
 install_deps()
 
+# java_external
```
ddelnano, line 103: I wasn't sure the best way to structure this. It appears most of the bazel code is refactored to into macros, etc. I tried to fit the rule installation into the existing pattern, but I wasn't sure the best way to layout this code.

Let me know if there is more appropriate way to structure this.

bazel/repository_locations.bzl
```diff
@@ -69,9 +76,9 @@ REPOSITORY_LOCATIONS = dict(
         urls = ["https://github.com/bazelbuild/bazel-skylib/archive/1.0.2.tar.gz"],
     ),
     io_bazel_rules_docker = dict(
-        sha256 = "1f4e59843b61981a96835dc4ac377ad4da9f8c334ebe5e0bb3f58f80c09735f4",
-        strip_prefix = "rules_docker-0.19.0",
-        urls = ["https://github.com/bazelbuild/rules_docker/archive/v0.19.0.tar.gz"],
+        sha256 = "4349f2b0b45c860dd2ffe18802e9f79183806af93ce5921fb12cbd6c07ab69a8",
+        strip_prefix = "rules_docker-0.21.0",
+        urls = ["https://github.com/bazelbuild/rules_docker/archive/v0.21.0.tar.gz"],
```
ddelnano, line 81: This is required because there was a bug in the underlying `scala_rules` (https://github.com/bazelbuild/rules_docker/pull/1954)

bazel/repository_locations.bzl
```diff
@@ -31,6 +31,13 @@ REPOSITORY_LOCATIONS = dict(
             "https://github.com/bazelbuild/rules_go/releases/download/v0.26.0/rules_go-v0.26.0.tar.gz",
         ],
     ),
+    io_bazel_rules_scala = dict(
+        sha256 = "76e1abb8a54f61ada974e6e9af689c59fd9f0518b49be6be7a631ce9fa45f236",
+        urls = [
+            "https://github.com/bazelbuild/rules_scala/archive/e7a948ad1948058a7a5ddfbd9d1629d6db839933.zip",
+        ],
+        strip_prefix = "rules_scala-%s" % "e7a948ad1948058a7a5ddfbd9d1629d6db839933",
```
oazizi000, line 39: nit: Can't we do the following?

```
strip_prefix = "rules_scala-e7a948ad1948058a7a5ddfbd9d1629d6db839933"
```

oazizi000 (commented):
Overall looks good. I do share your question about whether there's a better way to structure the WORKSPACE file, so let me get another pair of eyes on that.

bazel/repository_locations.bzl
oazizi000: Because this has the potential to break things, let me try to pull this in as a separate change.

bazel/repository_locations.bzl
ddelnano: Updated. This was copy and pasted from when I had it in the WORKSPACE file and was taking advantage of using a variable.

bazel/repository_locations.bzl
oazizi000: 82386b8 pulled this in.

ddelnano: Thanks for doing that. Rebased this branch and removed this as an outstanding item to test.

oazizi000:
Committed in ce4ee41afd74e3d46418d3a85c05f92fca98746b

PR #367, opened 2021-11-29
Integrate mux parser into stirling

ddelnano:
This is my second attempt at integrating the mux parser into stirling. I'm opening this preemptively to show that the testing for #366 was successful, but we need to merge that PR first and rebase to pull out the docker image related changes. 

## Todo
- [x] Fix issue that is causing a subset of mux data from being collected
- [x] Debug why a bug in Tping/Rping parsing causes the mux stitcher to lose data 
  - Incorrectly stripping the buffer caused the Tdispatch/Rdispatch frames to go missing. Fixing that bug causes all of the frames to exist except the Rping/Tping (since the Rping isn't parsed correctly). After looking at this again with more experience debugging this problem, I'm able to verify that the stitcher behaves as expected even with the Rping parser bug
- [x] Decrease false positive rate of mux protocol inference. I attempted to do this by parsing more of the mux message when the entire frame is available.
- [x] Avoid mutating mux stitcher deques by using iterators
- [ ] Understand T/Rinit message structure better (6 unknown bytes) and update the mux parser accordingly

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
+++ b/src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
@@ -541,7 +541,7 @@ static __inline enum message_type_t infer_mux_message(const char* buf, size_t co
   }
 
   if (tag < 1 || tag > ((1 << 23) - 1)) {
-       return kUnknown;
+    return kUnknown;
   }
 
   return msg_type;
--- a/src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
+++ b/src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
@@ -487,6 +487,54 @@ static __inline bool is_redis_message(const char* buf, size_t count) {
   return true;
 }
 
+static __inline enum message_type_t infer_mux_message(const char* buf, size_t count) {
+  static const int8_t kTreq = 1;
+  static const int8_t kRreq = -1;
+  static const int8_t kTdispatch = 2;
+  static const int8_t kRdispatch = -2;
+  static const int8_t kTdrain = 64;
+  static const int8_t kRdrain = -64;
+  static const int8_t kTping = 65;
+  static const int8_t kRping = -65;
+  static const int8_t kTdiscarded = 66;
+  static const int8_t kRdiscarded = -66;
+  static const int8_t kTlease = 67;
+  static const int8_t kTinit = 68;
+  static const int8_t kRinit = -68;
+  static const int8_t kRerr = -128;
+  static const int8_t kTdiscardedOld = -62;
+  static const int8_t kRerrOld = 127;
+
+  if (count < 8) {
+    return kUnknown;
+  }
+
+  int8_t mux_type = buf[4];
+  switch (mux_type) {
+    case kTreq:
+    case kTdispatch:
+    case kTdrain:
+    case kTping:
+    case kTdiscarded:
+    case kTinit:
+    case kTlease:
+    case kTdiscardedOld:
+    // TODO(ddelnano): Verify if this should be a req or a resp
+    case kRerrOld:
+      return kRequest;
+    case kRreq:
+    case kRdispatch:
+    case kRdrain:
+    case kRping:
+    case kRdiscarded:
+    case kRinit:
+    case kRerr:
+      return kResponse;
+    default:
+      return kUnknown;
+  }
+}
+
 // NATS messages are in texts. The role is inferred from the message type.
 // See https://github.com/nats-io/docs/blob/master/nats_protocol/nats-protocol.md
 //
@@ -555,6 +603,8 @@ static __inline struct protocol_message_t infer_protocol(const char* buf, size_t
     inferred_message.protocol = kProtocolPGSQL;
   } else if ((inferred_message.type = infer_mysql_message(buf, count, conn_info)) != kUnknown) {
     inferred_message.protocol = kProtocolMySQL;
+  } else if ((inferred_message.type = infer_mux_message(buf, count)) != kUnknown) {
+    inferred_message.protocol = kProtocolMux;
   } else if ((inferred_message.type = infer_kafka_message(buf, count, conn_info)) != kUnknown) {
     inferred_message.protocol = kProtocolKafka;
   } else if ((inferred_message.type = infer_dns_message(buf, count)) != kUnknown) {
--- a/src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
+++ b/src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
@@ -488,56 +488,76 @@ static __inline bool is_redis_message(const char* buf, size_t count) {
 }
 
 static __inline enum message_type_t infer_mux_message(const char* buf, size_t count) {
-  static const int8_t kTreq = 1;
-  static const int8_t kRreq = -1;
+  // mux's on the wire format causes false positives for protocol inference
+  // In order to address this, we only infer mux messages by the
+  // most useful message types and if they are easy to identify
   static const int8_t kTdispatch = 2;
   static const int8_t kRdispatch = -2;
-  static const int8_t kTdrain = 64;
-  static const int8_t kRdrain = -64;
-  static const int8_t kTping = 65;
-  static const int8_t kRping = -65;
-  static const int8_t kTdiscarded = 66;
-  static const int8_t kRdiscarded = -66;
-  static const int8_t kTlease = 67;
   static const int8_t kTinit = 68;
   static const int8_t kRinit = -68;
   static const int8_t kRerr = -128;
-  static const int8_t kTdiscardedOld = -62;
   static const int8_t kRerrOld = 127;
+  uint32_t mux_header_size = 8;
+  // TODO(ddelnano): Determine why mux-framer text in T/Rinit is
+  // 6 bytes after the mux header
+  int32_t mux_framer_pos = mux_header_size + 6;
 
-  if (count < 8) {
+  if (count < mux_header_size) {
     return kUnknown;
   }
 
+  uint32_t length = read_big_endian_int32(buf) + 4;
   enum message_type_t msg_type;
 
   int32_t type_and_tag = read_big_endian_int32(buf + 4);
   int8_t mux_type = (type_and_tag & 0xff000000) >> 24;
   uint32_t tag = (type_and_tag & 0xffffff);
   switch (mux_type) {
-    case kTreq:
     case kTdispatch:
-    case kTdrain:
-    case kTping:
-    case kTdiscarded:
     case kTinit:
-    case kTlease:
-    case kTdiscardedOld:
-    // TODO(ddelnano): Verify if this should be a req or a resp
     case kRerrOld:
       msg_type = kRequest;
       break;
-    case kRreq:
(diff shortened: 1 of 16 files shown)


src/stirling/source_connectors/socket_tracer/mux_table.h
```diff
@@ -0,0 +1,66 @@
...
+         types::SemanticType::ST_NONE,
+         types::PatternType::GENERAL_ENUM},
+        {"resp_type", "Mux message response type",
+         types::DataType::INT64,
+         types::SemanticType::ST_NONE,
+         types::PatternType::GENERAL_ENUM},
+        {"tag", "Mux message request tag",
+         types::DataType::INT64,
+         types::SemanticType::ST_NONE,
+         types::PatternType::GENERAL_ENUM},
```
oazizi000, line 48: Is this tag used to match requests and responses. If so, do we need in the output table?

src/stirling/source_connectors/socket_tracer/mux_trace_bpf_test.cc
```diff
@@ -0,0 +1,176 @@
...
+    r.req.tag = tag;
+    r.req.type = static_cast<int8_t>(rb[kMuxReqTypeIdx]->Get<types::Int64Value>(idx).val);
+    r.resp.tag = tag;
+    r.resp.type = static_cast<int8_t>(rb[kMuxRespTypeIdx]->Get<types::Int64Value>(idx).val);
+    result.push_back(r);
+  }
+  return result;
+}
+
+mux::Record GetRecordWithTagAndType(uint24_t tag, mux::Type req_type, mux::Type resp_type) {
```
oazizi000, line 117: nit/naming: `GetRecordWithTagAndType` gets confusing with `GetTargetRecords` where "get" implies getting out of the records. Can we rename to just  `RecordWithTagAndType`

src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference_test.cc
```diff
@@ -369,3 +369,45 @@ TEST(ProtocolInferenceTest, NATS) {
...
+  // clang-format on
+
+  auto protocol_message =
+      infer_protocol(reinterpret_cast<const char*>(kReqFrame), sizeof(kReqFrame), &conn_info);
+  EXPECT_EQ(protocol_message.protocol, kProtocolMux);
+  EXPECT_EQ(protocol_message.type, kRequest);
+
+  /* protocol_message = infer_protocol(reinterpret_cast<const char*>(kReqHeaderFrame), */
+  /*                                   sizeof(kReqHeaderFrame), &conn_info); */
+  /* EXPECT_EQ(protocol_message.protocol, kProtocolUnknown); */
```
oazizi000, line 407: Remove?

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -831,6 +837,28 @@ void SocketTraceConnector::AppendMessage(ConnectorContext* ctx, const ConnTracke
...
+
+  DataTable::RecordBuilder<&kMuxTable> r(data_table, entry.resp.timestamp_ns);
+  r.Append<r.ColIndex("time_")>(entry.resp.timestamp_ns);
+  r.Append<r.ColIndex("upid")>(upid.value());
+  r.Append<r.ColIndex("remote_addr")>(conn_tracker.remote_endpoint().AddrStr());
+  r.Append<r.ColIndex("remote_port")>(conn_tracker.remote_endpoint().port());
+  r.Append<r.ColIndex("trace_role")>(conn_tracker.role());
+  r.Append<r.ColIndex("req_type")>(entry.req.type);
+  r.Append<r.ColIndex("resp_type")>(entry.resp.type);
+  r.Append<r.ColIndex("tag")>(int32_t(entry.req.tag));
```
chengruizhe, line 854: static_cast<int32_t>

chengruizhe (commented):
Mostly looks good. Let's talk about how to test the Mux protocol inference rules on slack.

src/stirling/source_connectors/socket_tracer/mux_table.h
ddelnano: Since mux multiplexes connections, I could see this being useful for differentiating when a service is making concurrent requests to the same backend instance (tag values would be different). This might be a niche use case, but I think it's worth keeping.

src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference_test.cc
ddelnano: Sorry I was planning to remove this once I revamped the protocol inference :). So this will be gone soon.

src/stirling/source_connectors/socket_tracer/mux_table.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include "src/stirling/core/types.h"
+#include "src/stirling/source_connectors/socket_tracer/canonical_types.h"
+#include "src/stirling/source_connectors/socket_tracer/protocols/mux/types.h"
+
+namespace px {
+namespace stirling {
+
+// clang-format off
+static constexpr DataElement kMuxElements[] = {
```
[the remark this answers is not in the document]
ddelnano: I think it depends how we handle the nested thrift protocol data in thriftmux. I was picturing that the "body" field would be stored in a different table populated by a thrift protocol parser. That would allow non thriftmux users to get thrift protocol parsing.

However, it's not clear to me how we would join the mux and thrift procotol data in that situation. In an ideal world, the mux and thrift data would be easily joined. Has there been any discussion on how to handle protocols that are nested like this?

src/stirling/source_connectors/socket_tracer/mux_trace_bpf_test.cc
ddelnano: Great point, updated.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
ddelnano: Updated

src/stirling/source_connectors/socket_tracer/mux_table.h
```diff
@@ -0,0 +1,66 @@
...
+        canonical_data_elements::kTime,
+        canonical_data_elements::kUPID,
+        canonical_data_elements::kRemoteAddr,
+        canonical_data_elements::kRemotePort,
+        canonical_data_elements::kTraceRole,
+        {"req_type", "Mux message request type",
```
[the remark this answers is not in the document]
ddelnano: They are all paired except for `Tlease`. So if we decided we should only keep one, then I think the `req_type` is the one to do.

src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
```diff
@@ -487,6 +487,66 @@ static __inline bool is_redis_message(const char* buf, size_t count) {
   return true;
 }
 
+static __inline enum message_type_t infer_mux_message(const char* buf, size_t count) {
+  static const int8_t kTreq = 1;
```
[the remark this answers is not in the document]
ddelnano: I attempted to enhance the mux protocol inference by reducing the number of types (R/Tinit, R/Tdispatch, Rerr(Old)) and by parsing 'special' strings that are present in those frame types. Ideally we would support all mux types, but there are some that aren't as high value.

If I remember correctly, the protocol inference runs on a connection until it's classified and is only provided the data from the syscall, correct? If that's the case I was hoping that having stronger verification on Rerr, Tinit and Rinit would help reduce false positives since it's earlier in the connection lifecycle. Let me know what you think of my latest update.

I am definitely interested in trying the protocol inference against the network traffic dataset. I shared the mux pcap file that I generated myself with you in slack. Unfortunately I don't know if I'll be able to provide real application traffic, but let me know how I can help with this testing.

src/stirling/source_connectors/socket_tracer/mux_table.h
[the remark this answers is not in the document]
oazizi000: We haven't had nested protocols like this before, so we'll be treading new ground. The problem is that we have only one ConnTracker per connection, and there's an assumption that the ConnTracker will output to one table.

I suppose we could detect the more specific inner protocol and choose the more specific one when possible. Needs more discussion.

src/stirling/source_connectors/socket_tracer/mux_table.h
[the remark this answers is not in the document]
oazizi000: It's not so much a question of storage, but rather one of not having redundant data. I'd suggest we just keep `req_type` and drop `resp_type`.

src/stirling/source_connectors/socket_tracer/mux_table.h
oazizi000: We typically try to hide details that are not at the protocol's logical level. So for consistency with other protocols, I'd like to leave it out for now.

Because of issues with creating legacy, it's also easier to add it in later versus removing it.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.h
```diff
@@ -341,6 +343,7 @@ class SocketTraceConnector : public SourceConnector, public bpf_tools::BCCWrappe
   FRIEND_TEST(SocketTraceConnectorTest, ConnectionCleanupInactiveDead);
   FRIEND_TEST(SocketTraceConnectorTest, ConnectionCleanupInactiveAlive);
   FRIEND_TEST(SocketTraceConnectorTest, ConnectionCleanupNoProtocol);
+  FRIEND_TEST(SocketTraceConnectorTest, MuxBasic);
```
oazizi000, line 346: I think we can remove this now.

ddelnano: Oops, this was leftover from before. Removed in the latest revision.

src/stirling/source_connectors/socket_tracer/mux_table.h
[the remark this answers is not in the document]
ddelnano: Removed `resp_type` in favor of `req_type`

src/stirling/source_connectors/socket_tracer/mux_table.h
ddelnano: Sounds good :+1:. This is now removed from the mux table.

src/stirling/source_connectors/socket_tracer/mux_table.h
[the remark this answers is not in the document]
ddelnano: I expected this would require some discussion :) and good point regarding the ConnTracker 1:1 mapping. For now I've added a TODO comment explaining the future work.

ddelnano:
@chengruizhe I removed the part of the mux protocol inference that was misclassifying postgres traffic.

oazizi000:
@ddelnano 

https://github.com/pixie-io/pixie/commit/95199dfcc2f2f6615d758ada086a0d46fab66a61 pulled in most of these changes.

The protocol inference code was omitted as it is caused us to breach the eBPF instruction limit. This also means that for now, the Mux protocol support is not yet active.

oazizi000:
Marking this as closed since the majority has been pulled in. The last piece (eBPF mux protocol inference) will be submitted as a separate PR.

PR #379, opened 2021-12-26
fix(logical_planner_test.cc): healthz filter

lloydchang:
The standard health endpoint for Kubernetes apps is `/healthz` (rather than `/health`).

See https://github.com/pixie-io/pixie/commit/588f2541d7f4bc901cb39a1f5009941d09025fb0

Signed-off-by: lloydchang <lloydchang@gmail.com>


philkuz:
The file you've updated is a test of the planner so the exact string does not matter, it never gets executed. 

Thanks regardless for taking a look. If you want to update the scripts that power the dashboards, they're located in this directory: https://github.com/pixie-io/pixie/tree/main/src/pxl_scripts

It looks like those that care already point to `healthz`.

Issue #407, opened 2022-02-22
Add TLS tracing support to netty (portion of JVM based languages)


ddelnano:
**Is your feature request related to a problem? Please describe.**

Projects like [finagle](https://github.com/twitter/finagle) uses netty for TLS (specifically [netty-tcnative](https://github.com/netty/netty-tcnative) in finagle's case). Netty is a popular open source library so this should allow TLS tracing for any JVM based application using netty-tcnative.

**Describe the solution you'd like**

* netty TLS tracing should allow tracing encrypted mux connections
* netty TLS tracing should have feature parity with the existing openssl tracing

I'm happy to contribute my own time for this solution so I'm looking for any guidance on the "Additional context" section below (assuming the Pixie team is happy to accept this contribution).

**Describe alternatives you've considered**
N/A

**Additional context**

[netty-tcnative](https://github.com/netty/netty-tcnative) has a few options for how it implements its tls: [statically linked boringssl](https://github.com/netty/netty-tcnative/tree/main/boringssl-static), [statically linked openssl](https://github.com/netty/netty-tcnative/tree/main/openssl-static) or a [dynamically linked openssl](https://github.com/netty/netty-tcnative/tree/main/openssl-dynamic). The pom.xml files in those directories explain how the compilation happens in more detail.

For the services I'm interested in instrumenting, they use the statically linked boringssl option. These services work by depending on finagle-netty4, which ultimately uses the [aforementioned option](https://github.com/twitter/finagle/blob/7568a22c18b5e4c10985bcd985ad2ef5fc1aee88/finagle-netty4/src/test/scala/BUILD#L17). This was determined by inspecting the main application jar's Manifest file:

```
grep netty  META-INF/MANIFEST.MF  | grep native
  libs/3rdparty.jvm.io.netty.netty4-tcnative-boringssl-static-0.jar li
```

Looking inside the `libs/3rdparty.jvm.io.netty.netty4-tcnative-boringssl-static-0.jar` jar, you will find that it has it's own shared library file. This shared object contains the expected `SSL_write` and `SSL_read` symbols used by Pixie's openssl tracing:

```
$ objdump --syms META-INF/native/libnetty_tcnative_linux_x86_64.so | grep SSL_write
00000000000249b0 l     F .text  0000000000000040              netty_internal_tcnative_SSL_writeToSSL
0000000000034920 l     F .text  000000000000015a              SSL_write
[...]

PR #419, opened 2022-03-31
Enhance openssl tracing to fallback to function pointer addrs when dlsym fails

ddelnano:
This is the next step for #407.

The netty-tcnative shared library does not use a dynamic `OpenSSL_num_version` symbol. This means the symbol ends up in the symtab rather than dynsym. The former is private and incompatible with dlsym, while the latter is public and is supported by the existing openssl tracing. See this [slack thread](https://pixie-community.slack.com/archives/C03461U99S6/p1646408481645759) for more background.

This PR needs some work, but it's far enough along that I wanted  to seek feedback on my approach

## Testing
- The `openssl_trace_bpf_test` tests cases pass
  - [x] nginx openssl 1.1.1
  - [x] nginx openssl 1.1.0
  - [x] Node 12.3.1
  - [x] Node 14.18.1

Assuming this approach is how we want to proceed, the following will need to be addressed

## Todo
- [x] Parameterized `openssl_trace_bpf_test` to test the `dlsym` and `RawFptrManager` implementations of openssl tracing
- [x] ~Write unit tests for `RawFptrManager`~ -- I think our bpf test is sufficient for now and the scaffolding required to test this would be quite a bit of work so I'm holding off on this for now
- [x] ~Clean up the code added to ProcParser (much of it is duplicated with another function)~ -- not sure this is necessary
- [x] ~Make `ProcessMap` more efficient (see TODO comment about using octal numbers)~ -- wasn't sure if doing so would be more efficient. Open to any feedback you have here
- [x] Fix remaining `openssl_trace_bpf_test` test cases. They currently cause seg faults. Each test case passes individually, but when more than one case is enabled it results in seg faults ([pastebin](https://pastebin.com/Bju3TcSB))


src/stirling/obj_tools/elf_reader.cc
```diff
@@ -167,7 +167,7 @@ StatusOr<std::unique_ptr<ElfReader>> ElfReader::Create(
 
   elf_reader->binary_path_ = binary_path;
 
-  if (!elf_reader->elf_reader_.load(binary_path, /* skip_segments */ true)) {
+  if (!elf_reader->elf_reader_.load(binary_path, /* skip_segments */ false)) {
```
ddelnano, line 170: Were the segments skipped intentionally or was this not a need before?

src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.h
```diff
@@ -19,14 +19,37 @@
...
+    explicit RawFptrManager(obj_tools::ElfReader* elf_reader, system::ProcParser* proc_parser, std::string lib_path);
+    ~RawFptrManager();
+
+    template <class T>
+    StatusOr<T*> RawSymbolToFptr(const std::string& symbol_name);
+    StatusOr<bool> Init();
+
+  private:
+    obj_tools::ElfReader* elf_reader_;
+    ProcParser* proc_parser_;
```
ddelnano, line 47: Please let me know if how I'm handling these pointers is dangerous. I'm assuming that since the caller is using a `std::unique_ptr` that I won't have to assume any responsibility of managing its lifecycle. This is an area where I'm not confident what the correct thing to do is.

yzhao1012: Should be fine.

We have a few other places use ElfReader* ProcParser* in function as input argument to be used inside.

However, based on what I read, I think a cleaner way is to let RawFptrManager's ctor accept ProcessMap, and instead letter caller of RawFptrManager to call ProcParser to get the ProcessMap. That should not affect the overall flow of RawFptrManager.

WDYT?

ddelnano: I actually considered this, but I implemented it this way for the following reasons.

When `dlsym` works, we won't perform any additional work (aside from constructing the object). Doing it as you described means we will parse `/proc/$PID/maps` and `dlopen` the library an extra time, regardless if we need to fall back on this implementation. We could defer the `ProcParser::GetExecutableMapEntry` call until we need it, but then the `ProcParser` will need to be passed much further in the `uprobe_symaddrs` code. So I thought this separation was cleaner in order to defer retrieving the ProcessMap.

In addition, if the RawFptrManager's ctor accepted a ProcessMap it means that the caller could construct an object with a mismatched `lib_path` and `process_map`. See the example below for what I mean

```
# ctr definition
RawFptrManager(obj_tools::ElfReader* elf_reader, ProcessMap process_map, std::string lib_path);

std::string libpath = " ... ";
auto map = proc_parser_->GetExecutableMapEntry(pid, "not the lib path above");

// dlopen will get called on an incorrect binary
RawFptrManager(elf_reader, map, libpath);
```

Let me know what you think on that justification.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
```diff
@@ -31,12 +31,20 @@
...
 #include "src/stirling/testing/common.h"
+#include "src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.h"
 
 namespace px {
 namespace stirling {
 
 namespace http = protocols::http;
 
+bool Init() {
+  FLAGS_openssl_force_raw_fptrs = true;
```
ddelnano, line 42: Ideally the test suite would be able to run both cases for all of the tests. Any suggestions on how that could be accomplished?

src/common/system/proc_parser.h
```diff
@@ -330,6 +330,34 @@ class ProcParser {
...
+        return H::combine(std::move(h), e.vmem_start, e.vmem_end, e.file_offset);
+      }
+
+      bool operator==(const ProcParser::ProcessMap& rhs) const {
+        /* return this->vmem_start == rhs.vmem_start && this->vmem_end == rhs.vmem_end && this->permissions == rhs.permissions && this->inode == rhs.inode && this->file_offset == rhs.file_offset && this->map_path == rhs.map_path; */
+        return this->vmem_start == rhs.vmem_start && this->vmem_end == rhs.vmem_end && this->permissions.compare(permissions) == 0 && this->inode == rhs.inode && this->file_offset == rhs.file_offset && this->map_path == rhs.map_path;
+      }
+  };
+
+  StatusOr<absl::flat_hash_set<ProcessMap>> GetMapEntries(pid_t pid, std::string libpath) const;
```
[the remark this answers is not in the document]
ddelnano: This is resolved in a later commit.

src/common/system/proc_parser.h
```diff
@@ -330,6 +330,34 @@ class ProcParser {
...
+      uint64_t inode = 0;
+      std::string map_path = "";
+
+      template <typename H>
+      friend H AbslHashValue(H h, const ProcessMap& e) {
+        return H::combine(std::move(h), e.vmem_start, e.vmem_end, e.file_offset);
+      }
+
+      bool operator==(const ProcParser::ProcessMap& rhs) const {
+        /* return this->vmem_start == rhs.vmem_start && this->vmem_end == rhs.vmem_end && this->permissions == rhs.permissions && this->inode == rhs.inode && this->file_offset == rhs.file_offset && this->map_path == rhs.map_path; */
```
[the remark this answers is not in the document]
ddelnano: Done, this was left over while I was working on fixing the more pressing segfault issues

src/common/system/proc_parser.h
```diff
@@ -330,6 +330,34 @@ class ProcParser {
...
+   */
+  struct ProcessMap {
+      uint64_t vmem_start = 0;
+      uint64_t vmem_end = 0;
+      // TODO(ddelnano): Change this to an octal like number like chmod
+      /* uint8_t permissions = 0; */
+      std::string permissions = "";
+      uint64_t file_offset = 0;
+      uint64_t inode = 0;
+      std::string map_path = "";
```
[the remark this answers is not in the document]
ddelnano: Fixed

src/common/system/proc_parser.h
```diff
@@ -330,6 +330,34 @@ class ProcParser {
...
 
+  /**
+   * ProcessMaps tracks details from /proc/<pid>/maps
+   */
+  struct ProcessMap {
+      uint64_t vmem_start = 0;
+      uint64_t vmem_end = 0;
+      // TODO(ddelnano): Change this to an octal like number like chmod
+      /* uint8_t permissions = 0; */
+      std::string permissions = "";
```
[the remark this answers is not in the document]
ddelnano: I've removed the `= "";` and this TODO since I wasn't sure if changing the `GetMapEntries` code would be more performant if it was backed by an 8 bit field.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
oazizi000: @ddelnano, a couple of pointers:
1) You can use `INSTANTIATE_TEST_SUITE_P` which is part of gtest to pass in parameters to a test. You can find plenty of examples in Stirling if you grep around.
2) You can use PL_SET_FOR_SCOPE for assigning flags in tests. Since flags are global, it helps to avoid unintended consequences when changing a flag (i.e. so the flag only applies for the test case in questions).

src/stirling/obj_tools/elf_reader.cc
oazizi000: May have been unintentional. Can you say more about what you're looking for in he segments?

ddelnano: Since segments are what end up memory mapped, we need to find the segment offset where the `.text` section is located to find the function's memory address.

Here is an example `libcrypto.so.1.1` file and the corresponding segments:

```
vagrant@vagrant:/vagrant$ readelf -lW libcrypto.so.1.1

Elf file type is DYN (Shared object file)
Entry point 0x79000
There are 9 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x077930 0x077930 R   0x1000
  LOAD           0x078000 0x0000000000078000 0x0000000000078000 0x19a80d 0x19a80d R E 0x1000
  LOAD           0x213000 0x0000000000213000 0x0000000000213000 0x090420 0x090420 R   0x1000
  LOAD           0x2a3ab0 0x00000000002a4ab0 0x00000000002a4ab0 0x02d0c0 0x0314f0 RW  0x1000
  DYNAMIC        0x2ce4c8 0x00000000002cf4c8 0x00000000002cf4c8 0x000230 0x000230 RW  0x8
  NOTE           0x000238 0x0000000000000238 0x0000000000000238 0x000024 0x000024 R   0x4
  GNU_EH_FRAME   0x256c58 0x0000000000256c58 0x0000000000256c58 0x00c904 0x00c904 R   0x4
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x10
  GNU_RELRO      0x2a3ab0 0x00000000002a4ab0 0x00000000002a4ab0 0x02b550 0x02b550 R   0x1

 Section to Segment mapping:
  Segment Sections...
   00     .note.gnu.build-id .gnu.hash .dynsym .dynstr .gnu.version .gnu.version_d .gnu.version_r .rela.dyn .rela.plt
   01     .init .plt .plt.got .text .fini
   02     .rodata .eh_frame_hdr .eh_frame
   03     .init_array .fini_array .data.rel.ro .dynamic .got .data .bss
   04     .dynamic
   05     .note.gnu.build-id
   06     .eh_frame_hdr
   07
   08     .init_array .fini_array .data.rel.ro .dynamic .got
```

The 0x078000 offset that corresponds to segment 01 needs to be determined in order to find a function's memory address properly. You can see the full calculation [here](https://github.com/pixie-io/pixie/pull/419/files#diff-e26509bdd023bc2adee54cfb2c8c0c039400d90c6c8e7a2a12edf454278d36ecR518-R521)).

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
ddelnano: 1. I did notice the `INSTANTIATE_TEST_SUITE_P` examples, but the openssl_trace_bpf_test code uses `TYPED_TEST`s. I wasn't sure how to incorporate the parameter aspects when typed tests are in use
2. Thanks for that tip. `PL_SET_FOR_SCOPE` is exactly what I was looking for.

src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.h
ddelnano: @yzhao1012 just checking in to see if you saw my previous comment.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
ddelnano: This [Stackoverflow post](type-parameterized tests and value-parameterized tests) post mentions that the two features are not complementary and links to a google group [discussion](https://groups.google.com/g/googletestframework/c/3d8H5J5Isnw) with the maintainers. And given the copy/move issues with the test cases due to the underlying ContainerRunner instances, I think we make the typed tests work.

I've crated a simpler example (copied below). I need to figure out how to declare each `TYPED_TEST` multiple times still, which is what I'm trying to accomplish with the `OPENSSL_TYPED_TEST` macro. Please let me know your thoughts on that approach or any other ideas you have.

```
#include <gtest/gtest.h>

class ContainerRunner {
  public:
    void Run() {
    }
};

class ContainerOne : public ContainerRunner { };
class ContainerTwo : public ContainerRunner { };

template <typename TServerContainer, bool TForceFptr>
class BaseOpenSSLTraceTest : public ::testing::Test {
  public:
    BaseOpenSSLTraceTest() {
      server_.Run();
    }
    TServerContainer server_;
    bool force_fptr_ = TForceFptr;

};

template <typename T>
using OpenSSLTraceTest = BaseOpenSSLTraceTest<T, false>;

typedef ::testing::Types<ContainerOne, ContainerTwo>
  OpenSSLServerImplementations;

// Old way that works
TYPED_TEST_SUITE(OpenSSLTraceTest, OpenSSLServerImplementations);

#define OPENSSL_TYPED_TEST(TestName, TestCase, CodeBlock) TYPED_TEST(TestName, TestCase) CodeBlock

OPENSSL_TYPED_TEST(OpenSSLTraceTest, TestOne, {
  printf("TestOne test");
})

```

ddelnano: @oazizi000 please see how I implemented this in https://github.com/pixie-io/pixie/pull/419/commits/6f0fbf0f127fcd0fb854a015bd5f5cb669e8cd4d

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
```diff
@@ -90,6 +91,19 @@ class OpenSSLTraceTest : public SocketTraceBPFTest</* TClientSideTracing */ fals
     sleep(1);
   }
 
+  void SetUp() override {
+    original_openssl_force_raw_fptrs_ = FLAGS_openssl_force_raw_fptrs;
```
ddelnano, line 95: Since this value is read during `SocketTraceBPFTest::SetUp()`, we can't use `PL_SET_FOR_SCOPE` as discussed [here](https://github.com/pixie-io/pixie/pull/419#discussion_r840967938).

src/common/system/proc_parser.cc
```diff
@@ -774,5 +774,77 @@ StatusOr<absl::flat_hash_set<std::string>> ProcParser::GetMapPaths(pid_t pid) co
...
+StatusOr<absl::flat_hash_set<ProcParser::ProcessMap>> ProcParser::GetMapEntries(
+    pid_t pid, std::string libpath) const {
+  absl::flat_hash_set<ProcParser::ProcessMap> map_entries;
+
+  static constexpr int kProcMapNumFields = 6;
+
+  const std::filesystem::path proc_pid_maps_path = ProcPidPath(pid) / "maps";
+  PL_ASSIGN_OR_RETURN(std::string content, px::ReadFileToString(proc_pid_maps_path));
+  std::ifstream ifs;
+  ifs.open(proc_pid_maps_path);
```
oazizi000, line 786: Looks like we're both using `ReadFileToString` and `ifs.open()`. Isn't that redundant?

src/common/system/proc_parser.cc
```diff
@@ -774,5 +774,77 @@ StatusOr<absl::flat_hash_set<std::string>> ProcParser::GetMapPaths(pid_t pid) co
...
+    // 55e816b37000-55e816b65000 r--p  00000000 103:02 55579316          /usr/bin/vim.basic
+    // We differentiate these headers from the subsequent key-value pairs that include
+    // data about the memory usage for each of the process mappings.
+    auto idx = line.find(':');
+    if (idx == std::string::npos) {
+      continue;
+    }
+    // If the character after the colon is whitespace, this is a key-value line.
+    // Else the colon is part of the device (major:minor) and this is a header.
+    // Perhaps we should look for other indicators?
```
oazizi000, line 805: This looks like code we have in `ParseProcPIDSMaps()`. Seems like an opportunity to refactor and share?

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
oazizi000: Makes sense

src/stirling/obj_tools/elf_reader.cc
oazizi000: Nice. Thanks for the explanation. :)

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
oazizi000: On second thought, since BaseOpenSSLTraceTest is the only test class in this test, you don't really need to reset the value on the destructor, because you always explicitly set the value during SetUp(), which is safe.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
```diff
@@ -151,9 +167,22 @@ http::Record GetExpectedHTTPRecord() {
...
+
+#define OPENSSL_TYPED_TEST(TestCase, CodeBlock)            \
+  TYPED_TEST(OpenSSLTraceDlsymTest, TestCase)              \
+  CodeBlock TYPED_TEST(OpenSSLTraceRawFptrsTest, TestCase) \
+  CodeBlock
+
+TYPED_TEST_SUITE(OpenSSLTraceDlsymTest, OpenSSLServerImplementations);
+TYPED_TEST_SUITE(OpenSSLTraceRawFptrsTest, OpenSSLServerImplementations);
+
+OPENSSL_TYPED_TEST(ssl_capture_curl_client, {
```
oazizi000, line 185: Clever. Though also wonder if there's a better way. Apparently, we can't mix typed parameters with value parameters in gtest, which is unfortunate. So this might be the best way.

In theory, you might be able to turn the second value as a second type parameters, such that we `OpenSSLServerImplementations` gets converted to a sequence of `std::pairs`(like https://stackoverflow.com/questions/52256953/c-multiple-parameters-with-gtest-typed-test).

You can feel free to leave this as is, though, since it may not be worth the effort to change this.

oazizi000 (commented):
This looks great @ddelnano. A few final comments, but overall, I have no objections. Hoping we can get this in soon.

Sorry for the tardy reviews on this!

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
ddelnano: I tried something similar to that, but ran into difficulties so I'd prefer to leave this as is.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
ddelnano: But isn't it possible that any other 'trace bpf' tests that use uprobes could run later and they would have whatever the last value that was set be enabled? I thought that had the potential to pollute future tests, so I erred on the side of caution.

src/common/system/proc_parser.cc
ddelnano: This was refactored in a032909130d1246d66d6fbb1adadabc6ee7b277e. I enhanced `ProcParser::ProcessSMaps` to work for my use case and consolidated the proc_parser function to work for both.

ddelnano:
@oazizi000 the most pressing issue (refactoring the proc_parser code) is complete. The rest I was going to hold off on. Let me know what you think.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
oazizi000: There's no risk of that because each test is compiled into its own binary. So tests can't interfere with each other. The only risk would be if a new test class was added to `openssl_trace_bpf_test.cc`.

Anyways, not a big deal either way.

oazizi000 (approved):
Left a comment, but approved! Will look into pulling this in when I get a chance. As part of that, we'll run some perf regression tests to make sure it doesn't break anything performance-wise.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
ddelnano: ah, I see. Since this is an easy fix, I've pushed another commit and retested to remove the `TearDown` method. Thanks for explaining that!

oazizi000:
@ddelnano Can you rebase your changes onto main and squash them all into one commit? Trying to merge this in, but hitting merge conflicts.

ddelnano:
I'm not confident that squashing and rebasing this branch (since I merged main in previously) would work well. So I opened up #428 with this branch's changes.

oazizi000:
Replaced by #428.

PR #427, opened 2022-04-27
Add functions to create UPIDs from their underlying components.

### LINKED ISSUE #405: Add PxL UDFs for standard math and string operations
PxL is used to process and transform data collected by Pixie. In its current state, we have already have many functions implemented for most basic data processing (see: https://docs.px.dev/reference/pxl/)
However, we are still missing functions that users may want to execute on their data. In particular, this includes math and string operations such as itoa/atoi, string replace, and more.

sethtroisi:
Partially fixes #405, by implementing functions to create UPIDs.

added unit tests.


zasgar:
Thanks for the contribution @sethtroisi. We require a DCO for out commits, can you sign-off the commit using the process here: https://github.com/pixie-io/pixie/blob/main/CONTRIBUTING.md#sign-your-commits.

sethtroisi:
Thanks for the pointer @zasgar, fixed now.

zasgar (approved):

PR #520, opened 2022-07-12
gRPC-C bugfix / Added explicit writing of null terminators to grpc-c metadata

avivzgroundcover:
Current implementation of the grpc_c probes doesn't zero-out the metadata buffers (HTTP2 headers) between traces. Therefore, old header data could be wrongly appended to new header data if the old header was longer than the new one. 

To solve this, we explicitly add a null terminator at the end of the new header being read, so that the leftovers from previous headers are ignored.


zasgar:
ok to test

vihangm:
Merged as 0bb76017a54b31e68164d766b08d9e3549c03cf7

PR #535, opened 2022-07-14
Fix data-table bug with anchors

s-starostin:
Long values are hidden in data-tables:
image (3)
It can be fixed if anchors in cells will have "white-space: nowrap" property set:

### DIFF
--- a/src/ui/src/components/data-table/data-table.tsx
+++ b/src/ui/src/components/data-table/data-table.tsx
@@ -148,6 +148,9 @@ const useDataTableStyles = makeStyles((theme: Theme) => createStyles({
     maxHeight: '100%',
     width: '100%',
     lineHeight: `${ROW_HEIGHT_PX}px`,
+    '& a': {
+      whiteSpace: 'nowrap', 
+    },
   },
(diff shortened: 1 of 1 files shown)


zasgar:
Ok to test

src/ui/src/components/data-table/data-table.tsx
```diff
@@ -148,6 +148,9 @@ const useDataTableStyles = makeStyles((theme: Theme) => createStyles({
     maxHeight: '100%',
     width: '100%',
     lineHeight: `${ROW_HEIGHT_PX}px`,
+    '& a': {
+      whiteSpace: 'nowrap', 
+    },
```
NickLanam, line 153: I think setting the `pre` above to `nowrap` will work well here too - it's not just links that are affected.

Good catch, and thanks for the contribution!

NickLanam (changes requested):

src/ui/src/components/data-table/data-table.tsx
s-starostin: Hello, ok, I've pushed changes.

Issue #548, opened 2022-08-01
Python package pins the `pip` utility


Lawouach:
**Describe the bug**

Currently, the Python api package pins the `pip` utility in its requirements:

https://github.com/pixie-io/pixie/blob/a20c2ec5d46ff715f99ecb25613618386e923b25/src/api/python/requirements.txt#L7

This is not really convenient as `pip` is often updated and kept to its latest version. Distributing anything relying pixie will create some friction.

**Expected behavior**

Maybe you can be more forgiving and pin to the major version only?

```
pip~=22.0
```

Lawouach:
Looking at the code, `pip` seems to only be used to package the distribution, not in the actual code. Can we remove that dependency altogether?

htroisi:
@Lawouach I don't think we need to pin pip. This was probably added accidentally when we froze the requirements. 

Would you be willing to submit a PR? Feel free to review the other requirements too. Thanks!

PR #549, opened 2022-08-01
Fix setup

Lawouach:
A few minor changes.

I'll let you make a decision if pinning to major versions is sound enough or if you need specific versions.

### DIFF
--- a/src/api/python/BUILD.bazel
+++ b/src/api/python/BUILD.bazel
@@ -24,6 +24,8 @@ py_wheel(
     classifiers = [
         "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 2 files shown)


htroisi:
Fixes https://github.com/pixie-io/pixie/issues/548.

Lawouach:
I think we should then distinguish setup requirements from install requirements. 

The former should be part of the `setup_requires` field in the `setup` fonction. This allows you to define such dependencies. Note this particular bit has actually moved to the `pyproject.toml` file nowadays:

https://setuptools.pypa.io/en/latest/userguide/dependency_management.html

Regarding the strict pinning, I can appreciate that. I'll revert back my commits.

Lawouach:
I've removed my commits.

vihangm (approved):

PR #555, opened 2022-08-04
Fix UUID serialization in Python API

### LINKED ISSUE #551: Python API fails to encode row when data is a UUID
**Describe the bug**

The default json encoder of the Python stdlib doesn't know how to serialize UUID objects. It therefore fails with:

```
TypeError: Object of type UUID is not JSON serializable
```

**To Reproduce**

Run this script with the Python client:

```import px

px.display(px.GetAgentStatus())
```

iterating over the rows and printing the rows will lead to the error.

**Expected behavior**

Pixie should provide a more capable json encoder to handle these base types.

brandonwnguyen:
Fixes #551 

Test plan: 
- Add an example which fetches agent data with UUIDs

### DIFF
--- a/src/api/python/examples/script_with_uuids.py
+++ b/src/api/python/examples/script_with_uuids.py
@@ -1,0 +1,41 @@
+# Copyright 2018- The Pixie Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


zasgar:
ok to test

zasgar:
Thanks @brandonwnguyen!

PR #557, opened 2022-08-05
Fix CLI so that it does not show results for random cluster if the targeted cluster is not healthy.

prasoonblueluck:
Currently `px run`, `px live` and `px get pems` will fall back to using the the "first other known healthy cluster" if the targeted cluster is not healthy and the user has Pixie deployed to multiple clusters. Most users will probably expect to get results for the cluster in their kubeconfig and may accidentally attribute the displayed data to the targeted cluster.

This fixes https://github.com/pixie-io/pixie/issues/527 by removing most of the calls to `GetCurrentOrFirstHealthyVizier()` and instead using `GetCurrentVizier()`. Note that there is one remaining call to `GetCurrentOrFirstHealthyVizier()` for the `px update` CLI command, since I'm less familiar with how to test that. I also updated the error to make it a little more clear how the problem should be fixed. 

Test plan: Check that `px get pems`, `px run px/cluster` and `px live px/cluster` no longer return results if kubeconfig points to a cluster without Pixie installed.

### DIFF
--- a/src/pixie_cli/pkg/cmd/get.go
+++ b/src/pixie_cli/pkg/cmd/get.go
@@ -73,7 +73,7 @@ var GetPEMsCmd = &cobra.Command{
 		clusterID := uuid.FromStringOrNil(selectedCluster)
 		var err error
 		if !allClusters && clusterID == uuid.Nil {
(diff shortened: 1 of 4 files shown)


zasgar:
Ok to test

zasgar:
Thanks for the contribution! 🚀

Issue #537, opened 2022-07-19
UI should not try to get an export status on Disabled scripts


philkuz:
**Is your feature request related to a problem? Please describe.**
On https://work.dev.withpixie.dev/configure-data-export/
Users reading the export script view will find that the status icon to be in an unknown state - signaling a system error. In reality, we're just not querying the state of the plugin script because it is disabled. 
<img width="644" alt="image" src="https://user-images.githubusercontent.com/5456019/179801973-9172d02a-7ca6-4938-85c5-a84e70585403.png">

**Describe the solution you'd like**
Instead, we should change the message to say "Script is disabled"

PR #558, opened 2022-08-05
GH-537: Fix UI showing an export status on Disabled scripts

phillip-kravtsov:
Signed-off-by: Phillip Kravtsov <phillip@trucksmarter.com>


vihangm:
ok to test

vihangm (approved):

Issue #590, opened 2022-09-13
server_names_hash_bucket_size too low


Sindvero:
**Describe the bug**
When deploying self-hosted pixie and if you set a very long domain name, you get the following error in the cloud-proxy-server container:

```
[emerg] 8#8: could not build server_names_hash, you should increase server_names_hash_bucket_size: 64
nginx: [emerg] could not build server_names_hash, you should increase server_names_hash_bucket_size: 64
```

**To Reproduce**
Set PL_DOMAIN_NAME to a very long domain name (e.g. very.very.long.domain.name.used.in.this.example.org) and see cloud-proxy-server container crash.

**Expected behavior**
It doesn't crash

**Screenshots**
If applicable, add screenshots to help explain your problem. Please make sure the screenshot does not contain any sensitive information such as API keys or access tokens.

**Logs**
[emerg] 8#8: could not build server_names_hash, you should increase server_names_hash_bucket_size: 64
nginx: [emerg] could not build server_names_hash, you should increase server_names_hash_bucket_size: 64

**Additional context**
I tried to set the env var SERVER_NAMES_HASH_BUCKET_SIZE to 128, but the env var didn't get apply to nginx.

Sindvero:
The workaround is to add this line "server_names_hash_bucket_size 128;" to the http context of the nginx.conf file then mount it to the cloud proxy pod. 

But what do you think about having it directly to the file, like that people using long domain name won't have to workaround like that, I can PR the changes if you want? or create a new file just for it and include it in the nginx config?

htroisi:
@Sindvero - thanks for reporting back with the solution. We'd be happy to accept a PR with this modification!

Sindvero:
@htroisi - Here's the PR #596

zasgar:
Thanks for the contribution @Sindvero!

zasgar:
Thanks for identifying the fix. Can you make a PR it?

PR #596, opened 2022-09-16
Increase the default server names length in nginx config file

Sindvero:
It's going to increase the default server names length in the nginx config to avoid the following error:

```
[emerg] 8#8: could not build server_names_hash, you should increase server_names_hash_bucket_size: 64
nginx: [emerg] could not build server_names_hash, you should increase server_names_hash_bucket_size: 64
```

That will allow people with very long domain names (_e.g._ [very.very.long.domain.name.used.in.this.example.org](http://very.very.long.domain.name.used.in.this.example.org/)) to make it work out of the box without having to mount the nginx.conf file in the pod.

Issue #590
And slack discussion: https://pixie-community.slack.com/archives/CQ63KEVFY/p1663087667528489

Reference to change this value:
- https://sleeplessbeastie.eu/2019/11/18/how-to-increase-the-default-number-of-maximum-server-names-and-their-length-when-using-nginx/
- https://gist.github.com/muhammadghazali/6c2b8c80d5528e3118613746e0041263


zasgar:
ok to test

src/cloud/proxy/nginx.conf
```diff
@@ -39,6 +39,9 @@ http {
     # Tune nginx keepalives to work with the GCP HTTP(S) Load Balancer:
     keepalive_timeout 650;
     keepalive_requests 10000;
+  
+    # Increase default server names lenght to 128:
```
zasgar, line 43: nit: can you fix the typo lenght -> "length"? It's triggering one of our lint rules. Thanks!

zasgar (changes requested):

src/cloud/proxy/nginx.conf
Sindvero: Fixed it. And looks like your Jenkins build also passed

zasgar:
ok to test

zasgar (approved):

PR #607, opened 2022-10-03
Add finagle demo to pixie cli

ddelnano:
Once #569 is merged, mux tracing will be enabled by default (on kernels >= 5.2). This unblocks adding a finagle demo to the pixie cli. I thought this would be a good addition especially with the eBPF talk later this month.

As part of Twitter's hackathon week, we created [sedarG/finagle-helloworld](https://github.com/sedarG/finagle-helloworld) to demonstrate Pixie's mux protocol tracing. I've modified that repo to match the same setup as the [pixie-labs/microservice-kafka](https://github.com/pixie-labs/microservice-kafka) repo (using kompose, kustomize, etc).

## Testing
- [x] docker-compose example works locally
- [x] Verify that the client and server work on minikube
 1. Set docker client to use minikube cluster -- `eval $(minikube docker-env)`
 2. Build docker image from [sedarG/finagle-helloworld](https://github.com/sedarG/finagle-helloworld) -- `cd finagle-helloworld; docker build -t finagle-server .`
 4. Create `finagle-client` docker image -- `docker tag finagle-server finagle-client`
 5. Update k8s manifest to use `finagle-client` and `finagle-server` as the images and deploy -- `kubectl apply -f demos/finagle/finagle.yaml`
 6. Verify that the finagle-client container logs show successful requests

```
[tw-mbp-ddelnano pixie (ddelnano/add-finagle-demo)]% docker ps | grep finagle
612048c824dd   53b86303c71e                                          "sbt client/run"         2 minutes ago       Up About a minute             k8s_finagle-client_finagle-client-76f887b768-vg2s6_px-finagle_89b131c1-899c-4f98-9d8a-b03235544d40_1
8919609f58de   53b86303c71e                                          "sbt server/run"         2 minutes ago       Up 2 minutes                  k8s_finagle-server_finagle-server-77b8865ccd-c8gwn_px-finagle_d25e033c-eb50-40e0-bf00-e8acfa97c7a3_0
c0a685805ba2   k8s.gcr.io/pause:3.6                                  "/pause"                 2 minutes ago       Up 2 minutes                  k8s_POD_finagle-server-77b8865ccd-c8gwn_px-finagle_d25e033c-eb50-40e0-bf00-e8acfa97c7a3_0
[...]

PR #609, opened 2022-10-04
Add mux_frame_type_name UDF to ease developing mux pxl scripts

ddelnano:
This adds a `mux_frame_type_name` UDF function that will be used in #607. It seems there is prior art for mysql, amqp and other protocols. See this discussion [here](https://github.com/pixie-io/pixie/pull/607#discussion_r986171497) for more details.

## Testing done
- [x] New unit tests pass
```
vagrant@ubuntu-jammy:/vagrant$ bazel test src/carnot/funcs/protocols:all
INFO: Analyzed 3 targets (0 packages loaded, 0 targets configured).
INFO: Found 2 targets and 1 test target...
INFO: Elapsed time: 7.737s, Critical Path: 6.71s
INFO: 5 processes: 1 internal, 4 linux-sandbox.
INFO: Build completed successfully, 5 total actions
//src/carnot/funcs/protocols:protocol_ops_test                           PASSED in 0.1s

INFO: Build completed successfully, 5 total actions
```

### DIFF
--- a/src/carnot/funcs/protocols/mux.h
+++ b/src/carnot/funcs/protocols/mux.h
@@ -60,7 +60,7 @@ inline std::string FrameTypeName(int frame_type) {
     case -62:
       return "Tdiscarded (legacy)";
     default:
-      return std::to_string(frame_type);
+      return absl::Substitute("Unknown ($0)", frame_type);
   }
 }
 
--- a/src/carnot/funcs/protocols/mux.h
+++ b/src/carnot/funcs/protocols/mux.h
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2018- The Pixie Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <string>
+
+namespace px {
+namespace carnot {
+namespace funcs {
+namespace protocols {
+namespace mux {
+
+inline std::string FrameTypeName(int frame_type) {
+  switch (frame_type) {
+    case 1:
+      return "Treq";
+    case -1:
+      return "Rreq";
+    case 2:
+      return "Tdispatch";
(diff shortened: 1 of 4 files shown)


src/carnot/funcs/protocols/mux.h
```diff
@@ -0,0 +1,72 @@
...
+    case 68:
+      return "Tinit";
+    case -68:
+      return "Rinit";
+    case -128:
+      return "Rerr";
+    case 127:
+      return "Rerr (legacy)";
+    case -62:
+      return "Tdiscarded (legacy)";
```
ddelnano, line 61: These are the same as the enum class used within mux's parsing code ([source](https://github.com/pixie-io/pixie/blob/ea7c6bc552fcd0f2808e3e9d31647ab49bfb4aa9/src/stirling/source_connectors/socket_tracer/protocols/mux/types.h#L35))

vihangm:
ok to test

src/carnot/funcs/protocols/protocol_ops.cc
```diff
@@ -69,6 +71,10 @@ types::StringValue MySQLCommandNameUDF::Exec(FunctionContext*, Int64Value api_ke
   return mysql::CommandName(api_key.val);
 }
 
+types::StringValue MuxFrameTypeUDF::Exec(FunctionContext*, Int64Value frame_type) {
+  return mux ::FrameTypeName(frame_type.val);
```
zasgar, line 75: rm space between mux and ::

src/carnot/funcs/protocols/mux.h
```diff
@@ -0,0 +1,72 @@
...
+    default:
+      return std::to_string(frame_type);
```
zasgar, line 63: Can we use the absl functions? std::string functions generally look at locale and we avoid using them in carnot. We might also want to print it as `absl::StringFormat("Unknown (%d), frame_type);`.

ddelnano: For sure. This was copy and pasted from the amqp example, but it seems a few others are using that as well.

```
[tw-mbp-ddelnano pixie (ddelnano/add-mux-carnot-func-for-mux-type-translation)]% git grep std::to_string src/carnot/funcs/protocols
src/carnot/funcs/protocols/amqp.h:      return std::to_string(frame_type);
src/carnot/funcs/protocols/kafka.h:      return std::to_string(api_key);
src/carnot/funcs/protocols/mysql.h:      return std::to_string(cmd);
```

mux's case was updated in the latest revision.

src/carnot/funcs/protocols/protocol_ops.cc
ddelnano: Fixed.

src/carnot/funcs/protocols/mux.h
oazizi000: We should fix the other ones as well, but that can be done separately.

oazizi000 (approved):

PR #624, opened 2022-10-17
Add libcrypt1 to CC base image so PEM can dlopen netty shared libraries

ddelnano:
#### What type of PR is this?

Feature -- this addresses an issue that prevents a k8s PEM from tracing netty applications using TLS.

With #407's previous PRs merged, we thought that netty TLS tracing is fully functional. It turns out there are a few remaining issues (documented [here](https://github.com/pixie-io/pixie/issues/407#issuecomment-1281240803)).

#### What is the testplan for the PR:

- Deployed pixie to minikube, enabled `FLAGS_openssl_raw_fptrs_enabled` and updated striling's [StirlingImpl::GetContext()](https://github.com/pixie-io/pixie/blob/9d0620f71729a89d25b09c36307286ec558a285a/src/stirling/stirling.cc#L422) to always return a `SystemWideStandaloneContext` object.
- Ran the `src/stirling/source_connectors/socket_tracer/testing/containers/thriftmux:server_image` image manually to generate encrypted traffic
- Used `stirling_ctrl` to enable debug logging for the server's pid
- Verified that finagle TLS tracing was working after those changes
```
I20221017 17:54:24.759876 624027 conn_tracker.h:271] conn_id=[upid=458961:777207 fd=143 gen=13416457246767] state=kTransferring remote_addr=127.0.0.1:39556 role=kRoleServer protocol=kProtocolMux ssl=true req_frames=3 resp_frames=3
I20221017 17:54:24.759899 624027 conn_tracker.h:278] conn_id=[upid=458961:777207 fd=143 gen=13416457246767] state=kTransferring remote_addr=127.0.0.1:39556 role=kRoleServer protocol=kProtocolMux ssl=true records=3
I20221017 17:54:24.966593 624027 conn_tracker.cc:150] conn_id=[upid=458961:777207 fd=143 gen=13416457246767] state=kTransferring remote_addr=127.0.0.1:39556 role=kRoleServer protocol=kProtocolMux ssl=true Data event: attr:[[ts=13416621420623 conn_id=[upid=458961:777207 fd=143 gen=13416457246767] protocol=kProtocolMux role=kRoleServer dir=kIngress ssl=true source_fn=kSSLRead pos=293 size=204 buf_size=204]] msg_size:204 msg:[\x00\x00\x00\xC8\x02\x00\x00\x02\x00\x03\x00(com.twitter.finagle.tracing.TraceContext\x00 \x8F\x95G\x88\x85z\xF4J\x8F\x95G\x88\x85z\xF4J\x8F\x95G\x88\x85z\xF4J\x00\x00\x00\x00\x00\x00\x00\x00\x00\x1Bcom.twitter.finagle.Retries\x00\x04\x00\x00\x00\x00\x00\x1Ccom.twitter.finagle.Deadline\x00\x10\x17\x1E\xEC\xCA\x0F\x11@\x80\x7F\xFF\xFF\xFF\xFF\xFF\xFF\xFF\x00\x00\x00\x00\x80\x01\x00\x01\x00\x00\x00\x05query\x00\x00\x00\x00\x0B\x00\x01\x00\x00\x00\x06String\x00]
I20221017 17:54:24.966631 624027 conn_tracker.cc:150] conn_id=[upid=458961:777207 fd=143 gen=13416457246767] state=kTransferring remote_addr=127.0.0.1:39556 role=kRoleServer protocol=kProtocolMux ssl=true Data event: attr:[[ts=13416622005181 conn_id=[upid=458961:777207 fd=143 gen=13416457246767] protocol=kProtocolMux role=kRoleServer dir=kEgress ssl=true source_fn=kSSLWrite pos=73 size=48 buf_size=48]] msg_size:48 msg:[\x00\x00\x00,\xFE\x00\x00\x02\x00\x00\x00\x80\x01\x00\x02\x00\x00\x00\x05query\x00\x00\x00\x00\x0B\x00\x00\x00\x00\x00\x0CStringString\x00]

```

I'll be opening up other PRs to address the other two issues: updating `FLAGS_openssl_raw_fptrs_enabled`'s default value and fixing the UPID discovery issue

Not until the other issues are addressed.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -108,6 +108,7 @@ buildifier(
 
 pl_cc_base_packages = [
     packages["libc6"],
+    packages["libcrypt1"],
     packages["libelf1"],
     packages["liblzma5"],
     packages["libtinfo6"],


oazizi000:
We shouldn't need to have `libcrypt1` in the PEM, since Stirling should be able to find the appropriate libraries in the target container. In fact, I'd be worried that the `libcrypt1` in the PEM is the wrong version. But obviously, there is something not quite  working yet. Do you know if netty even really needs `libcrypt1`. Is it possible this is a false dependency? Since netty is using boringssl, I would think it's not required.

ddelnano:
@oazizi000 while netty-tcnative can use boringssl, it is capable of using dynamically linked openssl as well. When I first created the thriftmux container, I ran into a linking issue before I realized that `libcrypt1` package needed to be installed ([slack](https://pixie-community.slack.com/archives/C03461U99S6/p1646092059360319?thread_ts=1646062594.533129&cid=C03461U99S6)). Using `ldd` you can see that the tcnative shared library requires `libcrypt.so.1` as well.

```
vagrant@ubuntu-jammy:/vagrant$ ldd libnetty_tcnative_linux_x86.so  | grep crypt
        libcrypt.so.1 => /lib/x86_64-linux-gnu/libcrypt.so.1 (0x00007fa2cdb7a000)
```

I assume that it should be possible to remove the share library's dependency on `libcrypto1` when using boringssl or libressl since they are both statically linked. However, it appears the shared library is compiled to work with all permutations of ssl library linkage.

I definitely agree that version mismatches could be problematic. It seems the `RawFptrManager` needs to be more namespace aware so it can find any library the traced process can load. Maybe adding the necessary container library paths to `LD_LIBRARY_PATH` or a similar mechanism would allow the RawFptrManager to load this library without any dependencies within the PEM?

After discussing this more, it definitely seems like we need a more appropriate solution so I'll close this for now.

PR #674, opened 2022-12-27
Add global.cluster recognition for clusterName

maxlemieux:
Signed-off-by: Max Lemieux <mlemieux@newrelic.com>

#### What type of PR is this?
/kind feature

Currently, when this chart is installed as a subchart, the `pixie-chart.clusterName` value must still be set, although other charts may already set a required value `global.cluster` to store the cluster name. 

As a person installing the chart bundle as a subchart, this is not a great experience as I must enter the same value twice - only for this value.

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:
This change makes it unnecessary to set `pixie-chart.clusterName` for the chart bundle, as long as `global.cluster` is set.

```release-note

```

#### Additional documentation:

```docs

```

### DIFF
--- a/k8s/operator/helm/templates/04_vizier.yaml
+++ b/k8s/operator/helm/templates/04_vizier.yaml
@@ -14,7 +14,9 @@ spec:
   cloudAddr: {{ .Values.cloudAddr }}
   disableAutoUpdate: {{ .Values.disableAutoUpdate }}
   useEtcdOperator: {{ .Values.useEtcdOperator }}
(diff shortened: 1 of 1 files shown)


aimichelle:
ok to test

maxlemieux:
Per conversation, also added `registry` check for `global.images.registry` to reduce another point of potentially duplicated configuration.

aimichelle (approved):

PR #675, opened 2022-12-28
Update internal LBs to be internal on AKS

JeffBelgum:
#### What type of PR is this?
/kind bug

The load balancer resources in AKS are external by default whereas the load balancers in GKE are internal. The correct default is internal.

This PR adds the annotation specified in [Azure's AKS docs](https://learn.microsoft.com/en-us/azure/aks/internal-lb#create-an-internal-load-balancer) so that the load balancers for the vzconn and cloud-proxy services are assigned internal IP addresses.

This now matches the behavior of the load balancers in a GKE environment.

#### What is the testplan for the PR:
Verify internal IP addresses for the vzconn and cloud-proxy services when creating a new self-hosted Pixie on AKS.

#### Which issue(s) this PR fixes:
Fixes #

```release-note
Vzconn and cloud proxy service Load Balancers are internal by default under AKS
```

#### Additional documentation:
N/A

### DIFF
--- a/k8s/cloud/base/vzconn_service.yaml
+++ b/k8s/cloud/base/vzconn_service.yaml
@@ -6,6 +6,7 @@ metadata:
   annotations:
     cloud.google.com/load-balancer-type: internal
     cloud.google.com/app-protocols: '{"tcp-http2":"HTTP2"}'
+    service.beta.kubernetes.io/azure-load-balancer-internal: "true"
 spec:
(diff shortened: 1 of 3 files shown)


philkuz:
For core-maintainers: this is basically identical to the GKE annotation to make load balancers internal.

aimichelle (approved):

aimichelle:
ok to test

PR #695, opened 2023-01-19
Add securityContext for pixie cloud workloads

MrSupiri:
Summary: This PR adds securityContext to all Kubernetes workloads with least privileges that are required to run the pixie cloud on self hosted setting. This is important in a multi-tenant Kubernetes cluster, as it allows administrators to ensure that each workload runs in a secure and isolated environment, preventing one tenant's workload from affecting or compromising the security of other tenants' workloads.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: 1. Checkout the current branch.
2. Deploy [Self hosted Pixie](https://docs.px.dev/installing-pixie/install-guides/self-hosted-pixie/).

### DIFF
--- a/k8s/cloud/base/api_deployment.yaml
+++ b/k8s/cloud/base/api_deployment.yaml
@@ -138,8 +138,8 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:
-              drop:
-                - ALL
+            drop:
+            - ALL
           runAsNonRoot: true
           runAsUser: 10100
           seccompProfile:
--- a/k8s/cloud/base/api_deployment.yaml
+++ b/k8s/cloud/base/api_deployment.yaml
@@ -135,6 +135,20 @@ spec:
           mountPath: /vizier-image-secret
         - name: elastic-certs-pub
           mountPath: /elastic-certs-pub
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+              drop:
+                - ALL
+          runAsNonRoot: true
+          runAsUser: 10100
+          seccompProfile:
+            type: RuntimeDefault
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 10100
+        seccompProfile:
+          type: RuntimeDefault
       volumes:
       - name: certs
         secret:
--- a/k8s/cloud_deps/base/nats/statefulset.yaml
+++ b/k8s/cloud_deps/base/nats/statefulset.yaml
@@ -235,16 +235,16 @@ spec:
           allowPrivilegeEscalation: false
           capabilities:
             drop:
-              - ALL
+            - ALL
             add:
-              - CHOWN
-              - DAC_OVERRIDE
-              - FSETID
-              - FOWNER
-              - MKNOD
-              - SETFCAP
-              - SETGID
-              - SETUID
+            - CHOWN
+            - DAC_OVERRIDE
+            - FSETID
+            - FOWNER
+            - MKNOD
+            - SETFCAP
+            - SETGID
+            - SETUID
           runAsNonRoot: false
           runAsUser: 0
           seccompProfile:
--- a/k8s/cloud_deps/base/nats/statefulset.yaml
+++ b/k8s/cloud_deps/base/nats/statefulset.yaml
@@ -135,6 +135,9 @@ spec:
                   values:
                   - pl-nats
       securityContext:
+        runAsNonRoot: true
+        runAsUser: 10100
+        fsGroup: 10100
         seccompProfile:
           type: RuntimeDefault
       # Common volumes for the containers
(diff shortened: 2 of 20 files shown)


vihangm:
ok to test

MrSupiri:
@vihangm I pushed two more changes to postgres and nats deployments, could you verify them as well?

k8s/cloud/base/api_deployment.yaml
```diff
@@ -135,6 +135,20 @@ spec:
           mountPath: /vizier-image-secret
         - name: elastic-certs-pub
           mountPath: /elastic-certs-pub
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+          runAsNonRoot: true
+          runAsUser: 10100
```
aimichelle, line 144: Is there any reason why you chose `10100` rather than `1000` which is more commonly used?

k8s/cloud_deps/base/nats/statefulset.yaml
```diff
@@ -225,3 +228,44 @@ spec:
...
+          capabilities:
+            drop:
+            - ALL
+          runAsNonRoot: true
+          runAsUser: 10100
+          seccompProfile:
+            type: RuntimeDefault
+
+      initContainers:
+      - name: init-chown-data
```
aimichelle, line 243: Would it be possible to use `fsGroup` and `fsGroupChangePolicy` to accomplish this rather than an initContainer?

k8s/cloud/base/api_deployment.yaml
MrSupiri: It's recommends to have higher UIDs (10000+) for containers won't overlap with host system's 
- https://docs.bridgecrew.io/docs/bc_k8s_37
- https://hub.datree.io/built-in-rules/prevent-uid-conflicts

k8s/cloud_deps/base/nats/statefulset.yaml
MrSupiri: I tested with `fsGroup` for with both nats and postgres workloads. Using `fsGroup` for nats server worked as expected.

But adding the it thing for postgres deployment result in an error stated as follows.
```
chmod: changing permissions of '/var/lib/postgresql/data': Operation not permitted
chmod: changing permissions of '/var/run/postgresql': Operation not permitted

FATAL:  data directory "/var/lib/postgresql/data" has wrong ownership 
```

postgres user within the container has the id of `999` yet having the following security context did fix the issue.

```
securityContext:
  fsGroup: 999
  runAsGroup: 999
  runAsNonRoot: true
  fsGroupChangePolicy: "Always"
  runAsUser: 999
  seccompProfile:
    type: RuntimeDefault
```

aimichelle:
@pixie-labs-buildbot , test this please.

k8s/cloud_deps/base/nats/statefulset.yaml
vihangm: I believe the `postgres` user is 70.
```
docker run --rm -it postgres:14-alpine id -u postgres
70
```

And it seems to work with the uid set to 70. So let's use that instead?

vihangm: The user is `999` under bullseye but we use alpine postgres which uses `70` :)

k8s/cloud_deps/public/postgres/postgres_deployment.yaml
```diff
@@ -4,11 +4,16 @@ kind: Deployment
...
     metadata:
       labels:
         name: postgres
     spec:
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - name: postgres
         image: postgres:13.3
```
MrSupiri, line 19: @vihangm we use the debian based version here right? I set the `runAsUser: 999` and replaced the container command with a `sleep 3600` and SSH into the pod and checked what's going on.

Screenshot 2023-02-06 at 15 27 55

 Turns out `/var/lib/postgresql` is owned by root even though it was `postgres` user was given the ownership of this according to https://github.com/docker-library/postgres/blob/186c93e85d4c4fcee8c300fdfd2e9991c5d3efc9/13/bullseye/Dockerfile#L27. So it fails at https://github.com/docker-library/postgres/blob/master/13/bullseye/docker-entrypoint.sh#L41

MrSupiri: I did a bash trace and found out it actually fails at https://github.com/docker-library/postgres/blob/master/13/bullseye/docker-entrypoint.sh#L92

Screenshot 2023-02-06 at 16 16 06

MrSupiri:
Just adding fsGroup to securityContext of postgres pod wasn't sufficient since `initdb` process runs a chown on `$PGDATA`. Since `/var/lib/postgresql/data` was owned by the container under root this process gets failed. Adding suffix `db` to `$PGDATA` fixed the issue since `db` dir is now located within the volume itself.

With that change the postgres container started successfully. 

Screenshot 2023-02-06 at 17 38 30

vihangm:
Oh yeah this used to be a debian based container but if you rebase you'll see that I switched it to alpine based a little whole ago to align it with the rest of the postgres images we use.

The PGDATA solution also works, the downside is that it means we are changing the location of the db and if someone upgrades postgres, they will lose data right? (unless they manually move the db).

vihangm:
@MrSupiri Sorry this change has taken a while to get reviewed. If you don't mind splitting the `nats` and `postgres` change out from the PR, we can merge the rest of it in. I'll try to do some more testing for the `nats` and `postgres` deps and make sure that works as expected and we can work on that in a separate PR.

MrSupiri:
> The PGDATA solution also works, the downside is that it means we are changing the location of the db and if someone upgrades postgres, they will lose data right? (unless they manually move the db).

My bad, I forgot to account delta updates.


> @MrSupiri Sorry this change has taken a while to get reviewed. If you don't mind splitting the `nats` and `postgres` change out from the PR, we can merge the rest of it in. I'll try to do some more testing for the `nats` and `postgres` deps and make sure that works as expected and we can work on that in a separate PR.

Sure, I will make the necessary changes.

vihangm (approved):

aimichelle (approved):

Issue #147, opened 2020-10-08
Support ARM CPUs


JamesAtIntegratnIO:
**Describe the bug**

**To Reproduce**
1. Follow this [tutorial](https://ubuntu.com/tutorials/how-to-kubernetes-cluster-on-raspberry-pi#1-overview) to deploy Microk8s on a pi cluster. I used the channel for Microk8s 1.19
2. enable the following features in microk8s,
```
addons:
  enabled:
    dashboard            # The Kubernetes dashboard
    dns                  # CoreDNS
    ha-cluster           # Configure high availability on the current node
    host-access          # Allow Pods connecting to Host services smoothly
    ingress              # Ingress controller for external access
    metrics-server       # K8s Metrics Server for API access to service metrics
    rbac                 # Role-Based Access Control for authorisation
    registry             # Private image registry exposed on localhost:32000
    storage              # Storage class; allocates storage from host directory
```
3. Create an account with cluster admin privilege using certificates.
4. Create a kubeconfig on an external instance of kubectl using that cluster admin account
4. Install Pixie alongside kubectl
5. Deploy the Pixie Demo
6. Deploy Pixie
7. Pods fail in various states. Log output for each pod below.

**Expected behavior**
successful deployment

**Logs**
```
boboysdadda@DESKTOP-US92ARK:~$ px deploy
Pixie CLI

Running Cluster Checks:
 ✔    Kernel version > 4.14.0
 ✔    Cluster type is supported
 ✔    K8s version > 1.12.0
 ✔    Kubectl > 1.10.0 is present
 ✔    User can create namespace
Installing version: 0.5.2
Generating YAMLs for Pixie
Deploying Pixie to the following cluster: microk8s

Is the cluster correct? (y/n) [y] : y
Found 5 nodes
 ✔    Creating namespace
 ✔    Deleting stale Pixie objects, if any
 ✔    Deploying secrets and configmaps
 ✔    Deploying Cloud Connector
 ⠼    Waiting for Cloud Connector to come online
[0142] FATAL Timed out waiting for cluster ID assignment
```
```
boboysdadda@DESKTOP-US92ARK:~$ kubectl get pods -n pl
NAME                                      READY   STATUS              RESTARTS   AGE
vizier-cloud-connector-5696d4d66b-2td4h   0/1     ContainerCreating   0          20h
cert-provisioner-job-kjw6n                0/1     Error               0          20h
cert-provisioner-job-trgcv                0/1     Error               0          20h
```
```
[...]

PR #761, opened 2023-02-03
[build/sysroots] Add test runner for sysroot builds.

JamesMBartlett:
Summary: Adds test runner for chrooting into an environment with the sysroot installed at root.
This PR allows running all tests in either the x86 or arm64 sysroots.

The chroot environment runs in its own mount and PID namespaces (via linux's unshare command). Once we replace `docker` with `podman` we can also run the chroot in its own network namespace.
As well as mounting the sysroot at root, the chroot environment is also setup to have both bazel's output_base and the top of the Pixie repo, bind mounted into the same locations in the chroot.
This makes bazel's symlinks work as expected.
Additionally, these two mounts use an overlayfs, such that the actual filesystem is not changed if there are any changes made inside the chroot environment.

Relevant Issues: #147

Type of change: /kind build

Test Plan: This shouldn't affect the normal build (relying on Jenkins to test that)
Also all tests pass for `bazel test //... --config=x86_64_sysroot`
Additionally, all but 7 tests pass for `bazel test //... --config=aarch64_sysroot`. These failing tests will be taken care of in future PRs.

Signed-off-by: James Bartlett <jamesbartlett@pixielabs.ai>

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -75,12 +75,22 @@ build:linux --action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
 build:clang-base --//bazel/cc_toolchains:compiler=clang
 build:clang-base --//bazel/cc_toolchains:libc_version=glibc_host
 
+# We build our own chroot for the sysroot tests, which doesn't work well under bazel's sandbox.
+build:sysroot-base --//bazel/test_runners:test_runner=sysroot_chroot
+test:sysroot-base --strategy TestRunner=standalone
+test:sysroot-base --run_under="bazel/test_runners/sysroot_chroot/test_runner.sh"
+
 build:x86_64_sysroot --config=clang
 build:x86_64_sysroot --//bazel/cc_toolchains:libc_version=glibc2_36
+build:x86_64_sysroot --config=sysroot-base
 
 build:aarch64_sysroot --config=clang
 build:aarch64_sysroot --//bazel/cc_toolchains:libc_version=glibc2_36
 build:aarch64_sysroot --platforms=//bazel/cc_toolchains:linux-aarch64
+build:aarch64_sysroot --config=sysroot-base
+# Increase test timeouts for qemu (don't increase the slowest ones because those are already very long).
+test:aarch64_sysroot --test_timeout=180,600,1800,3600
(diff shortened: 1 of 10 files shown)


.bazelrc
```diff
@@ -75,12 +75,22 @@ build:linux --action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
...
 build:x86_64_sysroot --//bazel/cc_toolchains:libc_version=glibc2_36
+build:x86_64_sysroot --config=sysroot_common
 
 build:aarch64_sysroot --config=clang
 build:aarch64_sysroot --//bazel/cc_toolchains:libc_version=glibc2_36
 build:aarch64_sysroot --platforms=//bazel/cc_toolchains:linux-aarch64
+build:aarch64_sysroot --config=sysroot_common
+# Increase test timeouts for qemu (don't increase the slowest ones because those are already very long).
+test:aarch64_sysroot --test_timeout=180,600,1800,3600
+test:aarch64_sysroot --test_env=QEMU_STRACE
```
zasgar, line 93: should we remove this?

.bazelrc
```diff
@@ -75,12 +75,22 @@ build:linux --action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
 build:clang-base --//bazel/cc_toolchains:compiler=clang
 build:clang-base --//bazel/cc_toolchains:libc_version=glibc_host
 
+# We build our own chroot for the sysroot tests, which doesn't work well under bazel's sandbox.
+build:sysroot_common --//bazel/test_runners:test_runner=sysroot_chroot
+test:sysroot_common --strategy TestRunner=standalone
+test:sysroot_common --run_under="bazel/test_runners/sysroot_chroot/test_runner.sh"
```
zasgar, line 81: The other terminology we have used is <...>-base. We should probably keep this consistent.

.bazelrc
JamesMBartlett: This allows people to set `QEMU_STRACE` to cause the test to dump strace logs, but if you don't set it in your env then it doesn't print the strace logs.

JamesMBartlett:
@pixie-labs-buildbot test this please

zasgar (approved):

PR #772, opened 2023-02-05
Restrict token permissions to readonly for workflows

vihangm:
Summary: These issues were flagged by CLOMonitor. Restrict permissions to Read
only.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: All workflows should continue to run.


JamesMBartlett (approved):
This is the default already, but I suppose its good to be explicit

PR #774, opened 2023-02-05
Upgrade rules_py

vihangm:
Summary: This upgrades the version of python being used which in turn fixes
the binary install of numpy.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: All python buildables and tests continue to work.

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -235,6 +235,7 @@ def _pl_cc_toolchain_deps():
     cc_toolchain_config_repo("unix_cc_toolchain_config", patch = "//bazel/cc_toolchains:unix_cc_toolchain_config.patch")
 
 def _pl_deps():
+    _bazel_repo("bazel_skylib")
     _bazel_repo("bazel_gazelle")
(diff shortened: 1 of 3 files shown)


bazel/repositories.bzl
```diff
@@ -235,6 +235,7 @@ def _pl_cc_toolchain_deps():
     cc_toolchain_config_repo("unix_cc_toolchain_config", patch = "//bazel/cc_toolchains:unix_cc_toolchain_config.patch")
 
 def _pl_deps():
+    _bazel_repo("bazel_skylib")
```
JamesMBartlett, line 238: A little odd listing this twice, but I don't really have a cleaner suggestion without creating a multi stage thing like tensorflow.

JamesMBartlett (approved):

bazel/repositories.bzl
vihangm: Yeah, both `rules_py` and our `cc_toolchains` depend on it and so we need this to be setup before either. I could pull it into a func of it's own but that seems strange too.

PR #779, opened 2023-02-06
Prevent pgsql stitcher from assuming that a matching response will exist when a request is seen

### LINKED ISSUE #697: Deflake `pgsql_trace_bpf_test`
`//src/stirling/source_connectors/socket_tracer:pgsql_trace_bpf_test` has a <90% pass rate in CI.
De-flake the test, run locally with `runs_per_test` set to a large number to ensure that the pass rate is higher than 90% and then re-enable the test in CI by removing the `disabled_flaky_test` tag.

ddelnano:
Summary: Prevent pgsql stitcher from assuming that a matching response will exist when a request is seen

The pgsql_trace_bpf_test has been flaky and led us to disabling it recently #697. This phab addresses the underlying issue so the test is more reliable.

The source of the flakiness is due to how the pgsql stitcher works. It previously assumed that when it sees a request that the matching response exists during the same invocation to StitchFrames. For our test harness, this is true the majority of the time. However, there are times where stirling sees the request in isolation, calls StitchFrames and then receives the response by the second call to StitchFrames. The previous stitcher implementation discards all requests at the [end of a run](https://github.com/pixie-io/pixie/blob/28a762a31673ee2e3956693df7871393bcacf156/src/stirling/source_connectors/socket_tracer/protocols/pgsql/stitcher.cc#L462), so when this happens we lose the request and can never match the lone response on the next StitchFrames invocation. This results in all of the test cases occasionally seeing the following error:

```
	src/stirling/source_connectors/socket_tracer/pgsql_trace_bpf_test.cc:219: Failure
  tablets.size()
    Which is: 0
  1
```

While this test was flaky, this detected a real bug in our stitcher and further demonstrates why having a true end to end test is valuable even if it's not as reproducible.

Relevant Issues: Fixes #697

Type of change: /kind bug

Test Plan: See the items below

- Compared the syscall and user space buffers in the failing ([P278](https://phab.corp.pixielabs.ai/P278)) and passing cases ([P279](https://phab.corp.pixielabs.ai/P279)). Specifically see how stirling's buffer for the initial server response is contains [CCREATE FUNCTION](https://phab.corp.pixielabs.ai/P279$5404) ([Command Complete 'B'](https://www.postgresql.org/docs/8.2/protocol-message-formats.html)) when the test passes, but [does not](https://phab.corp.pixielabs.ai/P278$3893) when the test fails. This results in the failing case seeing [14 frames](https://phab.corp.pixielabs.ai/P278$3896), while the passing case sees all [16 frames](https://phab.corp.pixielabs.ai/P279$5407)
- Ran the end to end test with --runs_per_test=10 (20 total runs) and verified it's no longer flaky ([P281](https://phab.corp.pixielabs.ai/P281))
- Verified the unit test fails on main so this flaky behavior would get caught outside of the end to end test

Changelog Message:

```release-note
Fixes a bug where postgres requests could be dropped causing missing protocol traces
```

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/pgsql/stitcher.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/pgsql/stitcher.cc
@@ -130,7 +130,10 @@ Status FillQueryResp(MsgDeqIter* resp_iter, const MsgDeqIter& end, QueryReqResp:
     }
   }
 
-  // Move the iterator pass the last message.
+  // TODO(ddelnano): This iterator incrementing should likely happen
+  // after the InvalidArgument error is returned below. This should only
+  // impact the case when there is an error, but we are leaving as is
+  // avoid a larger change without further verification.
   if (*resp_iter != end) {
     ++(*resp_iter);
   }
@@ -204,6 +207,10 @@ Status HandleQuery(const RegularMessage& msg, MsgDeqIter* resp_iter, const MsgDe
     }
   }
 
+  // TODO(ddelnano): This iterator incrementing should likely happen
+  // after the InvalidArgument error is returned below. This should only
+  // impact the case when there is an error, but we are leaving as is
+  // avoid a larger change without further verification.
   if (*resp_iter != end) {
     ++(*resp_iter);
   }
--- a/src/stirling/source_connectors/socket_tracer/protocols/pgsql/stitcher.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/pgsql/stitcher.cc
@@ -35,12 +35,6 @@ namespace pgsql {
 
 namespace {
 
-void AdvanceIterBeyondTimestamp(MsgDeqIter* start, const MsgDeqIter& end, uint64_t ts) {
-  while (*start != end && (*start)->timestamp_ns < ts) {
-    ++(*start);
-  }
-}
-
 template <typename TElemType>
 class DequeView {
  public:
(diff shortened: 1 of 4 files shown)


JamesMBartlett:
Rebase on #783 to fix pr description linter

zasgar (approved):

src/stirling/source_connectors/socket_tracer/protocols/pgsql/stitcher.cc
```diff
@@ -204,6 +207,10 @@ Status HandleQuery(const RegularMessage& msg, MsgDeqIter* resp_iter, const MsgDe
     }
   }
 
+  // TODO(ddelnano): This iterator incrementing should likely happen
+  // after the InvalidArgument error is returned below. This should only
+  // impact the case when there is an error, but we are leaving as is
+  // avoid a larger change without further verification.
```
ddelnano, line 213: @etep I left this as a comment since I didn't get the unit test setup as quickly as I hoped.

src/stirling/source_connectors/socket_tracer/protocols/pgsql/stitcher.cc
```diff
@@ -432,9 +432,6 @@ RecordsWithErrorCount<pgsql::Record> StitchFrames(std::deque<pgsql::RegularMessa
   while (req_iter != reqs->end() && resp_iter != resps->end()) {
     auto& req = *req_iter;
     ++req_iter;
-    if (resp_iter->consumed) {
-      continue;
-    }
```
ddelnano: I added a `CHECK(false)` statement here and re ran the bpf test 20 times to verify that this logic was never triggered.

This was identified as part of @etep's code review.

etep (approved):

ddelnano:
@pixie-io/maintainers this is ready to be merged now.

PR #789, opened 2023-02-06
Add Stat Chart (#787)

### LINKED ISSUE #787: Add Stat Chart Type to UI
We can already show these types of charts/graphs:
- Network map
- Timeseries (multiple stacked line graphs)
- Bar charts
- Histograms
- Flamecharts (`px/perf_flamegraph`, for example)

We could also use one to show the latest value of a given table column in large font.
Nothing fancy here, just the name and value of the metric with units.

NickLanam:

image

Summary: Adds a new chart type, 'StatChart', which just displays the latest value of a single metric with units and label.

Relevant Issues: Fixes #787

Type of change: /kind feature

Test Plan: Try this script with this vis spec.
```python
import px
import pxviews

ns_per_ms = 1000 * 1000
ns_per_s = 1000 * ns_per_ms
# Window size to use on time_ column for bucketing.
window_ns = px.DurationNanos(10 * ns_per_s)

def resource_timeseries(start_time: str, node: px.Node):
    ''' Gets the windowed process stats (CPU, memory, etc) for the input node.

    Args:
    @start_time Starting time of the data to examine.
    @node: The full name of the node to filter on.
    '''
    df = pxviews.container_process_timeseries(start_time, px.now(), window_ns)
    df = df[df.ctx['node'] == node]
    df['node'] = df.ctx['node']

    df = df.groupby(['time_', 'node']).agg(
        cpu_usage=('cpu_usage', px.sum),
        actual_disk_read_throughput=('actual_disk_read_throughput', px.sum),
        actual_disk_write_throughput=('actual_disk_write_throughput', px.sum),
        total_disk_read_throughput=('total_disk_read_throughput', px.sum),
        total_disk_write_throughput=('total_disk_write_throughput', px.sum),
        rss=('rss', px.sum),
        vsize=('vsize', px.sum),
    )

    df.groupby_col = df['node']
    return df
```
```json
{
  "variables": [
      {
          "name": "start_time",
          "type": "PX_STRING",
          "description": "The relative start time of the window. Current time is assumed to be now",
          "defaultValue": "-5m"
      },
      {
          "name": "node",
          "type": "PX_NODE",
          "description": "The node name to filter on"
      }
  ],
  "globalFuncs": [
      {
          "outputName": "resource_timeseries",
          "func": {
              "name": "resource_timeseries",
              "args": [
                  {
                      "name": "start_time",
                      "variable": "start_time"
                  },
                  {
                      "name": "node",
                      "variable": "node"
                  }
              ]
          }
      }
  ],
  "widgets": [
      {
          "name": "CPU Stat",
          "position": {
              "x": 0,
              "y": 0,
              "w": 1,
              "h": 1
          },
          "globalFuncOutputName": "resource_timeseries",
          "displaySpec": {
              "@type": "types.px.dev/px.vispb.StatChart",
              "title": "CPU Usage",
              "stat": {
                  "value": "cpu_usage"
              }
          }
      },
      {
        "name": "Bytes Read Stat",
        "position": {
            "x": 2,
            "y": 0,
            "w": 2,
            "h": 2
        },
        "globalFuncOutputName": "resource_timeseries",
        "displaySpec": {
            "@type": "types.px.dev/px.vispb.StatChart",
            "title": "Bytes Read",
            "stat": {
                "value": "total_disk_read_throughput"
            }
        }
    }
  ]
}
```

Changelog Message:
```release-note
New chart type: Stat. Shows a label, metric, and units for that metric.
```


aimichelle:
Screenshots?

NickLanam:
> Screenshots?

I think I attached those a couple of seconds after you loaded the page

src/api/proto/vispb/vis.proto
```diff
@@ -308,6 +308,17 @@ message TimeseriesChart {
   Axis y_axis = 4;
 }
 
+// StatChart The visual spec for a basic stat chart. Shows the latest value in the table as a big
+// number with units.
+message StatChart {
+  message Stat {
+    // The (numeric) data field that will be used as the Y axis
+    string value = 1;
```
zasgar, line 316: Shouldn't this be a numeric type? Like a double or string? 
How does this work with semantic types?

NickLanam: Same way it does in `message Timeseries` in the same file: it's the id of the field in the results. The `relation` gets used to then find the semantic type of that field, see `stat-chart.tsx` in this same PR.

src/api/proto/vispb/vis.proto
```diff
@@ -308,6 +308,17 @@ message TimeseriesChart {
...
+  }
+  Stat stat = 1;
```
aimichelle, line 318: Can you add comments for each of these fields? We try to do so in our protos even if it seems like it should be obvious.

src/pxl_scripts/watch.sh
```diff
@@ -52,7 +52,7 @@ cleanup() {
     kill "${SERVER_PID}"
   fi
   emph "Cleanup"
-  echo "Run ${tty_green}localStorage.clear('px-custom-bundle-path')${tty_reset}"
+  echo "Run ${tty_green}localStorage.clear('px-custom-bundle-paths')${tty_reset}"
```
aimichelle, line 55: These changes should probably go in a separate PR.

aimichelle (changes requested):
Marking as request changes, since I didn't for my previous comments.

src/api/proto/vispb/vis.proto
NickLanam: Done.

aimichelle (approved):

PR #791, opened 2023-02-07
Add a checker to ensure necessary generated files are updated in a PR

vihangm:
Summary: This linter runs generators for the various types of generated
files that checked into our codebase and fails if any of the files
are changed. i.e. they should have been regen for a PR.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Will trigger the linter with failures on this diff.


.github/workflows/pr_genfiles.yml
```diff
@@ -0,0 +1,78 @@
...
+jobs:
+  get-dev-with-extras-image:
+    runs-on: ubuntu-latest
+    outputs:
+      image-with-tag: ${{ steps.get-version.outputs.image }}
+    steps:
+    - uses: actions/checkout@v3
+    - id: get-version
+      run: >-
+        IMAGE_NAME="gcr.io/pixie-oss/pixie-dev-public/dev_image_with_extras";
```
JamesMBartlett, line 16: do we need extras?

JamesMBartlett (approved):
the caching stuff is nifty

.github/workflows/pr_genfiles.yml
vihangm: good catch, I don't think so.

zasgar (approved):

PR #792, opened 2023-02-07
Simplify operator etcd repair

aimichelle:
Summary: The operator tries to be smart by trying to switch the user to an etcd-based deployment based on some conditions, as long as they have been in that state for over 5 mins. These checks may be too hasty, as sometimes a user's cluster may just be slower for various reasons. This leads to some users getting unexpectedly moved to an etcd-based deploy.
There is also some extra logic to fix a state that users could have gotten into because of auto-repair. Now that the operator has been out in the wild for sometime, we can clean up this extra logic for cleanliness/simplification.

A followup PR is coming to attempt to fix etcd-based deploys which may have gotten to a bad state (lost quorum)

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Skaffold deploy operator

### DIFF
--- a/src/operator/controllers/monitor.go
+++ b/src/operator/controllers/monitor.go
@@ -37,7 +37,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	v1 "k8s.io/api/core/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/informers"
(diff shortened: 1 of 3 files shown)


src/operator/controllers/monitor.go
```diff
@@ -646,114 +644,6 @@ func (m *VizierMonitor) repairVizier(state *vizierState) error {
...
-		if err != nil {
-			log.WithError(err).Error("Failed to update certs")
-		}
-		m.certState = okState()
-
-		log.Info("Bouncing Vizier pods to get certs update")
-		err = k8s.DeletePods(m.clientset, m.namespace, "")
-		if err != nil {
-			return err
-		}
```
vihangm, line 667: we should keep this section right?

vihangm (changes requested):

src/operator/controllers/monitor.go
aimichelle: Whoops, yes, accidentally removed it. Thanks for catching!

vihangm (approved):

vihangm (approved):

PR #797, opened 2023-02-07
Use seeded random to fix GoHTTPTraceTest.LargePostMessage

### LINKED ISSUE #757: Fix `GoHTTPTraceTest.LargePostMessage` and re-enable assertion
One of the assertions `GoHTTPTraceTest.LargePostMessage` in was looking at the message body from a HTTPSever and comparing the body to a known value.
However the data generated by http server is generated with `rand.Int`. Expectations based on random data are brittle and can easily change. This was found during a golang upgrade and the assertion was disabled in ca238fbec08e9579bd9cecbd1c9f23561bf65d5d

The assertion or the test server should be updated so that we don't lose coverage without relying on randomly generated data to be static over time.

**To Reproduce**
Go to src/stirling/source_connectors/socket_tracer/http_trace_bpf_test.cc:138
Uncomment the assertion that looks at `record_batch[kHTTPReqBodyIdx]->Get<types::StringValue>(target_record_idx)`
Run the test

**Expected behavior**
The test should pass

vihangm:
Summary: go 1.20 changed the default random seeding. Since we do
actually want a seeded random for this test server/client, this
change ensures that we explicitly use a seeded random.

Relevant Issues: Fixes #757

Type of change: /kind bug

Test Plan: The disabled test assertion now passes again.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/http_trace_bpf_test.cc
+++ b/src/stirling/source_connectors/socket_tracer/http_trace_bpf_test.cc
@@ -132,19 +132,16 @@ TEST_F(GoHTTPTraceTest, LargePostMessage) {
       AllOf(HasSubstr(R"("Accept-Encoding":"gzip")"),
             HasSubstr(absl::Substitute(R"(Host":"localhost:$0")", go_http_fixture_.server_port())),
             ContainsRegex(R"(User-Agent":"Go-http-client/.+")")));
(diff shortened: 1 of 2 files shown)


JamesMBartlett (approved):

JamesMBartlett (approved):
Looks the same as when I reviewed it before?

zasgar (approved):

PR #803, opened 2023-02-07
fix: catalog operator param error

iutx:
Signed-off-by: Asher Liu <root@viper.run>

```console
catalog -h
```

current param will cause error:
image

### DIFF
--- a/k8s/operator/helm/templates/00_olm.yaml
+++ b/k8s/operator/helm/templates/00_olm.yaml
@@ -124,10 +124,10 @@ spec:
           command:
           - /bin/catalog
           args:
-          - '-namespace'
+          - '--namespace'
           - {{ .Values.olmNamespace }}
-          - -configmapServerImage={{ if .Values.registry }}{{ .Values.registry }}/quay.io-operator-framework-{{ else }}quay.io/operator-framework/{{ end }}configmap-operator-registry:latest
-          - -util-image
+          - --configmapServerImage={{ if .Values.registry }}{{ .Values.registry }}/quay.io-operator-framework-{{ else }}quay.io/operator-framework/{{ end }}configmap-operator-registry:latest
+          - --util-image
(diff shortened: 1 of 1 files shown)


k8s/operator/helm/templates/00_olm.yaml
```diff
@@ -124,10 +124,10 @@ spec:
...
           args:
-          - '-namespace'
+          - '--namespace'
           - {{ .Values.olmNamespace }}
-          - -configmapServerImage={{ if .Values.registry }}{{ .Values.registry }}/quay.io-operator-framework-{{ else }}quay.io/operator-framework/{{ end }}configmap-operator-registry:latest
-          - -util-image
+          - --configmapServerImage={{ if .Values.registry }}{{ .Values.registry }}/quay.io-operator-framework-{{ else }}quay.io/operator-framework/{{ end }}configmap-operator-registry:latest
+          - --util-image
           -  {{ if .Values.registry }}{{ .Values.registry }}/quay.io-operator-framework-olm{{ else }}quay.io/operator-framework/olm@sha256:b706ee6583c4c3cf8059d44234c8a4505804adcc742bcddb3d1e2f6eff3d6519{{ end }}
           image: {{ if .Values.registry }}{{ .Values.registry }}/quay.io-operator-framework-olm{{ else }}quay.io/operator-framework/olm@sha256:b706ee6583c4c3cf8059d44234c8a4505804adcc742bcddb3d1e2f6eff3d6519{{ end }}
```
vihangm, line 132: Looks like you are using a custom registry and this issue is happening because the image isn't pinned to the same SHA hash in the custom registry.
Instead of changing the flags, can you just add the SHA to the custom registry template here so that we are always consistently using the same images, and so that this template stays inline with the original olm yaml.

vihangm (changes requested):
Look

k8s/operator/helm/templates/00_olm.yaml
iutx: Thanks, got it.❤️

vihangm: I meant that you could update this PR to also update it for the Pixie charts!
Lmk if you would like to make a PR, else I can fix this on our end soon.

Issue #699, opened 2023-01-19
Deflake `openssl_trace_bpf_test`


vihangm:
`//src/stirling/source_connectors/socket_tracer:openssl_trace_bpf_test` has a <90% pass rate in CI.
De-flake the test, run locally with `runs_per_test` set to a large number to ensure that the pass rate is higher than 90% and then re-enable the test in CI by removing the `disabled_flaky_test` tag.

ddelnano:
After reviewing our build buddy logs, it appears all of the failures originate from failing to attach kprobes. These result in the error seen below:
```
Value of: IsOK(::px::StatusAdapter(source_->Init()))
  Actual: false (Internal : Unable to attach kprobe for __x64_sys_connect using syscall__probe_entry_connect)
Expected: true
I20230113 17:51:57.281163 11814 container_runner.cc:59] docker rm -f node_server_791890848756 &>/dev/null
[  FAILED  ] OpenSSLTraceRawFptrsTest/2.ssl_capture_node_client, where TypeParam = px::stirling::Node12_3_1ContainerWrapper (11274 ms)
[----------] 2 tests from OpenSSLTraceRawFptrsTest/2 (33423 ms total)
[----------] 1 test from OpenSSLTraceRawFptrsTest/3, where TypeParam = px::stirling::Node14_18_1AlpineContainerWrapper
[ RUN      ] OpenSSLTraceRawFptrsTest/3.ssl_capture_ruby_client
I20230113 17:51:57.798910 11814 container_runner.cc:43] Loaded image: bazel/src/stirling/source_connectors/socket_tracer/testing/containers:node_14_18_1_alpine_image
I20230113 17:51:57.798959 11814 container_runner.cc:121] docker run --rm --pid=host --name=node_server_802148102183 bazel/src/stirling/source_connectors/socket_tracer/testing/containers:node_14_18_1_alpine_image
Error: No such object: node_server_802148102183
I20230113 17:51:57.960911 11814 container_runner.cc:160] Container node_server_802148102183 status: 
I20230113 17:51:57.961170 11814 container_runner.cc:169] Container node_server_802148102183 not yet running, will try again (60 attempts remaining).
I20230113 17:51:58.989207 11814 container_runner.cc:160] Container node_server_802148102183 status: running
I20230113 17:51:59.016650 11814 container_runner.cc:191] Container node_server_802148102183 process PID: 28925
[...]

PR #804, opened 2023-02-07
Force openssl_trace_bpf_test to run 100 times to try to reproduce flakiness

ddelnano:
Summary: Force openssl_trace_bpf_test to run 100 times to try to reproduce flakiness

We recently disabled the `openssl_trace_bpf_test` in #699. I've been trying to reproduce the test failures, but in 200 runs of the test it was successful 100% of the time. If we can't reproduce the failures, I think we should consider re-enabling it as is. As a final attempt to reproduce the problem, I'd like to run the test 100 times in CI and see if the issue still exists.

Relevant Issues: #699

Type of change: /kind flake

Test Plan: Verify if CI passes and then re-enable the test -  #ci:bpf-build
- [x] Verified that running the test 200 times did not reproduce the failure.
```
root@px-dev-docker-ddelnano-test-machine:/px/src/px.dev/pixie# bazel test -c opt --config=bpf --build_metadata=COMMIT_SHA=6fe385c0deeede5fdb31c2a27b7e68d8af559377 --runs_per_test=100 --test_arg='-v=2' --test_arg='--vmodule=bcc_wrapper=0' --test_arg='--vmodule=socket_trace_connector=3' --target_pattern_file bazel_tests_bpf

Target //src/stirling/source_connectors/socket_tracer:openssl_trace_bpf_test up-to-date:
  bazel-bin/src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test
INFO: Elapsed time: 36510.166s, Critical Path: 2951.70s
INFO: 201 processes: 1 internal, 200 local.
INFO: Build completed successfully, 201 total actions
//src/stirling/source_connectors/socket_tracer:openssl_trace_bpf_test    PASSED in 185.9s
  Stats over 200 runs: max = 185.9s, min = 179.8s, avg = 182.5s, dev = 1.2s

Executed 1 out of 1 test: 1 test passes.

```

### DIFF
--- a/ci/bpf/03_run_tests.sh
+++ b/ci/bpf/03_run_tests.sh
@@ -40,7 +40,11 @@ read -ra BAZEL_ARGS <<< "${BAZEL_ARGS}"
 bazel build "${BAZEL_ARGS[@]}" --target_pattern_file "${BUILDABLE_FILE}"
 check_retval $?
 
-bazel test "${BAZEL_ARGS[@]}" --target_pattern_file "${TEST_FILE}"
+# Temporarily override which test is run since running the openssl_trace_bpf_test
+# 230 times did not reproduce the failure.
+echo "//src/stirling/source_connectors/socket_tracer:openssl_trace_bpf_test" > "${TEST_FILE}"
+
+bazel test "${BAZEL_ARGS[@]}" --runs_per_test=50 --target_pattern_file "${TEST_FILE}"
 check_retval $?
 
 rm -rf bazel-testlogs-archive
--- a/src/stirling/source_connectors/socket_tracer/BUILD.bazel
+++ b/src/stirling/source_connectors/socket_tracer/BUILD.bazel
@@ -438,8 +438,6 @@ pl_cc_test(
     flaky = True,
     shard_count = 2,
     tags = [
-        # TODO: This test has a <90% pass rate in CI, please fix and re-enable.
-        "disabled_flaky_test",
         "cpu:16",
         "no_asan",
         "requires_bpf",


JamesMBartlett:
I suspect you might need to run it at the same time as other bpf tests to reproduce the issue

ddelnano:
@JamesMBartlett don't we intentionally size our [gcp instances](https://github.com/pixie-io/pixie/blob/3a4b5c45fa33f300c0d3e540c856857364630d3a/ci/bpf/00_create_instance.sh#L57) and give the [test](https://github.com/pixie-io/pixie/blob/3a4b5c45fa33f300c0d3e540c856857364630d3a/src/stirling/source_connectors/socket_tracer/BUILD.bazel#L443) a cpu 'cost' so that they only execute one per anyway?

JamesMBartlett:
We used to use 48 cpu machines for BPF tests, in fact we still tell bazel that bpf tests have 48 cpus available (https://github.com/pixie-io/pixie/blob/main/ci/jenkins.bazelrc#L43). As far as I'm aware the `cpu:16` tag was never intended to make tests exclusive, just to mark them as requiring more cores. To mark tests as exclusive we use bazel's `exclusive` tag. However, we try to avoid marking tests as `exclusive` because it significantly slows down test runs.

@vihangm Do you know why we switched from 48 cores to 16 for bpf tests? (maybe we always ran on 16 cores and just told bazel we had 48?)

ddelnano:
Closing this based on @JamesMBartlett's comments about the parallel execution. #699 has more details on how to address this.

PR #806, opened 2023-02-07
Add auto-repair for crashing etcd in operator

aimichelle:
Summary: We've noticed that a lot of clusters running etcd are in a state where etcd is crashing. This is very likely to happen if etcd loses quorum. This PR updates the operator to detect that state and attempt to fix it by restarting etcd.

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Skaffold deploy operator

### DIFF
--- a/src/operator/controllers/monitor.go
+++ b/src/operator/controllers/monitor.go
@@ -341,7 +341,7 @@ func getEtcdState(pods *concurrentPodMap) *vizierState {
 			continue
 		}
 		for _, c := range pod.pod.Status.ContainerStatuses {
-			if c.State.Waiting != nil && c.State.Waiting.Reason == "CrashLoopBackOff" {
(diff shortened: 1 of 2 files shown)


src/operator/controllers/monitor.go
```diff
@@ -325,6 +325,31 @@ func getNATSState(client HTTPClient, pods *concurrentPodMap) *vizierState {
...
+
+	for podName, pod := range unlabeledPods {
+		if !strings.HasPrefix(podName, "pl-etcd-") {
+			continue
+		}
+		for _, c := range pod.pod.Status.ContainerStatuses {
+			if c.State.Waiting != nil && c.State.Waiting.Reason == "CrashLoopBackOff" {
+				return &vizierState{Reason: status.EtcdPodsCrashing}
+			}
+		}
```
vihangm, line 347: should we check for a certain number of restarts and number of pods that are in this state?

vihangm (approved):

Issue #719, opened 2023-02-01
Improve stability of the `perf_profiler_bpf_test` test


ddelnano:
The `//src/stirling/source_connectors/perf_profiler:perf_profiler_bpf_test`  test has a 25% pass rate in CI. We have disabled the test in XXX and need to ensure it is less flaky before re-enabling it by removing the disabled_flaky_test tag.

PR #812, opened 2023-02-07
Add pprof proto output as command line arg. to perf profiler test.

etep:
Summary: Adds pprof proto output to profiler test via a command line arg.
```
scripts/sudo_bazel_run.sh //src/stirling/source_connectors/perf_profiler:perf_profiler_bpf_test -- -pprof_pb_file cpp-test.pprof.pb --gtest_filter=PerfProfileBPFTest.PerfProfilerCppTest
```

Currently, we are using this to investigate flaky tests. We plan to (in a subsequent PR) refactor the pprof export code into a library function that enables pprof export from pixel scripts or from a standalone pem.

Type of change: /kind feature

Test Plan: Created a pprof proto file for each profiler test.

### DIFF
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -459,6 +459,8 @@ GIT_REPOSITORY_LOCATIONS = dict(
         commit = "fa36a4fa141b4e9486b9126640d54a94c1d36fce",
         shallow_since = "1638898188 -0800",
     ),
+
+    # Used by the perf profiler to create pprof proto output.
     com_github_google_pprof = dict(
         remote = "https://github.com/google/pprof.git",
         commit = "5a9e8f65f08f77f42ab2a02f18561033f44da75e",
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -459,6 +459,11 @@ GIT_REPOSITORY_LOCATIONS = dict(
         commit = "fa36a4fa141b4e9486b9126640d54a94c1d36fce",
         shallow_since = "1638898188 -0800",
     ),
+    com_github_google_pprof = dict(
+        remote = "https://github.com/google/pprof.git",
+        commit = "5a9e8f65f08f77f42ab2a02f18561033f44da75e",
+        shallow_since = "1675207505 -0800",
+    ),
 )
 
 # To use a local repo for local development, update the path to point to your local repo.
--- a/src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
+++ b/src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
@@ -37,9 +37,12 @@
 #include "src/stirling/source_connectors/perf_profiler/testing/testing.h"
 #include "src/stirling/testing/common.h"
 
+#include "proto/profile.pb.h"
+
 DEFINE_uint32(test_run_time, 30, "Number of seconds to run the test.");
 DEFINE_string(test_java_image_names, JDK_IMAGE_NAMES,
               "Java docker images to use as Java test cases.");
+DEFINE_string(pprof_pb_file, "", "Location to create pprof pb file.");
 DECLARE_bool(stirling_profiler_java_symbols);
 DECLARE_string(stirling_profiler_java_agent_libs);
 DECLARE_uint32(stirling_profiler_table_update_period_seconds);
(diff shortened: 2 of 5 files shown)


src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -251,6 +259,143 @@ class PerfProfileBPFTest : public ::testing::TestWithParam<std::filesystem::path
...
+
+    // Create a file.
+    std::ofstream ofs(FLAGS_pprof_pb_file, std::ios_base::out | std::ios_base::binary);
+
+    if (!ofs.is_open()) {
+      return error::ResourceUnavailable("Could not create or open file: $0.", FLAGS_pprof_pb_file);
+    }
+
+    // Serialize the pprof proto to that file.
+    if (!profile.SerializeToOstream(&ofs)) {
```
ddelnano, line 384: `WritePProfProtoFile` seems to be called by multiple tests. My understanding is that the resulting pprof file ends up being whichever test ran last. Is that correct?

etep: Yes, that's true. In the example usage, I show the test case with a gtest filter (i.e. so that only one test is run and then the resulting pprof proto is unambiguous). I will figure out a way to either clarify the usage, or fail if the feature is abused, or uniquify the file name.

bazel/repository_locations.bzl
```diff
@@ -459,6 +459,13 @@ GIT_REPOSITORY_LOCATIONS = dict(
         commit = "fa36a4fa141b4e9486b9126640d54a94c1d36fce",
         shallow_since = "1638898188 -0800",
     ),
+
+    # Used by the perf profiler to create pprof proto output.
+    com_github_google_pprof = dict(
```
JamesMBartlett, line 464: Please use http repos not git repos. git repos don't use the repository cache

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -251,6 +259,143 @@ class PerfProfileBPFTest : public ::testing::TestWithParam<std::filesystem::path
     VLOG(1) << std::endl;
   }
 
+  Status WritePProfProtoFile() {
```
JamesMBartlett, line 262: Im a little confused about the purpose of this in the test. It seems unrelated to testing the per profiler functionality. Is it just for debugging the test?

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
ddelnano: What if we specified the file prefix and then suffixed test name or test index on the end of it? I think naming the flag that way will make the usage more obvious as well.

bazel/repository_locations.bzl
etep: Hi @JamesMBartlett,

Could you take a look at the upstream pprof repo on github? I did not see any tags or any other artifacts that looked compatible with `http_archive` or `_bazel_repo`. Maybe there is a way to make it work, but it was not obvious to me. (or maybe you are suggesting a different way to bring the repo in). I had left a comment (above) where I invoked `_git_repo`:
```
# Using "git repo" because the pprof repo does not have any release tags.
```

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
etep: This is what I had in the summary of the PR:
```
Currently, we are using this to investigate flaky tests. We plan to (in a subsequent PR) refactor the pprof export code into a library function that enables pprof export from pixel scripts or from a standalone pem.
```

Since we are using this for test debug, I figured it was a nice moment to just review the basic pprof export code. We may refactor it into a library function soon.

Let's touch base in a different forum if you want to check out the results of the flaky test investigation.

zasgar (changes requested):

bazel/repository_locations.bzl
JamesMBartlett: We do this in a number of places in this file, you can just point to the commit itself. It's not ideal because github changes their hashes sometimes, but that's a separate issue we need to figure out. 
eg: "https://github.com/antlr/grammars-v4/archive/e53d7a1228505bfc80d8637808ef60e7eea92cc2.tar.gz"

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -239,6 +242,11 @@ class PerfProfileBPFTest : public ::testing::TestWithParam<std::filesystem::path
       histo_[stack_trace_str] += count;
     }
```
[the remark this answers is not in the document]
zasgar: I think it's better for us to make it a library first. If we want to write out files, it's probably best to write it in a utility exec. 

For verbose logging, we probably should not be writing to a file in the test.

PR #823, opened 2023-02-08
Add cacher workflow to main to generate caches for PRs

vihangm:
Summary: After reading the [docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache), I realized that github caches are local to PRs/branches.
This sets up a workflow on main pushes solely to populate the cache since other runs can restore from a cache on the main branch.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Test this to see if cache hits on PRs go up.

### DIFF
--- a/.github/workflows/pr_genfiles.yml
+++ b/.github/workflows/pr_genfiles.yml
@@ -24,21 +24,20 @@ jobs:
       options: --cpus 7
     steps:
     - uses: actions/checkout@v3
-    - name: Go cache
+    - name: go cache
       uses: actions/cache@v3
       with:
         path: /px/pkg/mod
-        key: build-cache-${{hashFiles('go.sum')}}
-    - name: Yarn cache
+        key: go-cache-${{ hashFiles('go.sum') }}
+        restore-keys: |
+          go-cache-
+    - name: yarn cache
       uses: actions/cache@v3
       with:
-        path: ./src/ui/.yarn
-        key: build-cache-${{hashFiles('src/ui/yarn.lock')}}
-    - name: Bazel cache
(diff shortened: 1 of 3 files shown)


.github/workflows/pr_genfiles.yml
```diff
@@ -24,21 +24,20 @@ jobs:
...
-        path: ./src/ui/.yarn
-        key: build-cache-${{hashFiles('src/ui/yarn.lock')}}
-    - name: Bazel cache
-      uses: actions/cache@v3
-      with:
-        path: ~/.cache/bazel
-        key: genfile-checker-${{hashFiles('WORKSPACE')}}
+        path: ./src/ui/.yarn/cache
+        key: yarn-cache-${{ hashFiles('src/ui/yarn.lock', 'src/ui/.yarnrc.yml') }}
+        restore-keys: |
```
JamesMBartlett, line 39: So restore-keys knows to pull from main cache somehow ?

JamesMBartlett (approved):

.github/workflows/pr_genfiles.yml
vihangm: Restore keys just lets you use a "stale cache" if the hash doesn't map. So basically if you change `yarn.lock` or `.yarnrc.yml`, you will get a cache miss but then this will populate the yarn cache with the LRU yarn cache and then create a new cache using that as a base.

Pulling from main cache is a actions/cache behavior that Github describes in the doc I linked to.

zasgar (approved):

PR #825, opened 2023-02-08
Add securityContext to nats and postgres

MrSupiri:
Summary: Introduced securityContext to nats and postgres workloads to limit access the containers has on the Kubernetes cluster. This will help to minimize some of the security risks these containers may impose on the cluster.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan:

1. Checkout the current branch.
2. Deploy Self hosted Pixie.

### DIFF
--- a/k8s/cloud_deps/public/postgres/postgres_deployment.yaml
+++ b/k8s/cloud_deps/public/postgres/postgres_deployment.yaml
@@ -7,11 +7,16 @@ spec:
   selector:
     matchLabels:
       name: postgres
+  strategy:
+    type: Recreate
   template:
     metadata:
       labels:
         name: postgres
     spec:
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - name: postgres
         image: postgres:14-alpine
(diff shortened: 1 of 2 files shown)


k8s/cloud_deps/public/postgres/postgres_deployment.yaml
```diff
@@ -7,11 +7,16 @@ spec:
   selector:
     matchLabels:
       name: postgres
+  strategy:
+    type: Recreate
```
MrSupiri, line 11: This was done to mitigate the Multi-Attach error for volume during upgrades.

k8s/cloud_deps/public/postgres/postgres_deployment.yaml
```diff
@@ -34,6 +39,46 @@ spec:
...
+            drop:
+            - ALL
+          runAsNonRoot: true
+          runAsUser: 10100
+          seccompProfile:
+            type: RuntimeDefault
+      initContainers:
+      - name: init-chown-data
+        image: busybox
+        command: ["chown", "-R", "10100:10100", "/var/lib/postgresql/data"]
```
MrSupiri, line 54: At the start of the postgres container, a process called `initdb` get executed which will run `chown` on the `$PGDATA` which is by default set to `/var/lib/postgresql/data`. 

Postgres container by default is [built](https://github.com/docker-library/postgres/blob/master/14/alpine/Dockerfile#L159) to own the `/var/lib/postgresql/data` by postgres (uid - 70) user. 
Screenshot 2023-02-08 at 14 59 07


But once a persistent volume get mounted to the path, the CSI driver[^1] set the uid of that mount point to root as shown below. Which in turns fails the `initdb`.
Screenshot 2023-02-08 at 14 57 51

The only way around this issue without giving the container the root access, is to run a temporary init container which will use root to change the ownership of the `data` directory within the persistent volume the UID we want. This is the [recommended workaround](https://learn.microsoft.com/en-us/troubleshoot/azure/azure-kubernetes/failure-setting-azure-disk-mount-options-uid-gid#workaround-2-use-an-init-container-and-the-change-owner-command-to-specify-the-user-and-group) by azure as well.

[^1]: We only tested this on AKS.

PR #843, opened 2023-02-09
Support pushing multi-arch images in vizier_build_release.sh

JamesMBartlett:
Summary: Adds pushing multi-arch images in the vizier release process.
Builds each architecture's images separately and pushes those individually. Then pushes a manifest list pointing to each architecture for each vizier image.

Relevant Issues: #147.

Type of change: /kind cleanup

Test Plan: Created an RC with these changes, saw that it built multi-arch images.
Ran `px deploy --vizier_version=<RC>` and it successfully deployed Pixie.

Changelog Message:
```release-note
Our releases now build multi-arch docker images that support x86_64 and aarch64.
ARM support is now in an early alpha stage. We'd love for the community to help test out ARM support.
```

### DIFF
--- a/ci/vizier_build_release.sh
+++ b/ci/vizier_build_release.sh
@@ -41,8 +41,36 @@ fi
 output_path="gs://${bucket}/vizier/${release_tag}"
 latest_output_path="gs://${bucket}/vizier/latest"
 
-bazel run --stamp -c opt --//k8s:image_version="${release_tag}" \
-    --stamp "${build_type}" //k8s/vizier:vizier_images_push "${extra_bazel_args[@]}"
+push_images_for_arch() {
+  arch="$1"
+  bazel run --stamp -c opt --//k8s:image_version="${release_tag}-${arch}" \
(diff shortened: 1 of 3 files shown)


ci/vizier_build_release.sh
```diff
@@ -41,8 +41,44 @@ fi
...
+  docker manifest push "${multiarch_image}"
+}
+
+push_all_multiarch_images() {
+  combined_image_lists="$(paste -d' ' "$@")"
+
+  while read -r -a images;
+  do
+    push_multiarch_image "${images[@]}"
+  done < <(echo "${combined_image_lists}")
```
vihangm, line 74: This relies on both lists being the same order. Instead let's create a list of images to push with the release tag (and no arch suffix) and just use that as the input to `push_multiarch_image`
We should just assume that images with all arch suffixes exist when trying to push and can error if something is missing.

vihangm (changes requested):

vihangm (approved):

PR #849, opened 2023-02-10
draft - Profiler cleanup: remove dependency on full stirling lib.

etep:
Summary: We remove the dependency on the full Stirling cc-lib for the stand alone perf profiler binary. This requires some re-writing of the program logic. We follow the pattern established in the perf profiler test case (e.g. use of the transfer data thread).

In the next diff, we will introduce a library function to create pprof proto output and replace the text printout with a pprof proto file output.

Type of change: /kind feature

Test Plan: Tested this locally both all processes and for a specific pid.

### DIFF
--- a/src/stirling/binaries/stirling_profiler.cc
+++ b/src/stirling/binaries/stirling_profiler.cc
@@ -114,13 +114,6 @@ class Profiler {
       }
     }
 
-    // This may be paranoid, but it is possible for StopTransferDataThread() to get
-    // called from the signal handler too. In any case, by the time we are here and from
-    // any entry point, transfer_data_thread_ should have been joined already.
-    if (transfer_data_thread_.joinable()) {
-      return error::Internal("transfer_data_thread_ failed to join.");
-    }
-
     return Status::OK();
   }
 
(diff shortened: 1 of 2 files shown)


ddelnano:
@etep since this builds on #844, can you merged that PR and then rebase?

src/stirling/binaries/stirling_profiler.cc
```diff
@@ -22,111 +22,247 @@
...
+        source_->TransferData(ctx_.get());
+      }
+    }
+
+    // This may be paranoid, but it is possible for StopTransferDataThread() to get
+    // called from the signal handler too. In any case, by the time we are here and from
+    // any entry point, transfer_data_thread_ should have been joined already.
+    if (transfer_data_thread_.joinable()) {
+      return error::Internal("transfer_data_thread_ failed to join.");
+    }
```
ddelnano, line 122: Can you explain this case? I'm not sure I follow why it's worth this additional check.

ddelnano:
One question about how the threads join, but otherwise lgtm.

src/stirling/binaries/stirling_profiler.cc
etep: I'm going to remove this check. TBH my original thought process was to check that the join op (above) had succeeded.

etep: Removed that just now.

ddelnano (approved):

PR #852, opened 2023-02-10
Support finding RET instructions on ARM

JamesMBartlett:
Summary: Adds support for finding ret instructions for aarch64.
Also, uses the ELF file's machine type to determine the disassembly context triple (and which ret instructions to look for), instead of basing it off of which architecture ELFReader was built for.

Relevant Issues: #147

Type of change: /kind cleanup

Test Plan: Manually checked that it can properly find the ret instruction for the prebuilt_test_exe. Need to update the elf_reader_test to actually test based on an aarch64 ELF file instead of just the x86_64 prebuilt_test_exe.

### DIFF
--- a/src/stirling/obj_tools/elf_reader.cc
+++ b/src/stirling/obj_tools/elf_reader.cc
@@ -431,38 +431,18 @@ std::string_view ElfReader::Symbolizer::Lookup(size_t addr) const {
 
 namespace {
 
-/**
- * RAII wrapper around LLVMDisasmContextRef.
- */
-class LLVMDisasmContext {
- public:
-  LLVMDisasmContext() {
-    InitLLVMOnce();
-
-#if X86_64
-    constexpr char triple[] = "x86_64-pc-linux";
-#elif AARCH64
-    constexpr char triple[] = "aarch64-pc-linux";
-#else
-#error Architecture not supported.
-#endif
-
-    // TripleName is ARCHITECTURE-VENDOR-OPERATING_SYSTEM.
-    // See https://llvm.org/doxygen/Triple_8h_source.html
-    ref_ = LLVMCreateDisasm(triple,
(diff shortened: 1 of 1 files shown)


src/stirling/obj_tools/elf_reader.cc
```diff
@@ -519,12 +570,26 @@ std::vector<uint64_t> FindRetInsts(utils::u8string_view byte_code) {
   return res;
 }
 
+StatusOr<Arch> GetArchFromELFMachine(ELFIO::Elf_Half machine) {
+  switch (machine) {
+    case ELFIO::EM_X86_64:
+      return ArchX86_64;
+    // ELFIO doesn't have the updated versions of EM_*, so we use the aarch64 value directly.
+    case 183:
```
vihangm, line 578: should we either update our fork with main or just update this one line in our fork from main?
https://github.com/serge1/ELFIO/blob/main/elfio/elf_types.hpp#L250

vihangm (approved):

ddelnano:
I agree that we should have an updated elf_reader_test to verify this works with an aarch64 elf file.

In addition to that have you ran the BPF tests on arm? My understanding is that this return instruction code is important for Go uretprobes, so I'd like to see that our go uprobe tests are pass for both architectures.

JamesMBartlett:
> I agree that we should have an updated elf_reader_test to verify this works with an aarch64 elf file.
> 
> 
> 
> In addition to that have you ran the BPF tests on arm? My understanding is that this return instruction code is important for Go uretprobes, so I'd like to see that our go uprobe tests are pass for both architectures.

The BPF tests don't currently work on ARM since the docker images we use for them are all built for x86. I was planning on delaying getting those to work until @zasgar qemu stuff lands, so that we can use an x86 machine to run them. I will manually verify that a go retprobe works.

JamesMBartlett:
I tested that go TLS probes work on ARM (at least for HTTP2 traffic). There seems to be some issue on both ARM and x86 with go TLS probes for HTTP1. But dom is looking into that separately.

PR #865, opened 2023-02-14
Fix uprobe deployment on go1.20

### LINKED ISSUE #870: Fix uprobe deployment with go 1.20 binaries
**Describe the bug**
uprobe deployment fails on binaries build with the go 1.20 SDK and compiler toolchain.

**To Reproduce**
1. Update the `go_tls_trace_bpf_test.cc` to also run against binaries built with go 1.20
2. Run the test.
3. The test fails.

**Expected behavior**
The unit tests should pass with go 1.20 binaries. Also pixie should be able to trace data end-to-end with applications on k8s that are built with go 1.20

JamesMBartlett:
Summary: go 1.20 changed the prefix of some of their symbols from `go.` to `go:`. This meant we wouldn't attach uprobes to go1.20 binaries.
This fixes the issue by checking both prefixes in uprobe_symaddrs.cc.
From the go1.20 release notes:
>Go 1.20 uses go: and type: prefixes for compiler-generated symbols rather than go. and type.. This avoids confusion for user packages whose name starts with go.. The [debug/gosym](https://tip.golang.org/pkg/debug/gosym) package understands this new naming convention for binaries built with Go 1.20 and newer.

Relevant Issue: Fixes #870

Type of change: /kind bug

Test Plan: Tested manually that TLS traffic to a go1.20 binary is properly traced by stirling wrapper. Once @vihangm's changes to add go1.20 binaries to the bpf tests lands, I will test with those as well.


src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.cc
```diff
@@ -142,19 +142,33 @@ StatusOr<std::string> InferHTTP2SymAddrVendorPrefix(ElfReader* elf_reader) {
   return vendor_prefix;
 }
 
+std::optional<int64_t> ResolveSymbolWithEachGoPrefix(ElfReader* elf_reader,
```
ddelnano, line 145: Can you add a comment above this function explaining the symbol naming change (essentially what was in your PR description mentions)?

ddelnano:
Once #866 lands we should be able to add the ebpf test, right? I'd prefer that we merge this functionality in with the corresponding test so that we have coverage from the beginning.

JamesMBartlett:
> Once #866 lands we should be able to add the ebpf test, right? I'd prefer that we merge this functionality in with the corresponding test so that we have coverage from the beginning.

Yes I will wait for the testing stuff to land

src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.cc
etep, line 145: Do we know the go version?
If so, should we:
1. only check `go.` if the go version is < 1.20?
2. reverse the order in the array, i.e. if v >= 1.20, check `go:` first?

JamesMBartlett: We could get the go version from the `runtime.buildVersion` symbol, (ala `ReadBuildVersion`). Unfortunately, the way this file is setup we'd have to repeatedly call that function in a number of places and then pipe it through, since there's no governing object or any other structure we can store the go version in, without a significant refactor. So my preference would be to keep it simple for now, and someone can potentially do the go version optimization later.

ddelnano (approved):

vihangm (approved):

PR #881, opened 2023-02-15
Operator: Fetch and send K8s cluster version to ConfigManager

aimichelle:
Summary: You can now send your K8s cluster version to the config manager service, which allows the config manager service to modify the Vizier YAMLs accordingly based on your cluster version.
This PR updates the operator to fetch the K8s cluster version and pass it to the configmanager service when deploying a Vizier.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Skaffold deploy operator and verify that K8s version is fetched properly

### DIFF
--- a/src/operator/manager.go
+++ b/src/operator/manager.go
@@ -24,6 +24,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/discovery"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


src/operator/manager.go
```diff
@@ -77,11 +79,26 @@ func main() {
...
+	k8sVersion := ""
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(kubeConfig)
+	if err != nil {
+		log.WithError(err).Error("Failed to get discovery client for config")
+	} else {
+		version, err := discoveryClient.ServerVersion()
+		if err != nil {
+			log.WithError(err).Error("Failed to get server version from discovery client")
+		} else {
+			k8sVersion = strings.TrimPrefix(version.GitVersion, "v")
```
vihangm, line 92: hmm, do we do this elsewhere too? I wonder if we should use [ParseTolerant](https://pkg.go.dev/github.com/blang/semver/v4#ParseTolerant) instead? seems a bit less brittle that way.

vihangm (approved):

src/operator/manager.go
aimichelle: We do trim the "v" in other places, but agree ParseTolerant is better. Going to change this PR to not modify the string, and use ParseTolerant in configmgr service.

vihangm (approved):

PR #886, opened 2023-02-16
Ignore thread_struct when resolving task_struct offsets on ARM

JamesMBartlett:
Summary: When using our pre-packaged linux headers on ARM, the `task_struct_resolver` was failing because it found the `group_leader` value twice in the `task_struct`.
The extra occurence of the address of the task struct (which is what we use to search for the `group_leader` offset), is found inside the `cpu_context` inside the `thread_struct`. On ARM, linux will store the address of the task struct in register `x20` when it creates a thread. This will then get stored in the 2nd field of `cpu_context` inside of `thread_struct`, leading to our task struct resolver thinking it found `group_leader` twice. To avoid this issue, this PR attempts to ignore the `thread_struct` entirely when resolving `task_struct` offsets. The issue then becomes identifying the offset of the `thread_struct` within the `task_struct`. Since the `cpu_context` stores potentially arbitrary register values, its not a good candidate for identifying the `thread_struct`. The next field in the `thread_struct` is the `tp_value`, which stores the address of thread local storage. This address should be consistent and we can access it from userspace, so this seems like a good candidate for finding the `thread_struct`. The only issue is that there can potentially be padding between the `cpu_context` and the `tp_value`. This PR looks for the `tp_value` by looking for the address of thread local storage, within the `task_struct` buffer. From the machines, I've looked at, I've only ever seen 8 bytes of padding between `cpu_context` and `tp_value`, so I decided to hardcode 8 bytes of padding for now. Luckily, we shouldn't be too sensitive to this padding, because the values we care about in the `task_struct` should never be right at the end just before the `thread_struct`, so if we overestimate the padding, it shouldn't cause any issues. Once we've found the `thread_struct` offset we can just ignore all the buffer data after that offset. This change doesn't affect behaviour on x86_64, only on aarch64.

Relevant Issues: #147

Type of change: /kind bug

Test Plan: I tested that `task_struct_resolver_bpf_test` works on an ARM machine now. Also, tested that skaffold deploying to an ARM cluster without preinstalled headers works now.

### DIFF
--- a/src/stirling/bpf_tools/task_struct_resolver.cc
+++ b/src/stirling/bpf_tools/task_struct_resolver.cc
@@ -137,8 +137,59 @@ StatusOr<TaskStructOffsets> ScanBufferForFields(const struct buf& buf,
 
   TaskStructOffsetsManager offsets_manager(&task_struct_offsets);
 
+#if AARCH64
+  // On aarch64, the cpu_context struct inside the thread_struct, can contain the same address as
+  // the group_leader value of the task_struct. Since the thread_struct is always at the end of the
+  // task_struct, we can just stop searching when we get to the thread_struct. To identify the
+  // offset of the thread_struct, we search for the address of thread local storage which is always
+  // stored `sizeof(cpu_context) + some padding` after the beginning of the thread struct.
+  // See:
+  // https://github.com/torvalds/linux/blob/3ac88fa4605ec98e545fb3ad0154f575fda2de5f/arch/arm64/include/asm/processor.h#L147
+  // for the thread_struct definition.
+  uint64_t tls_addr;
+  // ARM magic to get the system register TPIDR_EL0, which stores the address of thread local
+  // storage.
+  asm volatile("mrs %0, TPIDR_EL0" : "=r"(tls_addr));
+  VLOG(1) << "TLS address = " << tls_addr;
+
+  // We leave some room for padding. The thread_struct almost always seems to have some padding
+  // between cpu_context and tp_value. In the case that there is no padding, it shouldn't affect
+  // us since the values we care about are never right before the thread_struct in the
+  // task_struct.
+  constexpr uint64_t padding = 8;
+  // aarch64's cpu_context struct in linux stores 13 registers. I checked this in v4.14 and v6.1
+  // and it hasn't changed.
+  constexpr uint64_t num_cpu_context_registers = 13;
+  constexpr uint64_t tp_value_offset_in_thread_struct =
+      padding + num_cpu_context_registers * sizeof(uint64_t);
+  // For now we iterate through the buffer twice. At some point we should refactor offsets_manager
+  // to eliminate the need to iterate through the buffer twice.
+  int stop_at_offset = -1;
+  for (uint64_t idx = tp_value_offset_in_thread_struct / sizeof(uint64_t);
+       idx < sizeof(buf.u64words) / sizeof(uint64_t); idx++) {
+    int current_offset = idx * sizeof(uint64_t);
+    if (buf.u64words[idx] == tls_addr) {
+      VLOG(1) << absl::Substitute("[offset = $0] Found tp_value", current_offset);
+      stop_at_offset = current_offset - tp_value_offset_in_thread_struct;
+      VLOG(1) << "Ignoring thread_struct at offset = " << stop_at_offset;
+      break;
+    }
+  }
+  if (stop_at_offset == -1) {
+    return error::Internal(
+        "Failed to find the thread_struct offset within the task_struct. This is required for "
+        "resolving task struct offsets on aarch64");
+  }
+#endif
+
   for (const auto& [idx, val] : Enumerate(buf.u64words)) {
     int current_offset = idx * sizeof(uint64_t);
+
+#if AARCH64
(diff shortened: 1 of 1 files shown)


src/stirling/bpf_tools/task_struct_resolver.cc
```diff
@@ -137,8 +137,54 @@ StatusOr<TaskStructOffsets> ScanBufferForFields(const struct buf& buf,
...
+      // task_struct.
+      constexpr int padding = 8;
+      // aarch64's cpu_context struct in linux stores 13 registers. I checked this in v4.14 and v6.1
+      // and it hasn't changed.
+      constexpr int num_cpu_context_registers = 13;
+      constexpr int tp_value_offset_in_thread_struct =
+          padding + num_cpu_context_registers * sizeof(uint64_t);
+      stop_at_offset = current_offset - tp_value_offset_in_thread_struct;
+      VLOG(1) << "Ignoring thread_struct at offset = " << stop_at_offset;
+      break;
```
etep, line 170: It's possible this is a never required, but perhaps good practice:
```
if (tp_value_offset_in_thread_struct <= current_offset) {
    stop_at_offset = current_offset - tp_value_offset_in_thread_struct;
    VLOG(1) << "Ignoring thread_struct at offset = " << stop_at_offset;
    break;
}
```

Related comment on line 154.

src/stirling/bpf_tools/task_struct_resolver.cc
```diff
@@ -137,8 +137,54 @@ StatusOr<TaskStructOffsets> ScanBufferForFields(const struct buf& buf,
...
+  // offset of the thread_struct, we search for the address of thread local storage which is always
+  // stored `sizeof(cpu_context) + some padding` after the beginning of the thread struct.
+  uint64_t tls_addr;
+  // ARM magic to get the system register TPIDR_EL0, which stores the address of thread local
+  // storage.
+  asm volatile("mrs %0, TPIDR_EL0" : "=r"(tls_addr));
+  VLOG(1) << "TLS address = " << tls_addr;
+  // For now we iterate through the buffer twice. At some point we should refactor offsets_manager
+  // to eliminate the need to iterate through the buffer twice.
+  int stop_at_offset = -1;
```
etep, line 153: (note: it won't let me attach this comment to line 154 on the "right hand side" of the diff).

Related, but, it seems that the buffer index must be at least 14 and the byte offset 112 to find the tls addr. We could re-write line 154 like so:
```
for (const auto& [idx, val] : Enumerate(buf.u64words + tp_value_offset_in_thread_struct / sizeof(uint64_t))) { ...
```
I'm not sure if that compiles, but there might be some way to start the iteration at a crafted location using some arithmetic like above.

src/stirling/bpf_tools/task_struct_resolver.cc
```diff
@@ -137,8 +137,54 @@ StatusOr<TaskStructOffsets> ScanBufferForFields(const struct buf& buf,
...
+    return error::Internal(
+        "Failed to find the thread_struct offset within the task_struct. This is required for "
+        "resolving task struct offsets on aarch64");
+  }
+#endif
+
+  for (const auto& [idx, val] : Enumerate(buf.u64words)) {
+    int current_offset = idx * sizeof(uint64_t);
+
+#if AARCH64
```
etep, line 183: And I see waht you mean about iterating over the buffer twice. Perhaps we should inline the ARM logic right now?

src/stirling/bpf_tools/task_struct_resolver.cc
```diff
@@ -137,8 +137,54 @@ StatusOr<TaskStructOffsets> ScanBufferForFields(const struct buf& buf,
 
   TaskStructOffsetsManager offsets_manager(&task_struct_offsets);
 
+#if AARCH64
+  // On aarch64, the cpu_context struct inside the thread_struct, can contain the same address as
```
etep, line 141: I'm not sure if there is a good reference to include, but if so, maybe a pointer to that. It would be neat to have a link that pulls up the `thread_struct` source code.

src/stirling/bpf_tools/task_struct_resolver.cc
JamesMBartlett: So I think the best way to do this would be to change the OffsetsManager to keep around a list of candidates for each offset, have a function to set an ignored region of offsets on the offsets manager, and then resolve the candidate lists to a single offset at the end. 

However, I kind of want to defer this change for now because its a pretty big change to the way the OffsetsManager works. Additionally, my current diff has the nice property that there's zero change to behaviour on x86_64, so there's no worry about regressions there.

src/stirling/bpf_tools/task_struct_resolver.cc
JamesMBartlett: No longer necessary after doing your other suggestion.

src/stirling/bpf_tools/task_struct_resolver.cc
JamesMBartlett: What you suggested doesn't quite work because of the Enumerate, but I did something to this effect.

etep (approved):

ddelnano (approved):

vihangm (approved):

PR #889, opened 2023-02-16
Add timeout to OTLP export

aimichelle:
Summary: This PR adds a timeout to our OTLP span/metrics exports. It also adds a metrics counter for tracking how many OTLP timeouts have occurred, if any.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Skaffold deploy Vizier

### DIFF
--- a/src/carnot/exec/otel_export_sink_node.cc
+++ b/src/carnot/exec/otel_export_sink_node.cc
@@ -49,7 +49,7 @@ const int64_t kOTelTraceIDLength = 16;
 
 const int64_t kB3ShortTraceIDLength = 8;
 
-const int64_t otelClientTimeoutS = 5;
+constexpr auto otelClientTimeout = std::chrono::seconds{5};
 
 std::string OTelExportSinkNode::DebugStringImpl() {
   return absl::Substitute("Exec::OTelExportSinkNode: $0", plan_node_->DebugString());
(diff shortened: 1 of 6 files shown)


src/carnot/exec/otel_export_sink_node.cc
```diff
@@ -49,6 +49,8 @@ const int64_t kOTelTraceIDLength = 16;
 
 const int64_t kB3ShortTraceIDLength = 8;
 
+const int64_t otelClientTimeoutS = 5;
```
JamesMBartlett, line 52: nit: just make this `std::chrono::seconds` and then you can get rid of the `S` in the name

JamesMBartlett: Also do you have an idea of how long these requests usually take?

JamesMBartlett (approved):

src/carnot/exec/otel_export_sink_node.cc
aimichelle: I tried timing it on a cluster with our demo apps + load_gens running, and also tried running a longer query (send HTTP spans from the last 10 mins), and it was within the .3-.7 seconds range.

src/carnot/exec/otel_export_sink_node.cc
```diff
@@ -49,6 +49,8 @@ const int64_t kOTelTraceIDLength = 16;
 
 const int64_t kB3ShortTraceIDLength = 8;
 
+constexpr auto otelClientTimeout = std::chrono::seconds{5};
```
zasgar, line 52: Instead of hardcoding, do you think we can pass it in with some default value through the node?

zasgar (changes requested):

zasgar (approved):

PR #896, opened 2023-02-17
perf_profiler/notes.md: added to document the perf profiler.

etep:
Summary: We add `perf_profiler/notes.md` to document the perf profiler. In this patch, we document missing stack frames that result from sampling a stack trace at function entry or function exit. We will extend this document in future patches.

Type of change: /kind document

Test Plan: This is a document.


src/stirling/source_connectors/perf_profiler/notes.md
```diff
@@ -0,0 +1,209 @@
...
+                     +---------------+
+                     |  return addr. |   # This return address will not get picked up into the stack trace.
+-------------------- +---------------+
+Leaf fn.    rsp -->  |  saved bp_2   |   # points to saved bp_1
+                     +---------------+
+```
+__Figure 2.__ The stack and register state at the moment of function entry or function exit. The stack will be in this state when the instruction pointer points either `push %rbp` (on function entry) or at `ret` (on function exit).
+
+##### Understanding the missing stack frame
+We can see that if register `rpb` is dereferenced with the stack and register state as shown in Figure (2), then `saved bp_2` is skipped and so the return address that points into caller 1 is skipped. Succinctly, the stack is walked back starting from `rbp` which points to `saved bp_1`.
```
ddelnano, line 106: `rpb` => `rbp`

src/stirling/source_connectors/perf_profiler/notes.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Perf profiler organization
+TODO(jps): describe the various source code files and their respective purposes.
+
+### Symbolization notes
+TODO(jps): cover symbolization for compiled & Java and symbol caching.
+
+### Missing stack frames
+While investigating issue [#719](https://github.com/pixie-io/pixie/issues/719), we noticed that infrequently (less than 0.5% of the time, e.g.) we observed a stack trace with a missing stack frame. Fixing the test case to accommodate the issue was easy, but not informative. Here, we document our investigation into the root cause of the issue.
+
```
ddelnano, line 11: What do you think about adding the dot visualization of the missing stack frame here? I think it helps to ease into the explanation so that it's very clear what the issue was.

src/stirling/source_connectors/perf_profiler/notes.md
```diff
@@ -0,0 +1,209 @@
...
+```
+4. The saved base pointers each, in turn, point to their caller's stack. Find the next caller's base pointer:
+5. If `saved_rbp` is not a valid stack address, exit, else goto step (3).
+
+In pseudo code:
+```
+fp = *rbp;
+stack_trace.push_back(rip);
+while(true) {
+  fn_return_address = *(saved_rbp + 8);
```
ddelnano, line 47: I realize this is pseudo-code, but shouldn't `saved_rbp` be updated to be equal to `fp` on each loop iteration?

src/stirling/source_connectors/perf_profiler/notes.md
```diff
@@ -0,0 +1,209 @@
...
+# Steps used for perf.
+perf record -F 201 -p <pid> --call-graph <fp|dwarf> sleep 30
+perf script < perf.data > perf.txt
+cat perf.txt | c++filt | gprof2dot.py -f perf -e 0 | dot -Tpdf -o output.pdf
+```
+
+For these experiments, we linked with `ld` vs. our toolchain default of `lld` because [`perf` + `dwarf` does not work with `lld`](https://github.com/llvm/llvm-project/issues/53156).
+
+##### Linux perf users group
+We also sent email to the Linux perf users group and in two different replies, found confirmation of this issue. Editing for brevity:
```
ddelnano, line 121: Can we add a link to the linux-perf-users group [thread](https://lore.kernel.org/linux-perf-users/CAK2+a3RUFEb4ooe9BehBnBkOKgVt8n4Z28Ryzr2E4SH0--iiWg@mail.gmail.com/T/#t)?

src/stirling/source_connectors/perf_profiler/notes.md
```diff
@@ -0,0 +1,209 @@
...
+                     |  return addr. |
+-------------------  +---------------+
+Caller_1    rbp -->  |  saved bp_1   |   # points to saved bp_0
+                     +---------------+
+                     |  ............ |
+                     +---------------+
+                     |  return addr. |   # This return address will not get picked up into the stack trace.
+-------------------- +---------------+
+Leaf fn.    rsp -->  |  saved bp_2   |   # points to saved bp_1
+                     +---------------+
```
ddelnano, line 101: This picture didn't match my mental model of how I thought function entry would look. In the case that we sample before function entry is complete, shouldn't the `rbp` be on the stack "lower" than the `rsp`? Meaning we executed the `push`, but not the `mov` instruction of function entry?

src/stirling/source_connectors/perf_profiler/notes.md
etep: Yes, that is true! I had that variable named `saved_rbp` originally, but then thought it would be good to align to Linux source code. Going with: `s/saved_rbp/fp`.

src/stirling/source_connectors/perf_profiler/notes.md
etep: Will try to explain what I have, and see if it matches your understanding (what you wrote does make sense, and where needed, I'll add some detail).

Your description of the situation seems right. `push` was executed, `mov` was not. I think for a `push` instruction, the stack pointer is decremented, but the convention is that the stack pointer points to the "top" of the stack (not one beyond the top).  Although the previous base pointer is indeed at the top of the stack, that doesn't help the stack trace walker because it cannot use the stack pointer to find the base pointer. Rather, it just reads register `rbp` and that is its base pointer. Here, `rbp` is pointing to the caller's stack frame not the current function's stack frame. For that reason, it immediately walks back one and starts the stack trace with `Caller_0` (i.e. using the illustration in this doc).

src/stirling/source_connectors/perf_profiler/notes.md
etep: Thank you!

src/stirling/source_connectors/perf_profiler/notes.md
ddelnano: What you explained makes sense, but put another way here's what I'm having trouble reconciling: If the stack and registers exist as mentioned in this diagram, how would we lose a non leaf symbol when stack walking?

The symptom we see is that the stack walker always jumps 1 layer, and with the way the diagram is currently written it makes me believe we would always lose the leaf symbol (not a layer between). Is that explanation sound?

If not, we don't need to block this doc on me fully understanding this. Even though I understand the concept of the issue, the inner details are still a bit fuzzy for me :)

etep: The stack trace we want is this:
```
ip -> return addr. 0 -> return addr. 1
```
But because `ip` points to `push` while `bp` points into the caller's stack, we get the following:

1. Symbol is from the leaf function (the push instruction is in the leaf function), 
2. We skip return addr. 0 (because we pickup return addr. 1 using the base pointer pointing into the caller's stack), and hence have a missing stack frame (the stack trace goes from leaf to the caller's caller).

The stack trace we get is this:
```
ip -> return addr. 1
```

ddelnano: Thanks for that explanation! That clears up my confusion 👍

src/stirling/source_connectors/perf_profiler/notes.md
etep: I added the call graph rendered by pprof (my current opinion is that pprof provides a better cleaner visualization).

ddelnano (approved):

aimichelle (approved):

PR #897, opened 2023-02-17
Allow timeout to be specified in OTel endpoint config

aimichelle:
Summary: #889 adds a timeout to the OTel collector request. This PR allows the user to specify the timeout, which is propagated to the plan node. Once this PR is merged, #889 can be updated to use the timeout from the plan node, and if none is specified, uses a default.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Updated unit tests


src/carnot/planner/objects/otel.cc
```diff
@@ -323,14 +325,15 @@ Status OTelModule::Init(CompilerState* compiler_state, IR* ir) {
...
-                                         {{"headers", "{}"}, {"insecure", "False"}},
-                                         /* has_variable_len_args */ false,
-                                         /* has_variable_len_kwargs */ false,
-                                         std::bind(&EndpointConfigDefinition, std::placeholders::_1,
-                                                   std::placeholders::_2, std::placeholders::_3),
-                                         ast_visitor()));
+  PX_ASSIGN_OR_RETURN(
+      std::shared_ptr<FuncObject> endpoint_fn,
+      FuncObject::Create(kEndpointOpID, {"url", "headers", "insecure", "timeout"},
+                         {{"headers", "{}"}, {"insecure", "False"}, {"timeout", "0"}},
```
JamesMBartlett, line 331: I think it makes sense to have a non-zero default timeout.

src/carnot/planner/distributedpb/distributed_plan.proto
```diff
@@ -137,6 +137,8 @@ message OTelEndpointConfig {
   map<string, string> headers = 2;
   // True if the OpenTelemetry collector is not protected with SSL.
   bool insecure = 3;
+  // The out of time in seconds before the request to the OTel collector should timeout.
```
JamesMBartlett, line 140: nit: amount

also should probably mention the default timeout here.

JamesMBartlett (approved):

PR #898, opened 2023-02-17
[2 of 2] Fix & re-enable perf profiler test: accomodate leaf node aliasing in histogram.

etep:
Summary: We fix and re-enable the flaky perf profiler test. The underlying issue was stack trace aliasing due to missing stack frames. We document this issue in https://github.com/pixie-io/pixie/pull/896. In brief, because the stack trace walker sometimes skips a stack frame, the test logic could find two different stack traces that matched by leaf symbol. We previously did not realize this, and hence we used assign (vs. accumulate) when building the leaf symbol histogram. In this patch, we accommodate the missing stack frames by accumulating when building the leaf symbol histogram.

After the switch to `podman`, the Java tests started failing, on Jenkins, based on expected sample rate. It is hard to predict the actual number of samples that will occur for a Java process because we do not have direct control over the number of processes and threads. Since we still have coverage for sample rate in cpp and go, we split the expectations into two parts and check sample rate only for cpp and go.

Test Plan: Tested locally using `--runs_per_test=32`. All tests passed. Verified with local testing that we did see test failures and flakiness without the change.

### DIFF
--- a/src/stirling/source_connectors/perf_profiler/BUILD.bazel
+++ b/src/stirling/source_connectors/perf_profiler/BUILD.bazel
@@ -58,8 +58,6 @@ pl_cc_test(
         "cpu:16",
         "exclusive",
         "requires_bpf",
-        # Disabling this test because it has a 25% pass rate on main at the moment.
-        "disabled_flaky_test",
     ],
     deps = [
         ":cc_library",
--- a/src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
+++ b/src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
@@ -98,7 +98,7 @@ absl::flat_hash_map<std::string, uint64_t> KeepNLeafSyms(
     const auto end_iter = symbols.end();
 
     const auto leaf_syms = absl::StrJoin(begin_iter, end_iter, ";");
-    leaf_histo[leaf_syms] = count;
+    leaf_histo[leaf_syms] += count;
   }
   return leaf_histo;
 }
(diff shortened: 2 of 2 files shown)


src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -453,7 +453,7 @@ TEST_F(PerfProfileBPFTest, PerfProfilerCppTest) {
 }
 
 // TODO(jps/oazizi): This test is flaky.
```
ddelnano, line 455: Sounds like we can remove this comment as well as part of re-enabling

ddelnano (approved):
One comment about a TODO that can be removed (referenced the flakiness we fixed 🎉), otherwise lgtm.

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
etep: Thank you!

src/stirling/source_connectors/perf_profiler/BUILD.bazel
```diff
@@ -53,9 +53,7 @@ pl_cc_test(
         "//src/stirling/source_connectors/perf_profiler/testing/java:graal-vm-aot-fib",
     ],
     defines = ['JDK_IMAGE_NAMES=\\"%s\\"' % ",".join(jdk_image_names)],
-    shard_count = 3,
     tags = [
-        "cpu:16",
```
ddelnano, line 58: Did you mean to change these? I thought we wanted to keep these in place.

etep: That crept in. Fixing it now. (thank you!)

etep:
> Seems like the test is still flaky?

I looked at the 3 failed builds. It seems that it is Java test #8 failing in the same way each time on kernel 5.19. That is disabled now with a note to investigate.

etep:
> > Seems like the test is still flaky?
> 
> I looked at the 3 failed builds. It seems that it is test Java/8 failing in the same way each time on kernel 5.19. That is disabled now with a note to investigate.

I'm going to update the comment to reflect that the issue is specific to kernel v5.19.

etep:
@JamesMBartlett 
[Test 377](https://jenkins.px.dev/blue/organizations/jenkins/pixie-oss%2Fbuild-and-test-pr/detail/build-and-test-pr/377/pipeline/)
[Test 378](https://jenkins.px.dev/blue/organizations/jenkins/pixie-oss%2Fbuild-and-test-pr/detail/build-and-test-pr/378/pipeline/)
passed twice now with Java/8 disabled.

etep:
Observed more flakiness with local testing.

zasgar (approved):

PR #902, opened 2023-02-18
Switch from docker to podman for running tests

JamesMBartlett:
Summary: Switch from docker to podman for running bpf tests.
The outer `run_docker` script still uses docker, but the tests themselves run containers with podman.

Podman doesn't have a daemon like docker does. Since podman is daemonless, it makes it much easier to run inside of our kernel emulated QEMU environment.

Type of change: /kind cleanup

Test Plan: All bpf tests pass under podman.


src/stirling/source_connectors/socket_tracer/redis_trace_bpf_test.cc
```diff
@@ -87,8 +87,8 @@ std::vector<RedisTraceRecord> GetRedisTraceRecords(
 TEST_F(RedisTraceBPFTest, VerifyBatchedCommands) {
   StartTransferDataThread();
 
-  constexpr std::string_view kRedisDockerCmdTmpl =
-      R"(docker run --rm --network=container:$0 redis bash -c "echo '$1' | redis-cli")";
+  constexpr std::string_view kRedisPodmanCmdTmpl =
+      R"(podman run --rm --network=container:$0 docker.io/redis bash -c "echo '$1' | redis-cli")";
```
ddelnano, line 91: I take it the `docker` cli special cases the docker hub registry whereas podman doesn't?

src/stirling/source_connectors/perf_profiler/symbolizers/symbolizer_test.cc
```diff
@@ -309,31 +296,18 @@ TEST_F(BCCSymbolizerTest, JavaEnoughSpaceAvailable) {
...
 
-  // Start the container/sub-proc. Use "-v" arg. to mount our tmpfs into /tmp in that container.
-  const std::string tmpfs_mount_opt = absl::Substitute("$0:/tmp", volume_name);
+  // Start the container/sub-proc.
   static constexpr auto timeout = std::chrono::seconds{600};
   static constexpr bool kUseHostPidNamespace = false;
-  const std::vector<std::string> options = {"-v", tmpfs_mount_opt};
+  const std::string tmpfs_size = "20M";
+  const std::vector<std::string> options = {
+      "--mount", absl::Substitute("type=tmpfs,tmpfs-size=$0,destination=/tmp", tmpfs_size)};
```
ddelnano, line 310: Didn't realize this was an option and it's great that this ensures volumes are cleaned up when the container is removed.

ddelnano (approved):
One question to clarify my understanding of the explicit prefixing of `docker.io/` to the official container image names but otherwise lgtm.

etep:
It might be nice to include more detail in the diff summary of why we are switching to podman.

src/stirling/source_connectors/perf_profiler/symbolizers/symbolizer_test.cc
etep: Also nice to have a few LOC removed.

src/stirling/source_connectors/socket_tracer/redis_trace_bpf_test.cc
JamesMBartlett: correct

src/stirling/source_connectors/perf_profiler/symbolizers/symbolizer_test.cc
JamesMBartlett: Yeah, was also necessary because `podman volume create` doesn't have the same support for sized volumes as docker.

aimichelle (approved):

Issue #692, opened 2023-01-17
Add protocol tracing support for applications using BoringSSL 


ddelnano:
Pixie currently supports tracing protocol traffic encrypted with certain TLS libraries (OpenSSL version 1.1.0 or 1.1.1, dynamically linked and Go TLS when a binary has [debug info](https://docs.px.dev/reference/admin/debug-info/)). This gives broad coverage, but there are other popular TLS libraries that are not supported as of today (BoringSSL being one of them). The recent work to trace netty tls traffic (#407) uncovered some common challenges to support BoringSSL more broadly.

This issue will track the work to enhance Pixie's TLS protocol tracing to include applications that use BoringSSL.

ddelnano:
I'm in the process of drafting a design document for this work. Once the first draft is finished I'll be sharing the document here.

ddelnano:
This is complete as of the v0.14.3 Vizier release!

PR #903, opened 2023-02-21
Update the socket tracer metrics to differentiate between tls and plaintext metrics

ddelnano:
Summary: This updates the `SocketTracerMetrics` class to differentiate between plaintext and tls metrics. Since we are working to instrument BoringSSL based applications more broadly (#692), this will allow us to experiment with the underlying tls tracing implementation and verify that there aren't protocol parsing issues introduced (indicated by more data loss).

Relevant Issues: #692

Type of change: /kind feature

Test Plan: Updated the `mux_trace_bpf_test` and `netty_tls_trace_bpf_test` with the following diff to verify that they increment the correct counter ([P317](https://phab.corp.pixielabs.ai/P317))
- [x] Inspect the counters values and dimensions despite end to end test mentioned above
- [x] Verified that this data was not used for any previous purposes (new dimension would likely cause breakage)

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/metrics.cc
+++ b/src/stirling/source_connectors/socket_tracer/metrics.cc
@@ -19,10 +19,30 @@
 #include <memory>
 #include <string>
 #include <unordered_map>
+#include <utility>
 
 #include <magic_enum.hpp>
 
 #include "src/stirling/source_connectors/socket_tracer/metrics.h"
+
+using metrics_key = std::pair<traffic_protocol_t, bool>;
+
+namespace std {
+
+// Provides hash specialization since stdlib doesn't provide one for std::pair.
+// This uses the first 32 bits for the traffic_protocol_t enum and the final 32
+// bits for the boolean value.
+template <>
+struct hash<metrics_key> {
+  std::size_t operator()(const metrics_key& p) const {
+    uint64_t protocol = p.first;
+    int32_t b = p.second ? 1 : 0;
+
+    return (protocol << 32) | b;
+  }
+};
+
+}  // namespace std
 
 namespace px {
 namespace stirling {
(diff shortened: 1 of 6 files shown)


src/stirling/source_connectors/socket_tracer/metrics.cc
```diff
@@ -28,38 +28,59 @@ namespace px {
...
 
 namespace {
-std::unordered_map<traffic_protocol_t, std::unique_ptr<SocketTracerMetrics>> g_protocol_metrics;
+std::unordered_map<traffic_protocol_t, std::unique_ptr<SocketTracerMetrics>>
+    g_plaintext_protocol_metrics;
 
-void ResetProtocolMetrics(traffic_protocol_t protocol) {
-  g_protocol_metrics.insert_or_assign(
-      protocol, std::make_unique<SocketTracerMetrics>(&GetMetricsRegistry(), protocol));
+std::unordered_map<traffic_protocol_t, std::unique_ptr<SocketTracerMetrics>>
```
JamesMBartlett, line 55: Instead of multiple maps can you just create a single map with a compound key

JamesMBartlett (changes requested):

src/stirling/source_connectors/socket_tracer/metrics.cc
ddelnano: I considered this but I was worried about causing memory allocations. I tried to use a `std::pair` as the key, but it seemed like I would need to construct a new object each time I accessed a key since the caller of this code wouldn't have a `std::pair`.

Therefore I chose an implementation that seemed safer. If you think that can be avoided or isn't a concern, I'm definitely open to using a compound key.

JamesMBartlett: If you put the `std::pair` on the stack it won't cause any memory allocations. And then whether or not the branch is worse or the couple of instructions for the construction of the pair, seems like a toss up to me, and the pair is probably worth it for keeping the code simpler.

JamesMBartlett: Another alternative if you really want to go for performance is to bit pack the enum and the bool into a 64 bit integer key. Since there's no way our enum can have 2**63 values.

etep: Would it be reasonable to create a flat array of metrics for each protocol with and w/out ssl? My mental model of a metric suggests that the memory cost of this will be low, and it could lead to simpler reasoning about perf and possibly simpler code also (i.e. removes any of the cruft that comes with handling map corner cases).

ddelnano: I incorrectly assumed that `std::make_pair` was going to allocate on the heap when I first looked at this.

> Would it be reasonable to create a flat array of metrics for each protocol with and w/out ssl?
I would prefer to use a complex key since my initial assumption about there being an allocation was incorrect. Unless there are still performance concerns, I think the readability is much nicer.

I needed to provide a specialization in order for the `unordered_map` to understand how to hash a `std::pair`, so I used the enum and bool packing mentioned above as part of this.

JamesMBartlett: `std::make_pair` will probably allocate but you can do:
```
std::pair<traffic_protocol_t, bool> key(protocol, ssl);
```
and that will not.

ddelnano: I see. I ended up using what you suggested with my latest update, so let me know if there are any concerns with that.

JamesMBartlett (approved):

ddelnano:
I was able to complete the final test I wanted to perform, so this is ready for review now.

PR #907, opened 2023-02-24
Expose conn tracker creation through prometheus metrics and migrate the existing stats interface to it

ddelnano:
Summary: Expose conn tracker creation through prometheus metrics and migrate the existing stats interface to it

As part of an upcoming change to expand pixie's tls support for BoringSSL (#692), we want to track conn tracker's lifecycle to have a proxy measurement for if a connection's socket file descriptor is identified correctly. Our TLS tracing requires this since it is fundamental to a connection's identity. When our tls tracing implementation is changed, we expect the `conn_tracker_created` metric will stay at its existing baseline. Any increases in this metric (assuming our protocol tracing support remains constant) would indicate that the file descriptors are inferred incorrectly.

Relevant Issues: https://github.com/pixie-io/pixie/issues/692

Type of change: /kind feature

Test Plan: Existing conn_tracker tests pass which rely on our counters being correct
- [x] Verify that metrics are visible in testing

### DIFF
--- a/src/common/metrics/metrics.h
+++ b/src/common/metrics/metrics.h
@@ -38,3 +38,12 @@ inline auto& BuildCounter(const std::string& name, const std::string& help_messa
       .Register(GetMetricsRegistry())
       .Add({{"name", name}});
 }
+
+// A convenience wrapper to return a counter with the specified name and help message.
+inline auto& BuildGauge(const std::string& name, const std::string& help_message) {
+  return prometheus::BuildGauge()
+      .Name(name)
+      .Help(help_message)
+      .Register(GetMetricsRegistry())
+      .Add({{"name", name}});
+}
--- a/src/common/metrics/metrics.h
+++ b/src/common/metrics/metrics.h
@@ -39,7 +39,7 @@ inline auto& BuildCounter(const std::string& name, const std::string& help_messa
       .Add({{"name", name}});
 }
 
-// A convenience wrapper to return a counter with the specified name and help message.
+// A convenience wrapper to return a gauge with the specified name and help message.
 inline auto& BuildGauge(const std::string& name, const std::string& help_message) {
   return prometheus::BuildGauge()
       .Name(name)
--- a/src/stirling/source_connectors/socket_tracer/conn_trackers_manager.cc
+++ b/src/stirling/source_connectors/socket_tracer/conn_trackers_manager.cc
@@ -112,10 +112,6 @@ uint64_t GetConnMapKey(uint32_t pid, int32_t fd) { return (static_cast<uint64_t>
 
 ConnTrackersManager::ConnTrackersManager()
     : trackers_pool_(kMaxConnTrackerPoolSize),
-      total_conn_trackers_(
-          BuildGauge("total_conn_trackers", "Total number of conn trackers in use")),
-      ready_for_destruction_(BuildGauge("ready_for_destruction",
-                                        "Number of conn trackers that are ready for destruction")),
       conn_tracker_created_(BuildCounter("conn_tracker_created",
                                          "Counter that tracks when a conn tracker is created")),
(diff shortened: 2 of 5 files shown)


ddelnano:
One test is experiencing a seg fault in tcmalloc from what appears to be some llvm stdlib code. I need to track that down, but for now I'd like feedback on the overall approach in this pull request.

cc @etep

JamesMBartlett:
IMO using the values of Prometheus metrics within our core logic is coupling two separate things that shouldn't be coupled. 

In general, I think metrics should be something you can completely disable and still have the rest of the code function fine.

ddelnano:
@JamesMBartlett that logic makes sense and I think that is sensible to do. At a first glance, I thought the original stats bordered the line of being internal logic but I agree those two things shouldn't be coupled.

src/stirling/source_connectors/socket_tracer/conn_trackers_manager.cc
```diff
@@ -109,7 +110,18 @@ uint64_t GetConnMapKey(uint32_t pid, int32_t fd) { return (static_cast<uint64_t>
 
 }  // namespace
 
-ConnTrackersManager::ConnTrackersManager() : trackers_pool_(kMaxConnTrackerPoolSize) {}
+ConnTrackersManager::ConnTrackersManager()
+    : trackers_pool_(kMaxConnTrackerPoolSize),
+      total_conn_trackers_(
+          BuildGauge("total_conn_trackers", "Total number of conn trackers in use")),
+      ready_for_destruction_(BuildGauge("ready_for_destruction",
```
JamesMBartlett, line 117: I suspect this gauge might be a bit noisy. My understanding is that ready-for-destruction conn trackers are cleaned up relatively quickly (at least on a time scale of seconds rather than minutes). At the moment, prometheus metrics are only sampled once a minute, so given this gauge is measuring something that's timescale is much smaller than that, it might be pretty noisy. Depending on what you want to use the number for, it might be more useful to keep a counter of the high water mark of conn trackers that are ready for destruction. Perhaps, you might also want a high water mark for total conn trackers, although I think sampling total conn trackers at once a minute is much more reasonable.

JamesMBartlett (approved):

src/stirling/source_connectors/socket_tracer/conn_trackers_manager.cc
ddelnano: I agree with your concern and I didn't have immediate plans for the two gauge metrics added in this PR. The counter metrics are the ones I plan to use to validate the tls changes I'm working on. I decided to instrument the gauge metrics since it filled out what we track with our internal stats interface. However, maybe we should just omit these for now since it may be noisy.

Is there any way to filter out metrics from the prometheus scraping side? Wondering if it's possible to have the instrumentation there but enable it later if we decided it was valuable.

src/common/metrics/metrics.h
```diff
@@ -38,3 +38,12 @@ inline auto& BuildCounter(const std::string& name, const std::string& help_messa
       .Register(GetMetricsRegistry())
       .Add({{"name", name}});
 }
+
+// A convenience wrapper to return a counter with the specified name and help message.
```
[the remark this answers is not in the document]
ddelnano: Fixed in latest update.

src/stirling/source_connectors/socket_tracer/conn_trackers_manager.cc
JamesMBartlett: We don't currently have a good way to filter them out on the scraping side. If you think you might use them, I think its fine to leave them in, otherwise I would just remove the instrumentation for now.

ddelnano: I went ahead and removed them since I don't see any concrete need for now.

etep (approved):

PR #909, opened 2023-02-27
Remove uses of bazel info in scripts and tests

vihangm:
Summary: The uses of `bazel info` to find built binaries were
inaccurate in some places. Additionally calling out to bazel
in a test causes issues when running under qemu. This cleans up
`bazel info` calls in non-test helpers and updates the tests
to not invoke bazel and expect the relevant args to always be
supplied.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: All the updated scripts and tests should continue to work.


vihangm:

zasgar (approved):

Issue #709, opened 2023-01-30
BPF tests should be easier to run and be able to run without root 


zasgar:
BPF tests currently require root access to run and break out of the bazel sandbox. 

We can try using something like qemu to provide an isolated environment that can be run with bazel support.

JamesMBartlett:
`bazel test --config=qemu-bpf` will now run BPF tests inside of qemu. This configuration still requires the user to have `kvm` permissions but no longer requires root.

PR #930, opened 2023-03-01
Update kernel config to add support for ftrace and posix timers

zasgar:
Summary: Our tests need support for syscall tracing and posix timers to run postgres.

Relevant Issues: #709

Test Plan: Tested on qemu.


JamesMBartlett (approved):

ddelnano:
Do you remember what the issue was with running postgres before timers were enabled? Just curious to understand what behavior we saw beforehand.

PR #939, opened 2023-03-01
Improve Command Palette and Breadcrumb sorting

NickLanam:
Summary: There were still a few scenarios where the filtering was a bit too aggressive, or the sorting of results didn't make much sense. This should fix quite a few of them.

Test Plan: Try all of the following searches in the `script` breadcrumb, and a few of them in the Command Palette as well.
They should both get similar results...

- The search `exec` should now find `bpftrace/EXEC_snoop` before it finds weaker matches
- The search `pod` should now greatly prefer scripts that have the string `pod` in them, even over shorter-named scripts that have vaguely similar spelling.
- The search `px/ace` should now do the same: `PX/namespACEs` should now be a stronger match than `PX/nodE`.
- And the same for `px/stat`.
- Finally, most searches should keep more results than before (further past "definitely relevant" into "eh, it might be what you meant" territory), but the most relevant ones should still come up first.

Type of change: /kind bug

### DIFF
--- a/src/ui/src/utils/string-search.ts
+++ b/src/ui/src/utils/string-search.ts
@@ -18,6 +18,18 @@
 
 const ALLOWED_CHARS = new Set('abcdefghijklmnopqrstuvwxyz01234567890/'.split(''));
 
+// Scoring by relevance involves a bit of math based on what properties a match has.
+// These numbers were determined experimentally, by trying a variety of search strings against
+// the full list of PxL script names until the ordering consistently got desirable results.
+const PREFIX_DISTANCE_FACTOR = 0.2;
+const SUBSTRING_DISTANCE_FACTOR = 0.25;
+const MAX_DISTANCE_RATIO = 0.8;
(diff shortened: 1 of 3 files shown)


src/ui/src/utils/string-search.ts
```diff
@@ -183,18 +183,23 @@ export function highlightScoredMatch(search: string, source: string): ScoredStri
     // Exact prefix is even better than exact substring, but not as good as a perfect match
     return {
       isMatch: true,
-      distance: Math.abs(searchNorm.length - sourceNorm.length) / 4,
+      distance: Math.abs(searchNorm.length - sourceNorm.length) / 5,
```
aimichelle, line 186: Is the 5 a magic number? How did you decide on it?

NickLanam: Extracted the magic numbers.

aimichelle (approved):

PR #943, opened 2023-03-02
Add NATSErrorCounter for tracking NATS errors

aimichelle:
Summary: The metrics server is known to heavily drop NATs messages due to a `slow consumer` error. We track the count of dropped messages for both vzconn and vzmgr, but don't track it for the metrics server yet. This PR adds the NATSErrorCounter so we can easily track nats errors in our cloud services.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Deploy to a testing cloud, ensure metrics server doesn't crash.

### DIFF
--- a/src/cloud/metrics/metrics_server.go
+++ b/src/cloud/metrics/metrics_server.go
@@ -22,8 +22,11 @@ import (
 	"context"
 	"net/http"
 	_ "net/http/pprof"
+	"strings"
 
 	"cloud.google.com/go/bigquery"
+	"github.com/nats-io/nats.go"
+	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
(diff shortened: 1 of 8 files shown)


src/cloud/metrics/metrics_server.go
```diff
@@ -39,12 +42,33 @@ import (
...
 
+var (
+	natsErrorCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "nats_error_count",
+		Help: "NATS message bus error",
+	}, []string{"shardID", "vizierID", "messageKind", "errorKind"})
+)
+
+// Extracts information from the subject and return shard, vizierID, messageType.
+func extractMessageInfo(subject string) (string, string, string) {
```
ddelnano, line 53: What's the string expected to look like for the subscription? I believe the metrics server's subscription is created [here](https://github.com/pixie-io/pixie/blob/518183204f9f350ad42d064e77b39ada306e8d12/src/cloud/metrics/controllers/server.go#L124), but I'm not sure if that's what this function would receive.

aimichelle: It will look something like `v2c.<shardID>.<vizierID>.VZMetrics`.
v2c stands for vizier to cloud, so a message originating from a Vizier to be sent to the cloud. 
`shardID` is the shard which the vizier sending the message belongs to, so that we can handle the subscriptions by shard. 
`vizierID` is the vizier of the ID sending the message.
`VZMetrics` is the channel which specifies the type of message.

We use this message format in our other services as well. for example vzmgr: https://github.com/pixie-io/pixie/blob/518183204f9f350ad42d064e77b39ada306e8d12/src/cloud/vzmgr/controllers/server.go#L118

ddelnano (approved):

aimichelle:
Just repushed (with the same changes) because the DCO check seemed to be stuck.

src/cloud/shared/messages/nats_metrics.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"strings"
+
+	"github.com/nats-io/nats.go"
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+)
+
+// NatsErrorCount counts the number of errors that occurs over NATs.
```
JamesMBartlett, line 29: I think it would be a bit cleaner to wrap this in a struct, something like:
```
type NatsErrorCounter struct {
   count *prometheus.CounterVec
}

func NewNatsErrorCounter() {
  c := ....
  prometheus.MustRegister(c.count)
}

func (c *NatsErrorCounter) HandleNatsError
```
And then each server can create a new nats error counter in its init function, this way the global state isn't in the shared file.

JamesMBartlett (approved):
Maybe change the title/description to reflect the new changes too

PR #953, opened 2023-03-03
No need to call init on podman command

zasgar:
Summary: Init is not needed here and was making podman in qemu unhappy.

Relevant Issues: #709

Test Plan: Jenkins, no functional change.

### DIFF
--- a/src/stirling/e2e_tests/stirling_wrapper_container_bpf_test.sh
+++ b/src/stirling/e2e_tests/stirling_wrapper_container_bpf_test.sh
@@ -48,7 +48,7 @@ run_uprobe_target "$go_grpc_server" "$go_grpc_client"
 echo "Running stirling_wrapper container."
 
 flags="--timeout_secs=0"
-out=$(podman run --init --rm \
+out=$(podman run --rm \
  -v /:/host \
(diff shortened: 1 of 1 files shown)


JamesMBartlett (approved):

ddelnano (approved):

src/stirling/e2e_tests/stirling_wrapper_container_bpf_test.sh
```diff
@@ -48,7 +48,7 @@ run_uprobe_target "$go_grpc_server" "$go_grpc_client"
 echo "Running stirling_wrapper container."
 
 flags="--timeout_secs=0"
-out=$(podman run --init --rm \
```
ddelnano, line 51: What error did you specifically see? Curious to know how it manifested.

zasgar: > What error did you specifically see? Curious to know how it manifested.

Error: container-init binary not found on the host: stat /usr/libexec/podman/catatonit: no such file or directory^M

We can probably add that to our sysroot, but it will just be bloat.

PR #957, opened 2023-03-03
Move the redis tests to use containers from bazel

zasgar:
Summary: Move the redis tests to use containers from bazel instead of pulling them
from the registry. This allows the tests to run without internet access and not be subject to
pull limits.

This changes also removes some output assertions because redis does not always produce message when running outside
of a terminal environemnt. The assertions aren't important for coverage since we check the test expectations as well.

Relevant Issues: #709

Type of change: /kind cleanup

Test Plan: existing tests should pass.


JamesMBartlett (approved):

ddelnano:
Overall changes look good, but can you do a final test run with `--runs_per_test` set to ~20-25? Since this test was recently stabilized, it would be good to verify that there isn't any new flakinesses.

PR #961, opened 2023-03-03
[perf_tool] Add BatchInserter for bigquery

JamesMBartlett:
Summary: BigQuery `Inserter().Put` calls can be quite slow, so we don't want to run that for every single metric row we receive during an experiment.
This adds a utility that reads from a channel of arbitrary data, and batches it up before sending it to bigquery. This reduces the number of `Put` calls.

Test Plan: Tested that rows sent to the BatchInserter show up in bigquery. Also tested that if not enough rows are sent the timeout will trigger a push anyway.

### DIFF
--- a/src/e2e_test/perf_tool/pkg/bq/batch_inserter.go
+++ b/src/e2e_test/perf_tool/pkg/bq/batch_inserter.go
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2018- The Pixie Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


src/e2e_test/perf_tool/pkg/bq/batch_inserter.go
```diff
@@ -0,0 +1,110 @@
...
+			return
+		case <-t.C:
+			b.uploadBatch()
+			t.Reset(b.PushTimeout)
+		default:
+			if b.batchSize.Load() >= int64(b.BatchSize) {
+				b.uploadBatch()
+				if !t.Stop() {
+					<-t.C
+				}
```
vihangm, line 86: I think you can remove the need for mutex-es and a separate watchdog by combining these.

something like:
```
func (b* BatchInserter) Run(data <- ch interface{}) {
  for {
    select {
    case <-t.C:
      b.Upload()
    case datum := <-data:
      if len(batch) >= maxLen:
        b.Upload()
      b.Add(datum)
    }
  }
}
```

src/e2e_test/perf_tool/pkg/bq/batch_inserter.go
```diff
@@ -0,0 +1,110 @@
...
+				t.Reset(b.PushTimeout)
+			} else {
+				time.Sleep(5 * time.Second)
```
vihangm, line 89: why is this sleep here?

vihangm (changes requested):

zasgar (approved):

PR #963, opened 2023-03-03
Use curl image for sock shop load test init.

etep:
Summary: We change the container image for sock shop load test init from `alpine:3.6` to `gcr.io/pixie-oss/pixie-dev-public/curl:multiarch-7.87.0`. The new image comes with curl baked in. This avoids a potential problem where (for unknown reasons) the init pod fails because apk fails to install curl. Our curl image is created from `tools/docker/curl_image/Dockerfile`.

Type of change: /kind bug fix

Test Plan: Tested locally by applying the sock shop demo yamls on my dev cluster. The new load test yaml worked (as expected).

### DIFF
--- a/demos/sock-shop/sock-shop-loadgen.yaml
+++ b/demos/sock-shop/sock-shop-loadgen.yaml
@@ -28,10 +28,10 @@ spec:
 
       initContainers:
       - name: wait-sock-shop
-        image: alpine:3.6
+        image: gcr.io/pixie-oss/pixie-dev-public/curl:1.0
         # yamllint disable rule:indentation
-        command: ['sh', '-c', 'set -x;  apk add --no-cache curl;
-          until timeout -t 2 curl -f "${SOCK_SHOP_HEALTH_ADDR}"; do
+        command: ['sh', '-c', 'set -x;
+          until timeout 2 curl -f "${SOCK_SHOP_HEALTH_ADDR}"; do
             echo "waiting for ${SOCK_SHOP_HEALTH_ADDR}";
             sleep 2;
           done;']
--- a/demos/sock-shop/sock-shop-loadgen.yaml
+++ b/demos/sock-shop/sock-shop-loadgen.yaml
@@ -28,7 +28,7 @@ spec:
 
       initContainers:
       - name: wait-sock-shop
-        image: gcr.io/pixie-oss/pixie-dev-public/curl:1.0
+        image: gcr.io/pixie-oss/pixie-dev-public/curl:multiarch-7.87.0
         # yamllint disable rule:indentation
         command: ['sh', '-c', 'set -x;
           until timeout 2 curl -f "${SOCK_SHOP_HEALTH_ADDR}"; do


demos/sock-shop/sock-shop-loadgen.yaml
```diff
@@ -28,10 +28,10 @@ spec:
 
       initContainers:
       - name: wait-sock-shop
-        image: alpine:3.6
+        image: gcr.io/pixie-oss/pixie-dev-public/curl:1.0
```
JamesMBartlett, line 31: can you use the multiarch version instead: `gcr.io/pixie-oss/pixie-dev-public/curl:multiarch-7.87.0`

JamesMBartlett (changes requested):

demos/sock-shop/sock-shop-loadgen.yaml
etep: Where is the dockerfile for that one?

JamesMBartlett: It's a meta image that points to the same dockerfile, but with arm and x86 versions that are picked based on the architecture of the docker platform

JamesMBartlett: Oh wait actually, we used the official `curlimages/curl` for this multiarch image

demos/sock-shop/sock-shop-loadgen.yaml
```diff
@@ -28,10 +28,10 @@ spec:
 
       initContainers:
       - name: wait-sock-shop
-        image: alpine:3.6
+        image: gcr.io/pixie-oss/pixie-dev-public/curl:multiarch-7.87.0
         # yamllint disable rule:indentation
-        command: ['sh', '-c', 'set -x;  apk add --no-cache curl;
-          until timeout -t 2 curl -f "${SOCK_SHOP_HEALTH_ADDR}"; do
+        command: ['sh', '-c', 'set -x;
+          until timeout 2 curl -f "${SOCK_SHOP_HEALTH_ADDR}"; do
```
JamesMBartlett, line 34: why did the `-t` change?

JamesMBartlett (approved):

demos/sock-shop/sock-shop-loadgen.yaml
etep: Something about `timeout` was different between the two. The new one did not like `-t`. I will retry with `-t` because things have evolved.

etep: Retried using`-t` and seeing the following:
```
>kc -n px-sock-shop logs load-test-76fb456564-mfb8r -c wait-sock-shop | head -n 20
+ timeout -t 2 curl -f http://orders.px-sock-shop.svc.cluster.local/health
timeout: unrecognized option: t
BusyBox v1.35.0 (2022-08-01 15:14:44 UTC) multi-call binary.

Usage: timeout [-s SIG] [-k KILL_SECS] SECS PROG ARGS

Run PROG. Send SIG to it if it is not gone in SECS seconds.
Default SIG: TERM.If it still exists in KILL_SECS seconds, send KILL.

+ echo 'waiting for http://orders.px-sock-shop.svc.cluster.local/health'
+ sleep 2
waiting for http://orders.px-sock-shop.svc.cluster.local/health
+ timeout -t 2 curl -f http://orders.px-sock-shop.svc.cluster.local/health
timeout: unrecognized option: t
BusyBox v1.35.0 (2022-08-01 15:14:44 UTC) multi-call binary.

Usage: timeout [-s SIG] [-k KILL_SECS] SECS PROG ARGS

Run PROG. Send SIG to it if it is not gone in SECS seconds.
Default SIG: TERM.If it still exists in KILL_SECS seconds, send KILL.
```

Issue #913, opened 2023-02-28
Mux protocol inference matches large amounts of non mux traffic


ddelnano:
When investigating metrics that track stitching and parsing data loss on one of my test clusters, I noticed that mux protocol has significant data loss compared to other protocols. The data loss for mux was in the 100-200 trillion range, while the protocol with the next highest data loss was in the ~10 billion range. Since I know that there aren't any mux protocol applications on this cluster, this indicates that pixie's protocol inference is mislabelling lots of unknown traffic as mux.

When the mux protocol inference was first implemented, we needed to scope it more tightly to prevent it from mislabelling the other protocols we support. The data mentioned above shows that there is more work to do here and highlights that this could cause a performance problem. When pixie matches traffic that it shouldn't, the protocol traffic is copied to user space when it could be skipped entirely.

ddelnano:
When this issue was created, we thought this misclassification would incur a performance impact. After reducing the misclassification in #1026 (see that PR for the magnitude of the reduction), we were unable to see any meaningful performance improvement. We believe this is related to stirling's ConnTracker object pool.

Closing this issue since further optimizing this is no longer necessary.

PR #968, opened 2023-03-06
Update data_collection protocol inference script to work with cri cli tool

ddelnano:
Summary: Update data_collection protocol inference script to work with cri cli tool

The previous implementation for the protocol inference data collection script relied on the docker cli. GKE clusters running k8s 1.24 and later do not support docker anymore, so this prevents the script from working on newer clusters. This change adds support for clusters using CRI (`crictl`) in addition to keeping the existing docker interface. The docker implementation had complicated parsing logic that was also simplified in this change.

Relevant Issues: #913 (script was used as part of the investigation)

Type of change: /kind bug

Test Plan: Followed the instructions in the protocol inference [docs](src/stirling/protocol_inference/README.md) and verified pcaps were successfully captured
```
ddelnano@turing:~/code/pixie$ bazel run src/stirling/protocol_inference:data_collection -- --node gke-testing-node --pods $(pwd)/pod_names.txt --duration 120

# Verify that pcap files are present on the host
$ ls -alh /captures/*
/captures/metrics-server-v0.5.2-6bf845b67f-pkncq:
total 440K
drwxr-xr-x 2 root root 4.0K Mar  6 17:54 .
drwxr-xr-x 4 root root 4.0K Mar  6 17:54 ..
-rw------- 1 root root 430K Mar  6 17:55 5540.pcapng

/captures/vizier-query-broker-6d5cddb8c5-jcq6m:
total 748K
drwxr-xr-x 2 root root 4.0K Mar  6 17:54 .
drwxr-xr-x 4 root root 4.0K Mar  6 17:54 ..
-rw------- 1 root root 738K Mar  6 17:55 2361577.pcapng
```

### DIFF
--- a/src/stirling/protocol_inference/data_collection.py
+++ b/src/stirling/protocol_inference/data_collection.py
@@ -19,29 +19,39 @@ import subprocess
 import json
 import argparse
 
+def create_pod2pid(runtime):
+    """
+    Builds a mapping of running pod names to host pids.
+    """
 
-def create_pod2pid():
-    """
-    Build a mapping of running pod names to host pids.
-    """
-    # This cmd gets the host pid of each running container and its corresponding labels.
-    list_ps_cmd = '''kubectl node-shell ''' + node + ''' -- bash -c "docker ps -q | \
-    xargs docker inspect --format '{{.State.Pid}}{{json .Config.Labels}}'"'''
+    # This command prints one line of output per container where each line is the pid of the container, a 
+    # single space and the name of the k8s pod determined by the io.kubernetes.pod.name container label.
+    # See the output below for example output:
+
+    # $ sudo crictl ps -q | sudo xargs crictl inspect --template '{{.info.pid}} {{index .info.config.labels "io.kubernetes.pod.name"}}' -o go-template
+    # 3635260 vizier-pem-sx7pr
+    # 3634678 kelvin-cdf78c57c-qlzlb
+
+    list_ps_cmd = '''kubectl node-shell ''' + node + ''' -- sudo bash -c "crictl ps -q | \
+    xargs crictl inspect -o go-template --template '{{.info.pid}} {{index .info.config.labels \\"io.kubernetes.pod.name\\"}}'"'''
+    if runtime == 'docker':
+      list_ps_cmd = '''kubectl node-shell ''' + node + ''' -- sudo bash -c "docker ps -q | \
+    xargs docker inspect --format '{{.State.Pid}} {{index .Config.Labels \\"io.kubernetes.pod.name\\"}}'"'''
 
     pod2pid = {}
     ps_out = subprocess.run(list_ps_cmd, shell=True, capture_output=True)
-    rows = str(ps_out.stdout).split('\\n')[1:-3]
+
+    rows = ps_out.stdout.decode('utf-8').splitlines()
     for row in rows:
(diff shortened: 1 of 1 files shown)


src/stirling/protocol_inference/data_collection.py
```diff
@@ -16,32 +16,44 @@
...
+        split_row = row.split()
+
+        if len(split_row) != 2:
+            print(f"Failed to parse config for row: {row}. Skipping to new pid")
             continue
-        pod_name = kube_info['io.kubernetes.pod.name']
+
+        pid, pod_name = split_row
         pod2pid[pod_name] = pid
+
```
etep, line 56: `pod2pid[pod_name] = int(pid)`?

src/stirling/protocol_inference/data_collection.py
```diff
@@ -16,32 +16,44 @@
...
+    #     '{{.info.pid}} {{index .info.config.labels "io.kubernetes.pod.name"}}' -o go-template
+    # 3635260 vizier-pem-sx7pr
+    # 3634678 kelvin-cdf78c57c-qlzlb
+
+    list_ps_cmd = '''kubectl node-shell ''' + node + ''' -- sudo bash -c "crictl ps -q | \
+    xargs crictl inspect -o go-template \
+    --template '{{.info.pid}} {{index .info.config.labels \\"io.kubernetes.pod.name\\"}}'"'''
+    if runtime == 'docker':
+        list_ps_cmd = '''kubectl node-shell ''' + node + ''' -- sudo bash -c "docker ps -q | \
+    xargs docker inspect --format '{{.State.Pid}} {{index .Config.Labels \\"io.kubernetes.pod.name\\"}}'"'''
```
etep, line 41: For this context, please consider the following construct:
```
ps_cmds = {
  'docker': docker_ps_cmd,
  'crio': crio_ps_cmd,
}
ps_cmd = ps_cmds[runtime]
```

Since this is not prod. code, this is a concise way to error check things, i.e. it reduces the possibility of picking up something wrong as new code is introduced. Granted this is not a place where we have a lot of change and churn, so maybe defaults and if/else is not as big of a hazard here. For non-perf critical prod code, I would still endorse the above scheme, but would craft things such that a missing map entry does not cause program termination.

src/stirling/protocol_inference/data_collection.py
```diff
@@ -87,5 +100,5 @@ def trace_pods(pod2pid, node, pods, duration):
     node = args.node
     duration = args.duration
 
-    pod2pid = create_pod2pid()
+    pod2pid = create_pod2pid(args.container_runtime)
     trace_pods(pod2pid, node, pods, duration)
```
etep, line 104: Could convert to using `fire` ... optional, but, it is a very concise way to handle cmd line args.

src/stirling/protocol_inference/data_collection.py
ddelnano: The rest of the program works fine using the pid as a string. I'd prefer to not introduce casting unless it's necessary.

src/stirling/protocol_inference/data_collection.py
ddelnano: Updated to use an approach in spirit with your suggestion. I added a bit more templating since I felt it made dealing with the various arguments and quoting more readable.

Let me know what you think.

etep (approved):

JamesMBartlett (approved):

PR #971, opened 2023-03-06
[perf_tool] Add bigquery storage format definitions

JamesMBartlett:
Summary: Add definitions for how both results and specs will be stored in bigquery. Specs and results will be stored in separate tables. The spec table is only written to on experiment success, so joining the spec and result tables on the ExperimentID will remove all result rows from failed experiments.

Test Plan: Tested that bigquery tables get created correctly based on the implicit schemas of these structs.


aimichelle (approved):

JamesMBartlett:

PR #980, opened 2023-03-07
Add QEMU kernel deps

zasgar:
Summary: This adds kernel deps for the QEMU testing build.
It currently only adds support for one kernel, but will plan to add the other kernels shortly.

Relevant Issues: #709

Test Plan: jenkins, bazel builds.

### DIFF
--- a/bazel/pl_workspace.bzl
+++ b/bazel/pl_workspace.bzl
@@ -62,6 +62,8 @@ def pl_workspace_setup():
     k8s_repositories()
     k8s_go_deps(go_version = None)
 
+    qemu_with_kernel_deps()
+
 def pl_container_images():
     _package_manager_setup()
     _container_images_setup()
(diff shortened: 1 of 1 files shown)


bazel/pl_workspace.bzl
```diff
@@ -82,3 +84,18 @@ def pl_model_files():
         downloaded_file_path = "sentencepiece.proto",
         sha256 = "7e17e04ecc207d9204dc8755357f988bf77c135f7a34a88984943c8649d6a790",
     )
+
+def qemu_with_kernel_deps():
+    http_file(
+        name = "linux_build_5_18_19_x86_64",
```
JamesMBartlett, line 90: It's unclear from the name what this actually is. Is it the `bzImage`? If so can you call it `linux_bzImage_...`

JamesMBartlett: Talked offline, this name is fine since its more than just bzImage

JamesMBartlett (approved):

PR #981, opened 2023-03-07
Update protocol inference scripts to take into account tls packets

ddelnano:
Summary: Update protocol inference scripts to take into account tls packets

While investigating #913, I deployed stirling with verbose logging to identify which processes are contributors to the mux protocol confusion: kubelet, metrics-server, query broker, cloud connector ([P325](https://phab.corp.pixielabs.ai/P325)). After fixing the data collection script (#968) and using it on these pods, I realized that it didn't change the output of the protocol confusion matrix. This is because the dataset generation script ignores anything that is encrypted ([source](https://github.com/pixie-io/pixie/blob/9d16ad1f218beb2b01e879192859b4a0e99adfce/src/stirling/protocol_inference/dataset_generation.py#L179)). Essentially if `tshark` doesn't recognize the protocol as one of [these](https://github.com/pixie-io/pixie/blob/9d16ad1f218beb2b01e879192859b4a0e99adfce/src/stirling/protocol_inference/dataset_generation.py#L136-L137) it is labeled as unknown and skipped.

Our ideal confusion matrix would label tls as `unknown` 100% of the time and should not be labeled as any other protocol. With this change, we can evaluate modifications to the protocol inference and work to get mux's `tls` classification down to 0%.

Relevant Issues: #913

Test Plan: Generated the confusion matrix (seen below) and saw the mux's mislabelling of tls traffic with our existing data set

conn_dataset_with_tls

While 0.6% doesn't seem like a large amount of traffic, this amounts to 126 packets from our data set some of which are quite large.

### DIFF
--- a/src/stirling/protocol_inference/dataset_generation.py
+++ b/src/stirling/protocol_inference/dataset_generation.py
@@ -134,7 +134,7 @@ def parse_protocol(frame_protocol):
     """
     # Putting http after http2/grpc since it can overlap with http2 and grpc.
     protocols = ["http2", "mysql", "pgsql", "cql", "amqp", "redis", "dns", "mongo", "http", "ssh",
-                 "kafka", "mux"]
+                 "kafka", "mux", "tls"]
     for protocol in protocols:
         if protocol in frame_protocol:
             return protocol
--- a/src/stirling/protocol_inference/model/metadata.py
+++ b/src/stirling/protocol_inference/model/metadata.py
@@ -29,4 +29,5 @@ kTargetProtocols = [
     "kafka",
     "nats",
     "mux",
+    "tls",
 ]


etep (approved):
This is pretty neat.

Even if 0.6% is "not that large" in my understanding, because the label is sticky, then the effect of this can amplified, i.e. in that one wrongly labeled connection can last a long time and represent a lot of network traffic.

ddelnano:
> Even if 0.6% is "not that large" in my understanding, because the label is sticky, then the effect of this can amplified, i.e. in that one wrongly labeled connection can last a long time and represent a lot of network traffic.

Exactly, long lived connections that are misclassified can cause huge amounts of data to be copied to user space. That's also why the `data_loss` metric for mux is many orders of magnitude higher (100-200 trillion) compared to other protocols (~10 billion) as mentioned in [this comment](https://github.com/pixie-io/pixie/issues/913#issue-1603420357).

The `src/stirling/protocol_inference:eval` binary allows you to pass an [--analysis](https://github.com/pixie-io/pixie/blob/4c27b15f58afebba7b0c99f70c6365d009adf038/src/stirling/protocol_inference/eval.py#L43) flag to identify specific payloads that were misclassified (defaults to showing [10 results](https://github.com/pixie-io/pixie/blob/4c27b15f58afebba7b0c99f70c6365d009adf038/src/stirling/protocol_inference/eval.py#L44)). When I ran that, the 10 connections resulted in a tsv of 213 MB. 
```
$ ls -alh mux-traffic-confusion/misclassification.tsv
-rw-rw-r-- 1 ddelnano ddelnano 213M Mar  7 11:12 mux-traffic-confusion/misclassification.tsv
```

JamesMBartlett (approved):

PR #982, opened 2023-03-07
[actions] Refactor actions to reuse common `get-image` action

JamesMBartlett:
Summary: Adds a `get_image.yaml` github action that can be reused by the various actions that need to look at `docker.properties` to get the image name for their subsequent job.
This reduces the need to repeat code for all the different actions that require our custom dev image.

Test Plan: Testing the actions with this PR.

### DIFF
--- a/.github/workflows/cacher.yaml
+++ b/.github/workflows/cacher.yaml
@@ -8,16 +8,9 @@ permissions:
   contents: read
 jobs:
   get-dev-image:
-    runs-on: ubuntu-latest
-    outputs:
-      image-with-tag: ${{ steps.get-version.outputs.image }}
-    steps:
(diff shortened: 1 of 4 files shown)


vihangm (commented):
Shouldn't get_image be a custom action instead of a workflow?

vihangm:
Or perhaps just a shell script

JamesMBartlett:
> Shouldn't get_image be a custom action instead of a workflow?

I'm following this guide: https://docs.github.com/en/actions/using-workflows/reusing-workflows

Seems unnecessary to make it a custom action to me, but if there's a reason you think its better I'm open to it.

vihangm (approved):

PR #1010, opened 2023-03-08
[perf_tool/runner] Teardown in correct order

JamesMBartlett:
Summary: The `defer *.Close` statements don't happen in the correct order, so this explicitly calls the `Close`s in the order they should be torn down.

Test Plan: Tested the `Runner` implementation as a whole


JamesMBartlett:

PR #1026, opened 2023-03-08
Enforce stricter checks on mux's TDispatch and RDispatch frames during protocol inference

ddelnano:
Summary: Enforce stricter checks on mux's TDispatch and RDispatch frames during protocol inference

This change addresses a large portion of the mux protocol misclassification identified in #913. As verified in #981, mux protocol traffic is very likely to be matched against tls traffic. This change addresses the 0.6% of traffic mislabelled in our test dataset.

Relevant Issues: #913

Type of change: /kind bug

Test Plan: Performed the following tests outlined below
- [x] Verified the protocol inference scripts no longer mislabel the 0.6% of tls traffic identified in #981
- [x] Deployed this change to my testing cluster and verified that mux data loss metric is no longer increasing by ~100 trillion every hour. The second graph decreases the y axis max so the "missing" data points can be seen.
<p float="left">

Screen Shot 2023-03-08 at 2 33 03 PM
Screen Shot 2023-03-08 at 8 07 17 AM
</p>

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
+++ b/src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
@@ -553,6 +553,11 @@ static __inline enum message_type_t infer_mux_message(const char* buf, size_t co
   static const int8_t kRerr = -128;
   static const int8_t kRerrOld = 127;
   uint32_t mux_header_size = 8;
+
+  static const uint8_t kReplyStatusOk = 0;
+  static const uint8_t kReplyStatusError = 1;
+  static const uint8_t kReplyStatusNack = 2;
+
   // TODO(ddelnano): Determine why mux-framer text in T/Rinit is
   // 6 bytes after the mux header
   int32_t mux_framer_pos = mux_header_size + 6;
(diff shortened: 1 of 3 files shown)


aimichelle (approved):

src/stirling/source_connectors/socket_tracer/bcc_bpf/protocol_inference.h
```diff
@@ -601,6 +606,32 @@ static __inline enum message_type_t infer_mux_message(const char* buf, size_t co
...
+    // com.twitter.finagle.tracing.{TraceContext, ClientIdContext, Deadline} are
+    // the only ones that exist today.
+    if (buf[first_ctx_pos] != 'c' || buf[first_ctx_pos + 1] != 'o' ||
+        buf[first_ctx_pos + 2] != 'm' || buf[first_ctx_pos + 3] != '.' ||
+        buf[first_ctx_pos + 4] != 't' || buf[first_ctx_pos + 5] != 'w' ||
+        buf[first_ctx_pos + 6] != 'i' || buf[first_ctx_pos + 7] != 't' ||
+        buf[first_ctx_pos + 8] != 't' || buf[first_ctx_pos + 9] != 'e' ||
+        buf[first_ctx_pos + 10] != 'r')
+      return kUnknown;
+  }
```
JamesMBartlett, line 626: This doesn't seem vendor neutral to me?

ddelnano: While mux is an open source protocol, it was developed by Twitter and it's only implementation is [finagle](https://github.com/twitter/finagle).

I understand that's not ideal, but the likelihood of mux having an additional set of well known context keys to support is very low.

We must parse [twitter](https://github.com/pixie-io/pixie/blob/2ba99cba01d2d9f4a5f7ffc91e2fafacba62999b/src/stirling/source_connectors/socket_tracer/protocols/mux/parse.cc#L81) specific [context](https://github.com/pixie-io/pixie/blob/2ba99cba01d2d9f4a5f7ffc91e2fafacba62999b/src/stirling/source_connectors/socket_tracer/protocols/mux/parse.cc#L74) keys [in](https://github.com/pixie-io/pixie/blob/2ba99cba01d2d9f4a5f7ffc91e2fafacba62999b/src/stirling/source_connectors/socket_tracer/protocols/mux/parse.cc#L92) order to access certain interesting protocol fields.

PR #1029, opened 2023-03-09
Use batchInserter in metrics service

aimichelle:
Summary: We regularly see that the metrics service drops ~20 NATS messages a second. From doing some exploratory logging, we send out the timeseries data as a "batch" to Bigquery. However, most of these timeseries only contain one datapoint, so we are actually only sending one datapoint at a time. We can instead send the metrics in batches which should speed up our NATS processing since we are less dependent on the BigQuery insert.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Skaffold deploy to a testing cloud instance and check that metrics are sent out.


JamesMBartlett (commented):
Can you fix the usages of `perf_tool/pkg/bq` inside `perf_tool?

src/cloud/metrics/controllers/server.go
```diff
@@ -85,11 +87,21 @@ func (s *Server) Start() {
...
+		Schema: schema,
+		Table:  table,
+	}
+
 	for _, shard := range vzshard.GenerateShardRange() {
-		bqWriteChan := make(chan []*bigquery.StructSaver, 2048)
+		bqWriteChan := make(chan interface{}, 4096)
+		bqBatchInserter := bq.BatchInserter{
+			BatchSize:   1024,
+			PushTimeout: 10 * time.Second,
```
JamesMBartlett, line 99: I would probably make the push timeout a bit longer. I think its fine to give up latency for less calls to bigquery.Put

JamesMBartlett (approved):

PR #1042, opened 2023-03-13
[perf_tool] Add github action to run perf_tool

JamesMBartlett:
Summary: Adds a github action to run performance evaluations. The github action can be triggered three ways:
- automatically each night
- manually through github's workflow dispatch
- on PRs by having an admin comment `/perf <suites> <tags>`

Test Plan: See #1037. I tested this by making a pull_request trigger for the action and testing on a separate pull request. We don't actually want that trigger so I removed it here.

### DIFF
--- a/.github/workflows/perf.yaml
+++ b/.github/workflows/perf.yaml
@@ -1,0 +1,113 @@
+---
+name: perf-eval
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+  # Run at 23:49 PST (07:49 UTC) every day. Github suggests not running actions on the hour.
+  - cron: '49 7 * * *'
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Branch or commit'
+        required: true
+        type: string
+      suite:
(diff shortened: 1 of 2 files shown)


.github/workflows/perf.yaml
```diff
@@ -0,0 +1,112 @@
...
+        args="${{ steps.command.outputs.command-arguments }}"
+        suites="$(echo ${args} | cut -d" " -f1)"
+        tags="$(echo "${args}" | cut -s -d" " -f2)"
+        echo "suites=${suites}" >> $GITHUB_OUTPUT
+        echo "tags=${tags}" >> $GITHUB_OUTPUT
+    - name: Add default tags
+      id: default-tags
+      # yamllint disable rule:indentation
+      run: |
+        default_tags="PR#${{ github.event.issue.number }},ignore"
```
vihangm: let's convert the ignore exclusion into a inclusion for cron based main runs only.

vihangm (approved):

PR #1044, opened 2023-03-13
Allow kernel version to be configurable for qemu

zasgar:
Signed-off-by: Zain Asgar <zasgar@pixielabs.ai>

Summary: This allows kernel version to be passed via a flag. There is a set list of kernels that can be executed and they are built using
//tools/docker/kernel_builder.

Relevant Issues: #709

Test Plan: `bazel test -c opt --config=qemu-bpf  --cache_test_results=no --remote_download_outputs=all //src/stirling/source_connectors/socket_tracer:redis_trace_bpf_test --test_output=streamed --//bazel/test_runners/qemu_with_kernel:kernel_version=4.14.304`
`--//bazel/test_runners/qemu_with_kernel:kernel_version` can be set to different values. There is a bug with the 4.14.304 kernel config that'll be fixed next, but the others are working.

### DIFF
--- a/bazel/pl_workspace.bzl
+++ b/bazel/pl_workspace.bzl
@@ -87,9 +87,27 @@ def pl_model_files():
 
 def qemu_with_kernel_deps():
     http_file(
+        name = "linux_build_4_14_304_x86_64",
+        url = "https://storage.googleapis.com/pixie-dev-public/kernel-build/20230228151027/linux-build-4.14.304.tar.gz",
+        sha256 = "58c474a6f3cf0c93af39d0549ddcb3e36b852026c42229fc6d4a40260e34329d",
+        downloaded_file_path = "linux-build.tar.gz",
+    )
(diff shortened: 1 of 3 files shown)


vihangm:
Why did you pick the given kernel versions? Should we instead use the latest patch versions for each LTE release?
https://github.com/pixie-io/pixie/blob/main/ci/bpf/00_create_instance.sh#L27-L38 for reference has those

zasgar:
> Why did you pick the given kernel versions? Should we instead use the latest patch versions for each LTE release? https://github.com/pixie-io/pixie/blob/main/ci/bpf/00_create_instance.sh#L27-L38 for reference has all the LTE releases (though not the latest patch versions).

That makes sense. Do you want to accept this and I will rework which kernel versions are used? I need to update the kernel config anyways.

vihangm (approved):

PR #1046, opened 2023-03-13
Always scroll hovered item into view in Command Palette

### LINKED ISSUE #1045: Keyboard scrolling in Command Palette can obscure results
**Describe the bug**
When using the arrow keys to look through a scrolling list of results in the Command Palette, and specifically scrolling up to an option under a heading, it will not scroll quite far enough up. This obscures the option, like so:
image

**To Reproduce**
Open the Command Palette. Enter a query like `script:`. Using the arrow keys, scroll down a couple of pages, then scroll back up the same way. When moving to an option that's above the current scroll view, it only partially scrolls into view.

**Expected behavior**
The highlighted option should scroll fully into view.

NickLanam:
Summary: Without this change, the browser already tries to scroll the selected option into view. However, due to `position: sticky`, there's a header that can obscure it. To account for that header, we use `scroll-margin-top` and force the browser to try again with `scrollIntoView`, which accounts for that property.
Fixes #1045

Relevant Issues: #1045

Test Plan: Open the Command Palette. Enter a query like script:. Using the arrow keys, scroll down a couple of pages, then scroll back up the same way. When moving to an option that's above the current scroll view, it should now fully scroll that option into view.

Type of change: /kind bug

### DIFF
--- a/src/ui/src/components/command-palette/command-text-field.tsx
+++ b/src/ui/src/components/command-palette/command-text-field.tsx
@@ -222,7 +222,7 @@ export const CommandTextField = React.memo<CommandTextFieldProps>(({
     onHighlightChange: (event: React.SyntheticEvent, option: CommandCompletion, reason: string) => {
       setHighlightedCompletion(option);
       if (option && (reason === 'keyboard' || reason === 'auto')) {
-        // The event target is the text field if the reaon is keyboard, and missing if the reason is auto.
+        // The event target is the text field if the reason is keyboard, and missing if the reason is auto.
         // So instead, we manually find the element that got focused and scroll it into view.
         // Accounts for scroll-margin-top and therefore the sticky headers.
         setTimeout(() => {
--- a/src/ui/src/components/command-palette/command-text-field.tsx
+++ b/src/ui/src/components/command-palette/command-text-field.tsx
@@ -132,6 +132,11 @@ const useFieldStyles = makeStyles((theme: Theme) => createStyles({
     '&:not(:first-child)': {
       borderTop: theme.palette.border.unFocused,
     },
+    '& > $optionGroupHeading + ul > li': {
+      // Tells scrollIntoView to scroll a bit further up, so it can avoid being obscured by the sticky headers.
+      // Only applies when the header actually exists within that option's group.
+      scrollMarginTop: theme.spacing(3.125),
+    },
   },
   optionGroupList: {
     listStyle: 'none',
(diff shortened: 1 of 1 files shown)


src/ui/src/components/command-palette/command-text-field.tsx
```diff
@@ -217,8 +219,18 @@ export const CommandTextField = React.memo<CommandTextFieldProps>(({
     onChange: (_, option: CommandCompletion) => { // Takes event, option, reason, details (in case we need them)
       activateCompletion(option);
     },
-    onHighlightChange: (_, option: CommandCompletion) => {
+    onHighlightChange: (event: React.SyntheticEvent, option: CommandCompletion, reason: string) => {
       setHighlightedCompletion(option);
+      if (option && (reason === 'keyboard' || reason === 'auto')) {
+        // The event target is the text field if the reaon is keyboard, and missing if the reason is auto.
```
JamesMBartlett, line 225: nit: reason

src/ui/src/components/command-palette/command-text-field.tsx
```diff
@@ -132,6 +132,11 @@ const useFieldStyles = makeStyles((theme: Theme) => createStyles({
     '&:not(:first-child)': {
       borderTop: theme.palette.border.unFocused,
     },
+    '& > $optionGroupHeading + ul > li': {
+      // Tells scrollIntoView to scroll a bit further up, so it can avoid being obscured by the sticky headers.
+      // Only applies when the header actually exists within that option's group.
+      scrollMarginTop: theme.spacing(3.125),
```
JamesMBartlett, line 138: This number seems a bit random to me, why 3.125?

src/ui/src/components/command-palette/command-text-field.tsx
```diff
@@ -132,6 +132,11 @@ const useFieldStyles = makeStyles((theme: Theme) => createStyles({
     '&:not(:first-child)': {
       borderTop: theme.palette.border.unFocused,
     },
+    '& > $optionGroupHeading + ul > li': {
```
JamesMBartlett, line 135: This is a bit incomprehensible to me, but I don't know CSS that well so maybe its common?

src/ui/src/components/command-palette/command-text-field.tsx
NickLanam: It matches the height of the headers defined above (computing the value would not be straightforward).

src/ui/src/components/command-palette/command-text-field.tsx
NickLanam: Yes, this is pretty common stuff plus two Material-specific things:

- `& ...` means "prepend the rule that got us here", which is just a shorthand.
- `$anotherRule` is also a Material thing. It makes sure to copy over the generated class name correctly.
- 'foo > bar` means "a bar that is a direct child of a foo"
- `foo + bar` means "a bar that is the immediate next sibling of a foo"

All in all, this means "list items that are in groups that have headers" (as the comment implies).

JamesMBartlett (approved):

PR #1054, opened 2023-03-14
[actions] Add build-and-test github action

JamesMBartlett:
Summary: Adds a github action to run builds and tests, including bpf tests. The action uses the configs defined in `ci/github/bazelrc` along with the `bazel_build_deps.sh` script to decide what tests to run and under what configurations to run them.

Test Plan: Tested in this PR that all the tests (including BPF tests) work. Also, tested that failing tests are reported properly, by adding an `ASSERT_EQ(0, 1)` only in opt configs, and saw that only opt configs failed.

### DIFF
--- a/ci/github/matrix.yaml
+++ b/ci/github/matrix.yaml
@@ -1,0 +1,51 @@
+---
+configs:
+- name: Clang opt + UI
+  runner: nokvm
+  args: "--config=clang --compilation_mode=opt"
+  buildables: bazel_buildables_clang_opt
+  tests: bazel_tests_clang_opt
+- name: Clang debug
+  runner: nokvm
+  args: "--config=clang --compilation_mode=dbg"
+  buildables: bazel_buildables_clang_dbg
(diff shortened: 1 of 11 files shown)


ci/github/matrix.yaml
```diff
@@ -0,0 +1,51 @@
...
+  name: BPF ASAN (latest kernel)
+  args: "--config=asan --config=qemu-bpf --compilation_mode=opt"
+  buildables: bazel_buildables_bpf_sanitizer
+  tests: bazel_tests_bpf_sanitizer
+- disabled: true
+  runner: kvm
+  name: BPF TSAN (latest kernel)
+  args: "--config=tsan --config=qemu-bpf --compilation_mode=opt"
+  buildables: bazel_buildables_bpf_sanitizer
+  tests: bazel_tests_bpf_sanitizer
```
vihangm, line 51: can we start these configs with `name` and then `disabled`
IMO harder to parse right now.

vihangm (approved):

PR #1058, opened 2023-03-15
[perf-eval] Don't error on comments without slash command

JamesMBartlett:
Summary: Skip the rest of the workflow if `pr-perf-setup` fails, instead of failing the workflow.

Test Plan: Testing with a comment in this PR.


JamesMBartlett:
test comment

PR #1062, opened 2023-03-15
[1 of 4] Add `UnitConnector` to Stirling, useful as an API to one source connector.

etep:
Summary: We add class `UnitConnector` to wrap and provide an API to a source connector. This wrapper provides the necessary logic to drive a source connector in isolation (i.e. without using `stirling.cc`). It will be used for test fixtures (e.g. in the socket tracer and perf profiler) and for individual binaries (for the profiler, and for planned work on record and replay binaries).

Type of change: /kind feature

Test Plan: This code is not used (yet). In the next set of PRs, we will switch over test fixtures and stirling binaries to use the `UnitConnector`. Continued passing tests will be the test plan as we switch code over. See #1065.

### DIFF
--- a/src/stirling/core/data_tables.h
+++ b/src/stirling/core/data_tables.h
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2018- The Pixie Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include "src/stirling/core/data_table.h"
+#include "src/stirling/core/types.h"
+
+namespace px {
+namespace stirling {
+
+class DataTables {
+ public:
+  template <size_t NumTableSchemas>
+  explicit DataTables(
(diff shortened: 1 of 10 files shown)


src/stirling/core/data_tables.h
```diff
@@ -0,0 +1,53 @@
...
+#include <utility>
+#include <vector>
+
+#include "src/stirling/core/data_table.h"
+#include "src/stirling/core/types.h"
+
+namespace px {
+namespace stirling {
+
+class DataTables {
```
ddelnano, line 31: Will we be using this class outside of any test scenarios? It seems like this class's responsibility is still test focused, so it seems slightly out of place to be in `src/stirling/core`.

ddelnano:
@etep can you explain how this drives the source connector without `stirling.cc`? From looking at how this is used in #1065, it appears this is largely replacing the `SocketTraceBPFTestFixture` methods with calls to an underlying member (as pictured below). 

Screen Shot 2023-03-20 at 9 46 16 AM

src/stirling/core/data_tables.h
ddelnano: @etep did you see my previous comment about this?

ddelnano: We talked about this offline. I missed that this was used in the UnitConnector 😅

ddelnano (approved):

src/stirling/core/unit_connector.h
```diff
@@ -0,0 +1,389 @@
...
+#include "src/stirling/core/data_tables.h"
+#include "src/stirling/core/frequency_manager.h"
+
+DEFINE_string(pids, "", "PIDs to profile, e.g. -pids 132,133. All processes profiled by default.");
+
+namespace px {
+namespace stirling {
+
+template <typename T>
+class UnitConnector {
```
JamesMBartlett, line 36: It seems like this is only used for testing? And only planned to be used for testing + dev binaries? So should it live in a testing folder instead of in `stirling/core`?

etep:
In the stirling sub-tree, we have not marked any of the cc_binaries as testonly, except for benchmarks. This suggestion marks a departure. Previously did not view the binaries (such as stirling_profiler) as being test binaries.

aimichelle (approved):

PR #1039, opened 2023-03-10
Make http_body_limit_bytes able to be set by environment variable

AdityaAtulTewari:
Summary: Changing the value of `max_body_bytes` by setting the `PL_STIRLING_MAX_BODY_BYTES` does not allow for sizes over `http_body_limit_bytes` for the body of ebpf traffic captured in the http setting. I've added an environment variable for setting `http_body_limit_bytes` using `PX_STIRLING_MAX_BODY_BYTES`.

Test Plan: Deployed Stirling in a self-hosted environment with a variety of values.

Type of change: /kind cleanup

Signed-off-as: AdityaAtulTewari <adityaatewari@gmail.com>

PR #1071, opened 2023-03-20
[ci/bazel_build_deps] Remove tab character in cquery command

JamesMBartlett:
Summary: For some reason there was a tab character in the bazel_build_deps script. For some unknown reason this tab character would cause `pipefail` to cause an error some of the time. Haven't really figured out why, but I tested with the changes in #1039, that used to cause a failure, and without the tab character the script works on those changes.

Type of change: /kind cleanup

Test Plan: Tested that the changes in #1039 no longer cause a failure in this script


JamesMBartlett:
Seems like this still fails

PR #1075, opened 2023-03-21
Add disableAutoUpdate to cloud connector env variables

aimichelle:
Summary: Auto-update disabling broke with the removal of one of our configmaps, which was done as cleanup.
This PR adds the auto-update env var to the cloudconnector template YAML so that if a user deploys with autoUpdate as true, this will be passed in properly.

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Generate the Vizier YAML templates, inspect the template to ensure it looks correct.


JamesMBartlett (approved):

aimichelle:
For some reason I recalled that the `pl-cluster-config` was cleaned up, but looks like it still exists. So there is likely another problem here--closing out this PR since it's not necessary.

PR #1085, opened 2023-03-23
Prevent qemu test runner false positive builds by adding a kvm/qemu validation check

ddelnano:
Summary: Prevent qemu test runner false positive builds by adding a kvm/qemu validation check

The qemu test runner silently fails when kvm is not accessible. This causes bazel to believe all tests have passed when the tests weren't started as seen in this output ([P331](https://phab.corp.pixielabs.ai/P331)). This change addresses this false positive by verifying that qemu/kvm are functional before attempting to run a VM.

Relevant Issues: #709

Test Plan: Verified that bazel reports failed tests when a qemu vm fails to start ([P330](https://phab.corp.pixielabs.ai/P330))

This test was performed when my user account could not access `/dev/kvm`(missing kvm group access).

### DIFF
--- a/bazel/test_runners/qemu_with_kernel/run_qemu.sh
+++ b/bazel/test_runners/qemu_with_kernel/run_qemu.sh
@@ -76,6 +76,11 @@ flags+=(-append "console=ttyS0 root=/dev/sda")
 # Disable graphics mode.
 flags+=(-nographic)
 
+if ! virt-host-validate qemu -q; then
+  echo "'virt-host-validate qemu' failed to validate qemu is functional. Check the output of the 'virt-host-validate' command and resolve any issues before retrying";
+  exit 1;
+fi
+
 retval=0
 qemu-system-x86_64 "${flags[@]}" || retval=$?
 


bazel/test_runners/qemu_with_kernel/run_qemu.sh
```diff
@@ -76,6 +76,11 @@ flags+=(-append "console=ttyS0 root=/dev/sda")
 # Disable graphics mode.
 flags+=(-nographic)
 
+if ! virt-host-validate qemu -q; then
```
JamesMBartlett, line 79: The docker container doesn't have this command. Also I think we probably want to just fix the error checking on the `qemu-system` command, such that any qemu error is returned as an exit code. The reason it doesn't return the error currently is because we transform the error code (because we assume the error code is coming from the isa exit device which is twos-complemented), so an error code of 1 from qemu becomes 0.

ddelnano: ah, I didn't consider the docker use case. I was aware that we transformed the error code, which is why I initially thought handling this before the qemu command may be more convenient.

I'll take a look into that error code transformation and see how this could be handled.

JamesMBartlett: We can also add the command to the docker container, but its unclear to me that if `qemu-system` were to fail, `virt-host-validate qemu` would be guaranteed to fail

JamesMBartlett: I think @zasgar also had a fix in the works for this

ddelnano: Sounds good. I'll wait to see if this is already addressed in any unmerged changes.

JamesMBartlett:
Closing in favor of #1093

Issue #611, opened 2022-10-04
[Self-hosted Pixie on non-minikube cluster]  Failed to authenticate. Please retry `px auth login`


amirkkn:
I've deployed pixie based on your [install documentation](https://docs.px.dev/installing-pixie/install-guides/self-hosted-pixie/) to our self-hosted kubernetes . it is about real cluster and not minikube. I'm trying to deploy pixie using CLI but i get the error "px auth login". 

````
px deploy  --dev_cloud_namespace plc --deploy_olm=false
Pixie CLI
*******************************
* ENV VARS
* 	 PL_CLOUD_ADDR=plc.e2e.example.net
*******************************
Failed to authenticate. Please retry `px auth login`.

`````

It opens the browser and i can login into. After login 
![image](https://user-images.githubusercontent.com/22954158/193857051-8f8134b4-0869-4ee9-8976-57e88350cb4e.png)

It seems everything fine, because in CLI it shows:
````
px auth login                                                                                                   ✔  10567  17:09:08
Pixie CLI
*******************************
* ENV VARS
* 	 PL_CLOUD_ADDR=plc.e2e.example.net
*******************************
Starting browser... (if browser-based login fails, try running `px auth login --manual` for headless login)
Fetching refresh token ...
Authentication Successful
````

If I deploy pixie again, i get the same error _"Failed to authenticate. Please retry `px auth login`"_. 
I checked the ~/.pixie/auth.json which the access token has been created:
````
{
  "Email": "admin@default.com",
  "IsAPIUser": false,
  "OrgID": "...",
  "Scopes": "user",
  "UserID": "...",
  "aud": "plc.e2e.example.net",
  "exp": 1672659326,
  "iat": 1664883326,
  "iss": "PL",
  "jti": "",
  "nbf": 0,
  "sub": "..."
}
````
I tries even it deployment key:
````
px deploy --deploy_key <deployment_key_generated_via_UI> --dev_cloud_namespace plc --deploy_olm=false
Pixie CLI
*******************************
* ENV VARS
* 	 PL_CLOUD_ADDR=plc.e2e.example.net
*******************************
Failed to authenticate. Please retry `px auth login`.
````
I don't find any errors in pods output. What causes this issue? Authentication seems to be OK, but deployment doesn't work. Could please provide more insights about this issue?

vihangm:
Hey @amirkkn can you please let us know what cloud release you are running? 

[...]

PR #1087, opened 2023-03-24
Fixed a ingress issue which prevented `px.cloudapi` calls

MrSupiri:
Summary: We have observed that the existing gRPC ingress is not functioning properly for any API call made for the `/px.cloudapi.*` path. This issue is due to the problem with the nginx ingress, where it sets the $HOST header of the request to upstream_balancer. This issue has been reported and documented in https://github.com/kubernetes/ingress-nginx/issues/8843.

As a result, when the traffic hits the cloud proxy service, it returns a 404 error since the request is picked up from the [default_server](https://github.com/pixie-io/pixie/blob/main/src/cloud/proxy/nginx.conf#L78) instead of [@PL_DOMAIN_NAME@](https://github.com/pixie-io/pixie/blob/main/src/cloud/proxy/nginx.conf#L101) server. To resolve this issue, we have set the ingress controller to direct the route to traffic to api-server. This has resolved the issue and restored the functionality of the gRPC ingress.

In addition, we have simplified the ingress rules by using regex since the nginx ingress controller supports regex path matching. This change has improved the readability of the ingress rules.

Relevant Issues: Could be related to https://github.com/pixie-io/pixie/issues/611, we also experience the same issue while login using `px` cli.

Type of change: /kind bug

Test Plan: Deploy the pixie self hosted cloud with this ingress, try using `px auth login --manual`.

### DIFF
--- a/k8s/cloud/overlays/exposed_services_nginx/cloud_ingress_grpcs.yaml
+++ b/k8s/cloud/overlays/exposed_services_nginx/cloud_ingress_grpcs.yaml
@@ -7,6 +7,7 @@ metadata:
   namespace: plc
   annotations:
     nginx.ingress.kubernetes.io/backend-protocol: "GRPCS"
+    nginx.ingress.kubernetes.io/use-regex: 'true'
 spec:
   ingressClassName: nginx
   tls:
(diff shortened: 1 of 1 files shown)


JamesMBartlett:

k8s/cloud/overlays/exposed_services_nginx/cloud_ingress_grpcs.yaml
```diff
@@ -112,91 +50,28 @@ spec:
   - host: work.pixie.example.com
     http:
       paths:
-      - path: /pl.cloudapi.ArtifactTracker/
+      - path: /pl.cloudapi.(.*)
         pathType: Prefix
         backend:
           service:
             name: cloud-proxy-service
```
aimichelle, line 57: Any reason you kept this as `cloud-proxy-service` rather than changing it to `api-service` like you did for `/px.cloudapi.(.*)`? 

Also, just for posterity, the `pl.cloudapi.ArtifactTracker` endpoint was kept around so that older CLIs which still excepted the endpoint to begin with `pl` would still function properly. My guess is we can actually start removing this (but we don't need to do it in this PR).

MrSupiri: We were doubting about this actually and IMO since we are using nginx ingress `cloud-proxy-service` seems. But `cloud-proxy-service` has some additional configurations related to caching and compression, so that's why we kept it around.

Will remove `pl.*` http paths.

JamesMBartlett:

k8s/cloud/overlays/exposed_services_nginx/cloud_ingress_grpcs.yaml
MrSupiri: @aimichelle shell I route all the traffic to `api-service` and ignore the `cloud-proxy-service`?

PR #1103, opened 2023-03-28
[artifact manifests] Add manifest poller to poll for manifest changes

JamesMBartlett:
Summary: Adds `manifest.Poller` to poll for new changes to a GCS manifest. The poller checks the GCS metadata to see if the MD5 hash has changed, and only downloads the manifest if the hash has changed.
When the manifest changes the poller parses the manifest from GCS and calls the callback given with the new manifest.

Type of change: /kind cleanup

Test Plan: Tested that the poller only calls the callback on new changes to the manifest.


JamesMBartlett:

aimichelle (approved):

PR #1105, opened 2023-03-28
Restrict GITHUB_TOKEN permissions for get_image workflow

vihangm:
Summary: TSIA

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: The workflow should continue to run without issues.

### DIFF
--- a/.github/workflows/get_image.yaml
+++ b/.github/workflows/get_image.yaml
@@ -13,6 +13,8 @@ on:
       image-with-tag:
         description: "A image with tag (from docker.properties) for the requested image"
         value: ${{ jobs.get-image.outputs.image-with-tag }}
+permissions:
+  contents: read
 jobs:
   get-image:
     runs-on: ubuntu-latest


.github/workflows/get_image.yaml
```diff
@@ -13,6 +13,8 @@ on:
       image-with-tag:
         description: "A image with tag (from docker.properties) for the requested image"
         value: ${{ jobs.get-image.outputs.image-with-tag }}
+permissions:
+  contents: read
```
JamesMBartlett, line 17: I thought this was the default?
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

JamesMBartlett (approved):

.github/workflows/get_image.yaml
vihangm: Depends on if the repo is configured in permissive or restrictive mode.
This ensures a lower level of access no matter the repo config.

aimichelle (approved):

PR #1109, opened 2023-03-28
[actions] Use buildbuddy only on self-hosted runners

JamesMBartlett:
Summary: We don't want to use buildbuddy from github actions directly because of potential egress/ingress costs from github actions to gcloud. Instead once #1108 lands, self-hosted runners will have access to buildbuddy. This PR replaces adding the buildbuddy api key with adding github specific bazelrc options.

Test Plan: Tested that these configs work in #1054


JamesMBartlett:

aimichelle (approved):

PR #1123, opened 2023-03-29
Implement new tls tracing method behind stirling cli flag (feature toggle)

ddelnano:
Summary: Implement new tls tracing method behind stirling cli flag (feature toggle)

The tls tracing method added in this PR determines a connection's identity (socket fd) through a different mechanism from our existing tracing. Instead of relying on struct offsets of user space data structures, it accesses the socket fd via the underlying socket syscalls while `SSL_write` / `SSL_read` calls occur. This is a prerequisite to support BoringSSL because its rolling release style makes the previous method of user space offsets untenable. This has the added benefit of reducing our maintenance cost for our existing OpenSSL tracing. Assuming future versions of OpenSSL maintain the same contract, we will not require any code changes to support them -- Pixie will gain OpenSSL v3 support once this new tracing is the default (as mentioned in the later testing).

Our assumption is that the applications that explicitly set the socket fd on the SSL struct (the applications supported by our previous tracing technique) -- [nginx](https://github.com/nginx/nginx/blob/dfe70f74a3558f05142fb552cea239add123d414/src/event/ngx_event_openssl.c#L1696), [python](https://github.com/python/cpython/blob/e375bff03736f809fbc234010c087ef9d7e0d384/Modules/_ssl.c#L836) all use Openssl with its native BIO interface. In order to verify that assumption, this change will be feature flagged and monitored carefully as its enabled on internal clusters. The existing [conn_stats_bytes metric](https://github.com/pixie-io/pixie/blob/f45ced1803e6e44406f20f1171c15a24f4d5a17a/src/stirling/source_connectors/socket_tracer/metrics.cc#L62) will be monitored for volume of tls traffic traced to verify there is no loss of instrumentation coverage between the new and old method.

Relevant Issues: #692

Type of change: /kind feature

Test Plan: Verified the following
- [x] New test passes with `enable_openssl_v3_testing` bool flag enabled ([P336](https://phab.corp.pixielabs.ai/P336)). This verifies that the new tracing technique and the feature toggle are functional since our previous tracing does not support OpenSSL v3
- [x] Existing `openssl_trace_bpf_test` and `netty_tls_trace_bpf_test` passes ([P335](https://phab.corp.pixielabs.ai/P335)). This is explicitly mentioned since `openssl_trace_bpf_test` is still disabled #699
- [x] Verified the `bool_flag` added from PR feedback conditionally enables the OpenSSL v3 testing ([P337](https://phab.corp.pixielabs.ai/P337))
- [x] Validate metrics from #1161 to verify our assumptions are correct

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/BUILD.bazel
+++ b/src/stirling/source_connectors/socket_tracer/BUILD.bazel
@@ -18,6 +18,18 @@ load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test
 
 package(default_visibility = ["//src/stirling:__subpackages__"])
 
+# TODO(ddelnano): Remove once new tls tracing implementation is
+# the default and we are ready to enable tracing of openssl v3.
+config_setting(
+    name = "enable_openssl_v3_testing",
+    values = {"define": "enable_openssl_v3_testing=1"},
+)
+
+enable_openssl_v3_tracing_defines = select({
+    ":enable_openssl_v3_testing": ["ENABLE_OPENSSL_V3_TRACING"],
+    "//conditions:default": [],
+})
+
 pl_cc_library(
     name = "cc_library",
     srcs = glob(
@@ -28,6 +40,7 @@ pl_cc_library(
         ],
     ),
     hdrs = glob(["*.h"]),
+    defines = enable_openssl_v3_tracing_defines,
     deps = [
         "//src/common/exec:cc_library",
         "//src/common/grpcutils:cc_library",
@@ -434,6 +447,7 @@ pl_cc_test(
     name = "openssl_trace_bpf_test",
     timeout = "long",
     srcs = ["openssl_trace_bpf_test.cc"],
+    defines = enable_openssl_v3_tracing_defines,
     flaky = True,
     shard_count = 2,
     tags = [
--- a/src/stirling/source_connectors/socket_tracer/BUILD.bazel
+++ b/src/stirling/source_connectors/socket_tracer/BUILD.bazel
@@ -14,8 +14,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test")
 load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test")
 
 package(default_visibility = ["//src/stirling:__subpackages__"])
 
--- a/src/stirling/source_connectors/socket_tracer/BUILD.bazel
+++ b/src/stirling/source_connectors/socket_tracer/BUILD.bazel
@@ -15,14 +15,22 @@
 # SPDX-License-Identifier: Apache-2.0
 
 load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 
 package(default_visibility = ["//src/stirling:__subpackages__"])
 
 # TODO(ddelnano): Remove once new tls tracing implementation is
 # the default and we are ready to enable tracing of openssl v3.
+bool_flag(
+    name = "enable_openssl_v3_testing_flag",
+    build_setting_default = False,
+)
+
 config_setting(
     name = "enable_openssl_v3_testing",
-    values = {"define": "enable_openssl_v3_testing=1"},
+    flag_values = {
+        ":enable_openssl_v3_testing_flag": "True",
+    },
 )
 
 enable_openssl_v3_tracing_defines = select({
--- a/src/stirling/source_connectors/socket_tracer/bcc_bpf/openssl_trace.c
+++ b/src/stirling/source_connectors/socket_tracer/bcc_bpf/openssl_trace.c
@@ -108,13 +108,13 @@ static int get_fd(uint32_t tgid, void* ssl) {
 BPF_HASH(active_ssl_read_args_map, uint64_t, struct data_args_t);
 BPF_HASH(active_ssl_write_args_map, uint64_t, struct data_args_t);
 
-static __inline void eval_and_cleanup_nested_syscall_detection(uint64_t pid_tgid) {
+static __inline int get_fd_and_eval_nested_syscall_detection(uint64_t pid_tgid) {
   struct nested_syscall_fd_t* nested_syscall_fd_ptr = ssl_user_space_call_map.lookup(&pid_tgid);
 
   if (nested_syscall_fd_ptr == NULL) {
     // This state should not occur since ssl_user_space_call_map for a given pid_tgid is set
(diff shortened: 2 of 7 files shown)


JamesMBartlett (commented):
My understanding is that there's two flags to control the new behaviour: a build time flag and a runtime flag. It seems to me like both flags need to be set for the new behaviour to be enabled. As far as I can tell, if the build-time flag is enabled but the runtime flag is not, it will attempt to trace openssl v3 traffic, but fail since the new fd tracing is not enabled (maybe?). And if only the runtime flag is enabled it won't attempt to trace openssl v3 traffic. If this is the case I think it might make sense to just have a runtime flag and not have a build time flag. Additionally, you won't be able to change the build-time flag with launchdarkly, so that side of things will be easier with just a runtime flag.

ddelnano:
> My understanding is that there's two flags to control the new behaviour: a build time flag and a runtime flag

@JamesMBartlett regarding your comment above, is the build time flag you are referencing the `enable_openssl_v3_testing` config setting? If not, can you please identify what you are referencing?

The `enable_openssl_v3_testing` setting is temporary and only intended for running tests. If `kLibSSLMatchers` was configurable at runtime, I would have opted for the test to override its entries rather than using a build setting.

This new method of tls tracing should provide the same coverage as our existing method, meaning that a PEM should trace the same amount of bytes before and after the change. However, I want to verify that assumption. In order to test that, I intend to roll out the new implementation while keeping our supported tls tracing targets constant (OpenSSL v3 must remain disabled as the feature toggle is enabled). Once the new implementation is fully rolled out and the coverage is validated against our existing method, we can remove the config setting and release OpenSSL v3 support.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
```diff
@@ -77,7 +83,7 @@ struct TraceRecords {
   std::vector<std::string> remote_address;
 };
 
-template <typename TServerContainer, bool TForceFptrs>
+template <typename TServerContainer, bool TUseNewImpl>
```
etep, line 86: We should just give the `bool` parameter a normal name. I'm assuming the `T` prefix means type in other contexts, but this is not a type parameter.

Can we use a name like `UseNestedSyscallFD`? because even if imperfect, that name describes the feature better than "NewImpl" which may become confusing in the future.

src/stirling/source_connectors/socket_tracer/BUILD.bazel
```diff
@@ -18,6 +18,18 @@ load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test
 
 package(default_visibility = ["//src/stirling:__subpackages__"])
 
+# TODO(ddelnano): Remove once new tls tracing implementation is
+# the default and we are ready to enable tracing of openssl v3.
+config_setting(
```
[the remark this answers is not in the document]
ddelnano: Addressed in [a382940](https://github.com/pixie-io/pixie/pull/1123/commits/a382940099a876629cf8a17617998f52c344814f)

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
ddelnano: > I'm assuming the T prefix means type in other contexts, but this is not a type parameter.

I wasn't sure what our convention was, but since I see `TEnableClientSideTracing` frequently I followed suit.

```
ddelnano@vigenere:~/code/pixie (ddelnano/add-feature-flag-to-opt-into-new-tls-tracing-method) $ git grep TEnableClientSideTracing
src/stirling/source_connectors/socket_tracer/testing/socket_trace_bpf_test_fixture.h:template <bool TEnableClientSideTracing = false>
src/stirling/source_connectors/socket_tracer/testing/socket_trace_bpf_test_fixture.h:    FLAGS_treat_loopback_as_in_cluster = !TEnableClientSideTracing;
src/stirling/source_connectors/socket_tracer/testing/socket_trace_bpf_test_fixture.h:    if (TEnableClientSideTracing) {
```

From searching the rest of the codebase, it doesn't appear we use that pattern elsewhere. So we should separately rename `TEnableClientSideTracing` to `EnableClientSideTracing`, which I did in https://github.com/pixie-io/pixie/pull/1125.

I initially used something similar to `UseNestedSyscallFD`, but opted for `TUseNewImpl` since this will eventually be removed once we fully roll out the new tracing method and clean up the temporary code. However, I agree that referring to what the implementation does is a better identifier so I'll update this.

ddelnano:
Moving this back to a draft since there is some additional testing the Stirling team wants to do prior to pushing forward with this.

src/stirling/source_connectors/socket_tracer/bcc_bpf/openssl_trace.c
```diff
@@ -52,6 +60,29 @@ static __inline void process_openssl_data(struct pt_regs* ctx, uint64_t id,
...
+  // happen in practice, however, the socket tracer BPF probes do trigger for generic read/write
+  // syscalls (non socket io).
+  int* expected_fd = openssl_native_bio_map.lookup(ssl);
+  if (expected_fd == NULL || *fd != *expected_fd) {
+    return kInvalidFD;
+  }
+
+  return *fd;
+}
+
```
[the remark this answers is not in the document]
ddelnano: I believe we touched on this in our meeting yesterday, and since we want to move forward with additional testing I'll hold off on responding to this until that testing is complete.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
ddelnano: This was updated in a later revision.

src/stirling/source_connectors/socket_tracer/bcc_bpf/openssl_trace.c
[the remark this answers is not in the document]
ddelnano: `SSL_set_fd` is no longer used since our additional testing from #1161 has confirmed our assumptions. While we are yet to make the final call on that data, I'm going to resolve this since the current direction no longer includes it.

ddelnano:
@etep this is ready for review again.

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
```diff
@@ -115,7 +121,7 @@ class BaseOpenSSLTraceTest : public SocketTraceBPFTestFixture</* TClientSideTrac
   }
 
   TServerContainer server_;
-  bool force_fptr_ = TForceFptrs;
+  bool use_new_tls_impl_ = UseNestedSyscallFD;
```
etep, line 124: It may be possible to make this `static constexpr` and/or you could just use the template parameter directly (i.e. there may be no need to create a "local" copy of the template parameter).

etep (approved):

src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
ddelnano: I went for using the template parameter directly. Updated in latest revision.

JamesMBartlett (approved):

PR #1125, opened 2023-03-29
Align boolean template parameter name to match our existing convention (only use T prefix for typename's)

ddelnano:
Summary: Align boolean template parameter name to match our existing convention (only use T prefix for typename's)

This originated from the following PR feedback from another change.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Searched our codebase and verified that other bool template params are not prefixed with T


etep (approved):

vihangm (approved):

ddelnano:
@vihangm my latest force push should have a gpg signed key now.

PR #1126, opened 2023-03-29
Include digests for dev images and use them where possible

vihangm:
Summary: Pinning docker images by digest is a good defensive move
to reduce security risks. Even though all these dev_images are
built by us, pinning the hash will help protect against substitution
attacks, so include the relevant digests in the docker properties
and use it as necessary.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: All the container builds and runs should continue to work.


vihangm:

JamesMBartlett (approved):

PR #1131, opened 2023-03-29
[actions] Run gazelle and go mod tidy on PRs

JamesMBartlett:
Summary: Run `make gazelle` and `make go-setup` on PRs and fail if there are any changes as a result.

Test Plan: Testing with this PR to see if a change to `go.mod` causes a check failure.

### DIFF
--- a/.github/workflows/pr_genfiles.yml
+++ b/.github/workflows/pr_genfiles.yml
@@ -45,6 +45,9 @@ jobs:
           - '**/*.proto'
           sql:
           - '**/*.sql'
+          gobuild:
+          - '**/*.BUILD'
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module px.dev/pixie
 go 1.20
 
 require (
-	cloud.google.com/go v0.81.0
+	cloud.google.com/go v0.81.1
```
vihangm, line 6: seems like an unrelated change?

vihangm (approved):

go.mod
JamesMBartlett: I'm testing it

PR #1133, opened 2023-03-30
[3 of 4] Convert perf profiler test cases to `UnitConnector`.

etep:
Summary: Convert perf profiler test cases to use `UnitConnector`. This will share code across source connector test cases.

Type of change: /kind cleanup

Test Plan: Existing tests passing.


etep:
Will re-open after converting the socket tracer tests to use unit connector.

PR #1135, opened 2023-03-30
[template_generator] Match on Kind instead of resourceType

JamesMBartlett:
Summary: There is no accurate way to get the resourceType of a k8s resource from its kind/version, without querying a k8s server. However, we only use the `resourceType` for our `TemplateMatchers` which determine which resources to apply templates to. Instead of using `resourceType` we can just as easily match on the resource's `Kind`. The only downside here is if one of the resources is a custom resource in a different group with the same `Kind` as one of the default resources. None of our current yamls hit this scenario and if we hit this in the future we can just match on `Group` as well as `Kind`.

The added benefit of this change is that we can remove the need for querying a k8s cluster in order to run `template_generator` and as a result `template_generator` goes from taking 1min to 100ms.

Test Plan: Tested that the templates generated by this PR are identical to the template generated off of main.


JamesMBartlett:

aimichelle (approved):

PR #1148, opened 2023-03-31
Disable BZip2 when building libarchive

vihangm:
Summary: Building libarchive on a system that has bzlib installed
currently fails since CMake detects it and enables it but the
sysroot doesn't include bzlib. Since we don't rely on bzip support,
lets explicitly disable it in the CMAKE config.

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: libarchive now builds successfully on my machine


JamesMBartlett (approved):

vihangm:
@pixie-io-buildbot retest this please.

PR #1155, opened 2023-04-02
Update dependabot file to track all possible deps

vihangm:
Summary: This will let dependabot auto create security
fixes but not version bumps for the various deps we have.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Wait for the next dependabot run after this lands.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,11 +5,157 @@ updates:
   directory: "/src/ui"
   schedule:
     interval: "daily"
-  # Alert but don't open PRs
+  # Disable version updates but keep security updates
   open-pull-requests-limit: 0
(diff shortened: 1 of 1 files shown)


ddelnano (approved):
How will dependabot behave if these directories are renamed? Is it smart enough to surface that issue or will it silently stop tracking the given directory?

vihangm:
> How will dependabot behave if these directories are renamed? Is it smart enough to surface that issue or will it silently stop tracking the given directory?

I'm unsure but I'd expect that the renamed stuff stops being tracked. There should be some tooling that might help keep this updated or flag changes so I'll try enabling that. Else hopefully folks `git grep` when they rename. :)

zasgar (approved):

PR #1159, opened 2023-04-04
Update and cleanup codeql action

vihangm:
Summary: The golang codeql autobuilder invokes the top level makefile without any arguments. This
means that we run `bazel clean` by default. Instead add a `make go` command that runs a simple go
compilation. Also skip the use of our devimage since we need to install golang on the running via an
action anyway.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Ensured that the codeql analyses still work.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -9,16 +9,8 @@ on:
 permissions:
   contents: read
 jobs:
-  get-dev-image:
-    uses: ./.github/workflows/get_image.yaml
-    with:
-      image-base-name: "dev_image"
   analyze-go:
     runs-on: ubuntu-latest-16-cores
-    needs: get-dev-image
-    container:
-      image: ${{ needs.get-dev-image.outputs.image-with-tag }}
-      options: --cpus 15
     permissions:
       actions: read
       contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yaml
```diff
@@ -14,6 +15,8 @@ jobs:
     with:
       image-base-name: "dev_image"
   analyze-go:
+    # go analysis is a bit slow and so skip it for PRs but run it on main
+    if: ${{ !github.event.issue.pull_request }}
```
JamesMBartlett, line 19: I think its just `github.event.pull_request`. `github.event.issue.pull_request` is for issue_comments on pull requests.

JamesMBartlett (approved):

.github/workflows/codeql.yaml
JamesMBartlett: https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

vihangm: Yeah noticed that it didn't work. Switched over to `event_name` instead.

zasgar (approved):

Issue #1132, opened 2023-03-30
The problem of flaming graph symbolization


asasas234:
**Describe the bug**
I built 1 demo monitoring system by myself and the target application for monitoring is the official px-sock-shop application.
I found that through the front-end flame chart, I couldn't locate the reason why it took 1.7s at all, because the grid up the flame chart, which shows all 0x00007f39a6723db something like this, is not converted into human readable form. For example:

<img width="1092" alt="image" src="https://user-images.githubusercontent.com/4795442/228739290-40fa43dd-e476-486b-9672-79fc64a6081e.png">

In addition, the flame chart of the Java process, I can only see the Java code execution of a very few squares, and are also the official Java classes, the application itself does not see the method-related squares, and I found that there are also similar to 0x000 such a grid, and I checked the value of the PL_PROFILER_JAVA_SYMBOLS environment variable, set to true.

**To Reproduce**
https://withpixie.ai/invite?invite_token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2ODA3NTk2NzcsInN1YiI6IjdjOTBlODNhLThiNDktNDUwNy1hMGYzLWUwZWRiMzJiODk2YSJ9.lA7lGvjEp3kxY64_5viRBVwzDkmlpgJMGG9AVxVqSlg

You can use the invitation link above to see the actual flame chart in action.

etep:
Hi @asasas234 ,

Is it possible that you are looking at a NodeJS application? We do not support symbolization for NodeJS yet. Although we plan to add NodeJS symbolization, we are not sure how to set expectations for when that feature may become available (we are not working on that feature right now).

I clicked through the link you provided and found a cluster named `xjp`. I was not able to see exactly what you saw, but I am including a fully zoomed out image of the flamegraph I found so that you can point me the region of the flamegraph to look at.

<img width="2324" alt="Screenshot 2023-03-30 at 2 18 28 PM" src="https://user-images.githubusercontent.com/809015/228967737-1fad9bfd-5125-463f-9e13-9c54b9d488bf.png">

asasas234:
@etep hi, The following diagram is an official example of 1 Java system, and you can see that there will also be symbolic problems at: https://work.withpixie.ai/live/clusters/xjp/namespaces/px-sock-shop/pods/carts-7c47dff488- kx4vg?start_time=-5m
[...]

PR #1161, opened 2023-04-04
Instrument OpenSSL tracing to detect validity of assumptions for new tls tracing method

ddelnano:
Summary: Instrument OpenSSL tracing to detect validity of assumptions for new tls tracing method

After discussing #1123 with @oazizi000 and @etep, we decided that invest in stronger validation that the assumptions for that PR and our future tls tracing are valid. Instead of relying on struct offsets of user space data structures, the new tracing method will access a connection's socket fd from the underlying socket syscalls during `SSL_write` / `SSL_read` calls.

This technique should only be compatible with "BIO native" OpenSSL use cases, which are the OpenSSL use cases Pixie supports today. BIO native means that a compatible application uses a [BIO](https://wiki.openssl.org/index.php/BIO) provided by OpenSSL (via [SSL_set_fd](https://www.openssl.org/docs/manmaster/man3/SSL_set_fd.html)) and results in OpenSSL issuing read/write syscalls on your behalf for the underlying socket.

There are two situations we wanted to understand prior to proceeding with #1123: how custom BIO implementations behave (netty, nodejs, etc) and detecting unrelated (non socket) syscalls while `SSL_write` and `SSL_read` are on the stack. The former was verified with an experiment based on this change and is described in the Test Plan section below. We believe the latter should not occur, but the changes in this PR instrument this situation in order to detect if doest occur. Assuming this condition isn't encountered, we will proceed with #1132 (with the minor changes learned from this experiment).

Relevant Issues: #692

Type of change: /kind feature

Test Plan: Verified the following:
- [x] `DCHECK` added to `socket_trace_connector` does not detect mismatched file descriptors for `openssl_trace_bpf_test` and `netty_tls_trace_bpf_test` ([P345](https://phab.corp.pixielabs.ai/P345) -- openssl_trace_bpf_test is still disabled - #699)
- [x] Verified with a BPF histogram that the number of active syscalls within SSL_write/SSL_read are non zero for BIO native use cases and zero for non BIO native cases (nodejs, netty). See more details in #1160

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
+++ b/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
@@ -162,11 +162,12 @@ static __inline void set_conn_as_ssl(uint32_t tgid, int32_t fd) {
   conn_info->ssl = true;
 }
 
-// Writes the syscall fd to a BPF map to be later consumed by the tls tracing probes in
-// addition to ensuring that set_conn_as_ssl is called. The majority of probes should call
-// this on function entry, however, there are certain probes which are not exlusively used
-// for networking and therefore must defer this fd propagation to its ret probe (where it
-// can be validated it is socket related first via sock_event).
+// Writes the syscall fd to a BPF map key created by an active tls call (the SSL_write/SSL_read
+// probes and ensures that set_conn_as_ssl is called for the tgid and pid. The majority of syscall
+// probes should call this on function entry, however, there are certain probes which are
+// not exlusively used for networking (read/write/readev/writev) and therefore must defer
+// this fd propagation to its ret probe (where it can be validated it is socket related
+// first via sock_event).
 static __inline void propagate_fd_to_user_space_call(uint64_t pid_tgid, int fd) {
   struct nested_syscall_fd_t* nested_syscall_fd_ptr = ssl_user_space_call_map.lookup(&pid_tgid);
   if (nested_syscall_fd_ptr != NULL) {
--- a/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
+++ b/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
@@ -91,6 +91,17 @@ BPF_HASH(active_write_args_map, uint64_t, struct data_args_t);
 // Key is {tgid, pid}.
 BPF_HASH(active_read_args_map, uint64_t, struct data_args_t);
 
+struct nested_syscall_fd_t {
+  int fd;
+  bool mismatched_fds;
+};
+
+// Map used to verify if SSL_write or SSL_read are on the stack during a syscall in
+// order to propagate the socket fd back to the SSL_write and SSL_read return probes.
+// Key is pid tgid.
+// Value is nested_syscall_fd struct.
+BPF_HASH(ssl_user_space_call_map, uint64_t, struct nested_syscall_fd_t);
+
 // Map from thread to its ongoing close() syscall's input argument.
 // Tracks close() call from entry -> exit.
 // Key is {tgid, pid}.
@@ -149,6 +160,29 @@ static __inline void set_conn_as_ssl(uint32_t tgid, int32_t fd) {
     return;
   }
   conn_info->ssl = true;
+}
+
+// Writes the syscall fd to a BPF map to be later consumed by the tls tracing probes in
+// addition to ensuring that set_conn_as_ssl is called. The majority of probes should call
+// this on function entry, however, there are certain probes which are not exlusively used
+// for networking and therefore must defer this fd propagation to its ret probe (where it
+// can be validated it is socket related first via sock_event).
+static __inline void propagate_fd_to_user_space_call(uint64_t pid_tgid, int fd) {
+  struct nested_syscall_fd_t*  nested_syscall_fd_ptr = ssl_user_space_call_map.lookup(&pid_tgid);
+  if (nested_syscall_fd_ptr != NULL) {
+    int expected_fd = nested_syscall_fd_ptr->fd;
+    if (expected_fd != kInvalidFD && expected_fd != fd) {
+      nested_syscall_fd_ptr->mismatched_fds = true;
+    }
+
+    nested_syscall_fd_ptr->fd = fd;
+
+    // TODO(ddelnano): Until the new tls tracing implementation is being rolled out
+    // the active user space tls function detection should not change the functionality
+    // of the existing tracing. This must be uncommented as part of pixie#1123.
+    // uint32_t tgid = pid_tgid >> 32;
+    // set_conn_as_ssl(tgid, fd);
+  }
 }
 
 static __inline struct socket_data_event_t* fill_socket_data_event(
@@ -1009,6 +1043,12 @@ int syscall__probe_ret_write(struct pt_regs* ctx) {
   // Unstash arguments, and process syscall.
   struct data_args_t* write_args = active_write_args_map.lookup(&id);
   if (write_args != NULL && write_args->sock_event) {
+    // Syscalls that aren't exclusively used for networking must be
+    // validated to be a sock_event before propagating a socket fd to the
+    // tls tracing probes
+    if (ACCESS_TLS_SK_FD_VIA_ACTIVE_SYSCALL) {
+      propagate_fd_to_user_space_call(id, write_args->fd);
+    }
     process_syscall_data(ctx, id, kEgress, write_args, bytes_count);
   }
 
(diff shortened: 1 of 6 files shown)


src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
```diff
@@ -151,6 +162,29 @@ static __inline void set_conn_as_ssl(uint32_t tgid, int32_t fd) {
...
+// this on function entry, however, there are certain probes which are not exlusively used
+// for networking and therefore must defer this fd propagation to its ret probe (where it
+// can be validated it is socket related first via sock_event).
+static __inline void propagate_fd_to_user_space_call(uint64_t pid_tgid, int fd) {
+  struct nested_syscall_fd_t* nested_syscall_fd_ptr = ssl_user_space_call_map.lookup(&pid_tgid);
+  if (nested_syscall_fd_ptr != NULL) {
+    int expected_fd = nested_syscall_fd_ptr->fd;
+    if (expected_fd != kInvalidFD && expected_fd != fd) {
+      nested_syscall_fd_ptr->mismatched_fds = true;
+    }
```
etep, line 176: Would this be easier to follow:
```
int current_fd = nested_syscall_fd_ptr->fd;
if (current_fd == kInvalidFD) {
  nested_syscall_fd_ptr->fd = fd;
}
else if (current_fd != fd) {
  // Found two different fds in nested syscalls, need to record this fact.
  nested_syscall_fd_ptr->mismatched_fds = true;
}
```
Maybe the existing functionality (setting the struct value on every invocation) is required.

Mostly, just trying to puzzle through what the "normal" path through this is.

Succinctly, if there is a way to write this that makes the "normal" path obvious, great. If not, perhaps we can add a comment that says "we expect to find the struct fd as invalid, and just set it. But if we see more than one nested syscall, we expect the same fd. If, however, we see more than one nested syscall and different fds, then we flag the issue."

ddelnano: When I was developing this I was worried that conditionally setting the file descriptor could be a source of error. It felt safer to always have the BPF map updated to the most recent fd provided to syscall.

src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
```diff
@@ -151,6 +162,29 @@ static __inline void set_conn_as_ssl(uint32_t tgid, int32_t fd) {
...
+
+    nested_syscall_fd_ptr->fd = fd;
+
+    // TODO(ddelnano): Until the new tls tracing implementation is being rolled out
+    // the active user space tls function detection should not change the functionality
+    // of the existing tracing. This must be uncommented as part of pixie#1123.
```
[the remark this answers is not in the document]
ddelnano: That's correct. Your suggestion sounds good to me so once there is clarity with the other comments I'll fix this.

src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
etep: Is there a good reason to always keep the most recent fd? I do see the intuition and wonder if we can pin down a solid reason. Gently suggesting a comment to explain the common case path through the code in any case.

ddelnano: Here's the difference I see between the two methods:
- Existing way could detect the number of times a socket fd is mismatched (although it would require more changes to truly measure that).
- Proposed way would only detect a change from whatever was initially recorded.

I'm unfortunately at a loss for which is "more correct". This validation attempts to detect a problem that we believe is not possible and as a result I don't have a strong rationale for which one to proceed with.

I think either case will catch the situation we are trying to defend against. I went ahead and updated this to match your suggestion and included a comment explaining why we must detect this.

src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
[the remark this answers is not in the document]
ddelnano: Updated in fadec6cc483118aaed0af196fcbe5a39421772fb

src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
```diff
@@ -151,6 +162,29 @@ static __inline void set_conn_as_ssl(uint32_t tgid, int32_t fd) {
   conn_info->ssl = true;
 }
 
+// Writes the syscall fd to a BPF map to be later consumed by the tls tracing probes in
+// addition to ensuring that set_conn_as_ssl is called. The majority of probes should call
+// this on function entry, however, there are certain probes which are not exlusively used
+// for networking and therefore must defer this fd propagation to its ret probe (where it
+// can be validated it is socket related first via sock_event).
```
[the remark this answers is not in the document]
ddelnano: Pete and I synced up yesterday to discuss this more. Here is my current draft based on that conversation:

```
+// Writes the syscall fd to a BPF map key created by an active tls call (the SSL_write/SSL_read probes and
+// ensures that set_conn_as_ssl is called. The majority of syscall probes should
+// call this on function entry, however, there are certain probes which are
+// not exclusively used for networking (read/write/readev/writev) and therefore must defer
+// this fd propagation to its ret probe (where it can be validated it is socket related
+// first via sock_event).
+//
+// By ensuring this is called at the appropriate time, we assume the following will be true:
+// 1. BIO native OpenSSL applications will propagate a syscall's fd one or more times for a given tls function
+//    probe depending on if read/write socket calls experience partial reads/writes. We expect that each of
+//    these syscall invocations will be for the same fd until SSL_write/SSL_read returns.
+// 2. Non BIO native cases (nodejs/netty) should never propagate a syscall fd because their SSL_write/SSL_read
+//    probes operate on in memory buffers (io is issued asynchronously). Ensuring that non socket syscalls are
+//    vetted prior to calling this function are crucial to avoid mistaking a file fd (non socket) as a socket
+//    fd in this case.
+// 3. If one or more syscalls occur within SSL_write/SSL_read, we expect the socket fd will always be the same.
+//    This assumption is a requirement for this method of tls tracing and if a fd mismatch were to occur, this
+//    function will detect the scenario and notifies stirling user space.
```

This explanation ended up longer than I expected and I'm starting to think it much be too detailed for this specific function. I think the first paragraph has important context added that was missing, but the enumerated cases in the second section feels slightly out of place to me.

@etep what are your thoughts?

etep (approved):

src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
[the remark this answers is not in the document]
ddelnano: Thanks! I ended up trimming the second paragraph. As I mentioned yesterday, I want to update the design doc with all of these findings and I think that will be an easier place to explain the second paragraph (with diagrams and links to our histogram testing).

ddelnano:
@pixie-io/maintainers can you review this and also trigger another jenkins build? The most recent build failure is unrelated (my latest commit only changed a comment and previous build passed).

ddelnano:

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -617,6 +627,30 @@ void SocketTraceConnector::UpdateTrackerTraceLevel(ConnTracker* tracker) {
...
+  // Since it doesn't initialize the BPF program, accessing the map will fail.
+  if (state() == State::kUninitialized) {
+    return;
+  }
+
+  int error_code;
+  openssl_trace_state_->get_value(kOpenSSLTraceStatusIdx, error_code);
+
+  if (error_code == kOpenSSLMismatchedFDsDetected) {
+    openssl_trace_mismatched_fds_counter_.Increment();
```
JamesMBartlett, line 643: This seems like it will undercount mismatched fds, if they occur more than once in between calls to this func. I think you said we don't expect this to ever happen so maybe its fine if we undercount, but just want to double check.

JamesMBartlett (commented):
LGTM, going to wait on the perf eval

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
ddelnano: Undercounting is fine. We just need to be able to detect if it happens despite our expectation that it should never occur (assuming our understanding of the approach taken in #1123 is correct). Appreciate the thoughtful review 👍

JamesMBartlett (approved):
The perf eval died, I think because cloud went down, but one of the 3 experiments worked: https://lookerstudio.google.com/u/0/reporting/9701de3b-f906-4dd2-a1e9-48ca0b1e07e6/page/p_g7fj6pf4yc?params=%7B%22experiment_ids%22:%22e717cae1-d763-4389-baec-3928c4eacba6%22%7D

It looks the same as what I would expect, so its probably good to go. Also, the other 2 experiments that didn't finish don't test TLS traffix so should really matter here.

PR #1168, opened 2023-04-05
Fix perf profiler bug causing higher BPF map utilization and add `DCHECK` and metrics for error conditions

ddelnano:
Summary: Fix perf profiler bug causing higher BPF map utilization and add `DCHECK` and metrics for error conditions

The perf profiler connector (and soon the socket trace connector - #1161) record error conditions in a BPF map. While the perf profiler has instrumented this value for a while, we haven't leveraged using this data in anyway. This change `DCHECK`s the error condition in addition to adding prometheus metrics.

This `DCHECK` caught a bug in our map switching logic. The perf profiler has an "A" and "B" map and alternates between them. Prior to this change we were using the A map exclusively since our variable for B was accidentally pointing to A ([source](https://github.com/pixie-io/pixie/blob/cf071f3608de5500213cef05ddb1bba88544ea66/src/stirling/source_connectors/perf_profiler/bcc_bpf/profiler.c#L64-L65)).

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Verified the following:
- [x] `DCHECK` triggers on the first commit of this PR ([P351](https://phab.corp.pixielabs.ai/P351))
- [x] Verified that the overflow prometheus metric value is incremented when the overflow condition occurs ([P350](https://phab.corp.pixielabs.ai/P350))
- [x] `DCHECK` no longer occurs with `profiler.c` fix

Changelog Message:
- Fixed a bug that may lead to missing stack frames for the continuous profiler. We believe this situation is unlikely to occur since our buffers are sized with sufficient margin.

### DIFF
--- a/src/stirling/source_connectors/perf_profiler/perf_profile_connector.cc
+++ b/src/stirling/source_connectors/perf_profiler/perf_profile_connector.cc
@@ -61,6 +61,12 @@ PerfProfileConnector::PerfProfileConnector(std::string_view source_name)
       sampling_period_(
           std::chrono::milliseconds{1000 * FLAGS_stirling_profiler_table_update_period_seconds}),
       push_period_(sampling_period_ / 2),
+      profiler_state_overflow_counter_(BuildCounter(
+          "perf_profiler_overflow",
+          "Count of times the perf profiler overran CFG_OVERRUN_THRESHOLD")),
+      profiler_state_map_read_error_counter_(BuildCounter(
+          "perf_profiler_map_read_error",
+          "Count of times the perf profiler encountered a map lookup error")),
       stats_log_interval_(std::chrono::minutes(FLAGS_stirling_profiler_log_period_minutes) /
                           sampling_period_) {
   constexpr auto kMaxSamplingPeriod = std::chrono::milliseconds{30000};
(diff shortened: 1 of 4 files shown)


src/stirling/source_connectors/perf_profiler/perf_profile_connector.cc
```diff
@@ -354,6 +360,27 @@ void PerfProfileConnector::ProcessBPFStackTraces(ConnectorContext* ctx, DataTabl
...
+  DCHECK_EQ(error_code, kPerfProfilerStatusOk);
+
+  switch (error_code) {
+      case kOverflowError:
+        profiler_state_overflow_counter_.Increment();
+        break;
+      case kMapReadFailureError:
+        profiler_state_map_read_error_counter_.Increment();
+        break;
+  }
```
ddelnano, line 376: I tried to add these prometheus metrics to a `std::array` so I could reduce the increment logic to a single line, but I ran into issues initializing the array and the prometheus library's deleted copy constructor.

etep (approved):

ddelnano:
@pixie-io/core-maintainers this is ready to be merged.

vihangm (approved):

PR #1170, opened 2023-04-05
[lint] Don't run cpplint on c files

JamesMBartlett:
Summary: cpplint makes c++ specific suggestions that shouldn't be applied to C only files. This avoids running the linter on C files.

Type of change: /kind cleanup

Test Plan: Tested that cpplint doesn't run on `.c` files.

### DIFF
--- a/.arclint
+++ b/.arclint
@@ -74,7 +74,7 @@
     "cpplint": {
       "type": "cpplint",
       "include": [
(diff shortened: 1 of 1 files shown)


vihangm (changes requested):
We should leverage the `LINT_C_FILE` directive but continue running the linter.

Issue #891, opened 2023-02-17
[ARM] Support Java Symbolization on ARM.


JamesMBartlett:
Java symbolization is currently disabled on aarch64 because the java agent is not built for aarch64.

- [ ] Add musl toolchain targeting ARM

PR #1171, opened 2023-04-05
[java/agent] Make agent compatible with C.

JamesMBartlett:
Summary: We would like to build the java agent for ARM as well as x86. We could make a fully-featured musl c++ toolchain. However, that's a lot of work and its much easier to just build musl libc with bazel and statically link it. This won't work with C++ though, because we wont have a compatible libc++. The java agent doesn't really make use of many c++ features beyond std::mutex and std::string, which are easily replaced with C versions. This PR converts the java agent to be C compatible, removing its uses of c++ features.

Relevant Issues: #891

Type of change: /kind cleanup

Test Plan: Tested that java symbols still show up in the flamegraph with this new version of the agent.


JamesMBartlett:
Lint will pass after #1170 lands

etep (approved):

zasgar (approved):

PR #1172, opened 2023-04-05
[musl build] Build musl libc with bazel.

JamesMBartlett:
Summary: Uses bazel to build musl libc. Adds `pl_cc_musl_binary` to build a `cc_binary` using musl. This macro is currently unused but will be used to build the java agent.

Test Plan: Tested that binaries built with `pl_cc_musl_binary` work.


JamesMBartlett:
This is using the GNU init artifacts, so will fix and reopen.

PR #1178, opened 2023-04-05
Upgrade openresty and switch to an alpine based image

vihangm:
Summary: The debian based openresty image has a lot of open CVEs due to all
the default packages and libs included in the base distro. Switch to an alpine
based image effectively lets us close all of these issues.
The alpine image also includes sed which is used by our entrypoint script but
is missing bash, so we switch to using sh instead. Ensured that the entrypoint
still behaves as expected.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Deployed a skaffold dev cloud. Ensured that the proxy service still
works and the UI loads and works as expected.

### DIFF
--- a/src/cloud/proxy/BUILD.bazel
+++ b/src/cloud/proxy/BUILD.bazel
@@ -66,8 +66,7 @@ container_layer(
 container_image(
     name = "proxy_server_image",
     base = "@openresty//image",
-    cmd = ["/scripts/entrypoint.sh"],
-    entrypoint = ["/bin/bash"],
+    entrypoint = ["/scripts/entrypoint.sh"],
     layers = [
         ":conf",
         ":conf_private",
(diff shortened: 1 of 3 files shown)


src/cloud/proxy/BUILD.bazel
```diff
@@ -67,7 +67,7 @@ container_image(
     name = "proxy_server_image",
     base = "@openresty//image",
     cmd = ["/scripts/entrypoint.sh"],
-    entrypoint = ["/bin/bash"],
+    entrypoint = ["/bin/sh"],
```
JamesMBartlett, line 70: why not make the script the entrypoint?

JamesMBartlett (approved):

src/cloud/proxy/BUILD.bazel
vihangm: Reasonable, will do.

PR #1180, opened 2023-04-05
Update vega-lite

NickLanam:

Summary: From 4.17 to 5.6.1. This resolves the last remaining warning in the UI build (about re-exporting defaults).
It doesn't appear to impact any of our charts at all. They are not faster, not slower, they have exactly the same bugs as before.

Type of change: /kind cleanup

Test Plan: Test various charts, such as with `px/node`. They should behave exactly the same.
The UI build should no longer emit a warning about default re-exports.


NickLanam:
Note: this PR only updates dependencies we already use, whose licenses have not changed.

PR #1185, opened 2023-04-06
Add autoscaler annotations to OLM CatalogSource object

### LINKED ISSUE #1118: Add annotations to CatalogSource objects
pixie-operator-chart deploys pixie-operator-index pod. This pod is managed by pixie-operator-index CatalogSource. 
This pod prevents GKE Cluster Autoscaler (CA) to scale-down nodes as it is not associated with any Controller. To fix this we need annotation on the pod `cluster-autoscaler.kubernetes.io/safe-to-evict: "true"`

We need this annotation on CatalogSource object. 

**Who are facing this issue?**
Organizations that are using nri-bundle helm chart to install New Relic with pixie in their cluster.

aimichelle:
Summary: #1118 reports an issue where the autoscaler doesn't autoscale down catalogsource nodes because it is missing an annotation. #1138 allows the ability to add custom annotations to the catalogsource. However, after some thought, we should just include the autoscaler annotation by default.

Relevant Issues: Fixes #1118

Type of change: /kind cleanup

Test Plan: Create operator rc, helm install the chart and verify that the catalogsource pod now has the annotation.

### DIFF
--- a/k8s/operator/helm/values.yaml
+++ b/k8s/operator/helm/values.yaml
@@ -18,7 +18,8 @@ olmBundleChannel: "stable"
 # Optional annotations and labels for CatalogSource.
 olmCatalogSource:
   # Optional custom annotations to add to deployed pods managed by CatalogSource object.
-  annotations: {}
+  annotations:
+    cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
   # Optional custom labels to add to deployed pods managed by CatalogSource object.
   labels: {}
 ## Vizier configuration


vihangm (approved):

vihangm:
This seems to also have been addressed [upstream](https://github.com/operator-framework/operator-lifecycle-manager/commit/dbd5b6f0d1dbc6981363c1345fa217629de97709). Given that we upgraded OLM recently, perhaps we don't need to add this manually after all?

aimichelle:
Nice find! I deployed the latest operator chart with the OLM upgrades, and it looks like the annotation is there now. Closing this out since it is now unneeded. :)

Issue #892, opened 2023-02-17
[ARM] Support Pixie operator and OLM on ARM


JamesMBartlett:
The olm images, and pixie operator images currently don't support ARM.
We're mainly blocked on OLM creating a new release here.

PR #1192, opened 2023-04-07
Add ARM tolerations to operator/OLM YAMLs

aimichelle:
Summary: We recently added ARM support for the operator + OLM deployed with Pixie. This PR adds the proper tolerations to these deployments. This includes the catalogsource pod that hosts the operator bundle and vizier-operator image. There remains one OLM job whose tolerations need to be updated. However, this is a change required in OLM: https://github.com/operator-framework/operator-lifecycle-manager/blob/3f598d51035702bac87368788390f801e7b21d45/pkg/controller/bundle/bundle_unpacker.go as the job spec is mostly hardcoded.

Relevant Issues: #892

Type of change: /kind cleanup

Test Plan: Deploy operator RC and `kubectl describe` pods to look at toleration.


JamesMBartlett (approved):

k8s/operator/deployment/base/arch_tolerations/deployment.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: unused
```
vihangm, line 5: do we need to include a name here? feels a bit strange to leave this in

can we move tolerations patch up a level and share across vizier and cloud?

JamesMBartlett: Kustomize requires the name. Also, kustomize requires that patch files are not in parent directories of the kustomization.yaml

aimichelle: The patch errors without the name, @JamesMBartlett  can probably give more context. They do the same thing in their example: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/patchMultipleObjects.md 

Unfortunately patches need to be in or below the directory that's getting kustomized, so we can't move this up. :(

vihangm (approved):

PR #1210, opened 2023-04-12
Add env vars used by release version stamping

vihangm:
Summary: The `save_version_info` and `get_workspace_status` scripts rely
on these env vars. Else the xdef-ed versions are incorrect and cannot
be parsed by the semver lib we use.

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Will test on a fork with a separate bucket.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -36,6 +36,8 @@ jobs:
     - name: Build & Push Artifacts
       env:
         REF: ${{ github.event.ref }}
+        BUILD_NUMBER: ${{ github.run_attempt }}
+        JOB_NAME: ${{ github.job }}
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -107,6 +111,8 @@ jobs:
     - name: Build & Push Artifacts
       env:
         REF: ${{ github.event.ref }}
+        BUILD_NUMBER: ${{ github.run_attempt }}
```
JamesMBartlett, line 114: I think you need a combination of run_number and run_attempt because I think run_attempt will always be 1 for the first attempt.

vihangm: That's fine since the ref/tag should be different right?
Attempts only rev if we rebuild the same tag.

JamesMBartlett (approved):

PR #1222, opened 2023-04-14
[actions] Add github action for CLI release.

JamesMBartlett:
Summary: Adds a github action to build, sign, and push CLI releases. I decided to split it out into a separate workflow from the other releases because it requires multiple jobs. I may also move the other release actions into their own workflows also.

Test Plan: Tested on my fork: https://github.com/JamesMBartlett/pixie/releases/tag/release%2Fcli%2Fv0.7.19

### DIFF
--- a/ci/cli_merge_sign.sh
+++ b/ci/cli_merge_sign.sh
@@ -18,17 +18,23 @@
 
 set -ex
 
+artifacts_dir="${ARTIFACTS_DIR:-?}"
+
 printenv
 
-#TODO(zasgar): Try to make it use another keychain, but it's not a big deal either way
-# since this machine is not used for anythign else.
-security unlock-keychain -p "$JENKINSKEY" login.keychain
+export KEYCHAIN_PATH=pixie.keychain
+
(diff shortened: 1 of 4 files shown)


ci/cli_merge_sign.sh
```diff
@@ -18,17 +18,26 @@
...
 
+artifacts_dir="${ARTIFACTS_DIR:-?}"
+
 printenv
 
-#TODO(zasgar): Try to make it use another keychain, but it's not a big deal either way
-# since this machine is not used for anythign else.
-security unlock-keychain -p "$JENKINSKEY" login.keychain
+export LANG=C
+export LC_ALL=C
```
vihangm, line 26: this seems like it probably isn't needed?

vihangm (approved):

ci/cli_merge_sign.sh
JamesMBartlett: Yeah this was from when I was trying to randomly generate the password. Will remove.

PR #1242, opened 2023-04-26
Fix comments in `container_runner.h`.

etep:
Summary: TSIA

Test Plan: This diff only touches comments.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/http2_trace_bpf_test.cc
+++ b/src/stirling/source_connectors/socket_tracer/http2_trace_bpf_test.cc
@@ -53,7 +53,7 @@ class HTTP2TraceTest : public testing::SocketTraceBPFTestFixture</* TClientSideT
     // Run the server.
     // The container runner will make sure it is in the ready state before unblocking.
     // Stirling will run after this unblocks, as part of SocketTraceBPFTest SetUp().
-    PX_CHECK_OK(server_.Run(std::chrono::seconds{60}));
+    PX_CHECK_OK(server_.Run(std::chrono::seconds{4*60}));
   }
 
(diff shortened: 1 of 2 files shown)


src/stirling/source_connectors/socket_tracer/http2_trace_bpf_test.cc
```diff
@@ -148,7 +148,7 @@ class ProductCatalogServiceTraceTest
     // The container runner will make sure it is in the ready state before unblocking.
     // Stirling will run after this unblocks, as part of SocketTraceBPFTest SetUp().
     // Note that this step will make an access to docker hub to download the HTTP image.
-    PX_CHECK_OK(server_.Run(std::chrono::seconds{60}));
+    PX_CHECK_OK(server_.Run(std::chrono::seconds{4 * 60}));
```
JamesMBartlett, line 151: I could be wrong but I believe this timeout is just for waiting for a ready message from the container, and it doesn't actually get killed at the end of the timeout.

JamesMBartlett, line 151: I could be wrong, but I think the timeout is only for waiting for the ready message, and it doesn't actually kill the container at the end of the timeout.

etep: I didn't actually go too deep into the code, I stopped after reading this possibly misleading comment from `container_runner.h:64`:
```
@param timeout Amount of time after which the container will be killed.
```

ddelnano (approved):

JamesMBartlett (approved):

PR #1248, opened 2023-04-27
In `SocketTraceConnector`, ensure that uprobe deployment thread is joined before dtor.

etep:
Summary: In `SocketTraceConnector`, we ensure that if a uprobe deployment thread was created, it is joined before calling the destructor. This change is motivated by observing test flakiness for `socket_trace_bpf_test` which would sometimes segfault in test case `SocketTraceServerSideBPFTest.StatsDisabledTracker`. For some reason, that test was (with some probability) able to create the exact behavior that is being fixed here.

Type of change: /kind bug fix

Test Plan: Tested locally, ran `socket_trace_bpf_test` with `--runs_per_test 8` and did not observe this specific test flake any more.


etep:
This is already handled in `SocketTraceConnector::StopImpl`.

Issue #1245, opened 2023-04-27
`dynamic_bpftrace_connector_bpf_test` shouldn't be timing dependent.


JamesMBartlett:
**Describe the Task**
The `dynamic_bpftrace_connector_bpf_test` currently sleeps for 1 second and then makes assertions. However, because QEMU executes things more slowly this fails. Instead the test should check for the records it expects in a loop, rather than relying on a sleep.
Additionally, the test makes assertions about how recently an event was received. Instead of doing this it should either not make assertions about the timestamp or do the assertion in a more robust way.

**Subtasks**
- [x] Fix sleeps
- [x] Fix timestamp assertions.

PR #1249, opened 2023-04-27
Fix a test flake in `socket_trace_bpf_test`.

etep:
Summary: In the prior impl. of test case `socket_trace_bpf_test`, there was a possible test flake based on capturing eBPF perf buffer data out of order. Because the test case started two threads, depending on which CPU owned which thread and when the perf buffers were populated, we could see the test expectations out of order (even though logically that was not what happened). We fix this test flake by removing the expectation of seeing the results in order.

Type of change: /kind bug fix

Test Plan: Tested locally, ran socket_trace_bpf_test with --runs_per_test 16 and did not observe this specific test flake any more.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/socket_trace_bpf_test.cc
+++ b/src/stirling/source_connectors/socket_tracer/socket_trace_bpf_test.cc
@@ -50,8 +50,8 @@ using ::px::system::TCPSocket;
 using ::px::system::UDPSocket;
 using ::px::system::UnixSocket;
 using ::px::types::ColumnWrapperRecordBatch;
-using ::testing::HasSubstr;
 using ::testing::Contains;
+using ::testing::HasSubstr;
 using ::testing::StrEq;
 
 constexpr std::string_view kHTTPReqMsg1 =
@@ -589,10 +589,9 @@ TEST_F(SocketTraceBPFTest, SendFile) {
 
   ASSERT_THAT(records, RecordBatchSizeIs(2));
 
-  const absl::flat_hash_set<std::string> responses {
-    records[kHTTPRespBodyIdx]->Get<types::StringValue>(0),
-    records[kHTTPRespBodyIdx]->Get<types::StringValue>(1)
-  };
+  const absl::flat_hash_set<std::string> responses{
+      records[kHTTPRespBodyIdx]->Get<types::StringValue>(0),
+      records[kHTTPRespBodyIdx]->Get<types::StringValue>(1)};
 
   const std::string kHTTPRespMsgContentAsFiller(kHTTPRespMsgContent.size(), 0);
   EXPECT_THAT(responses, Contains(kHTTPRespMsgContentAsFiller));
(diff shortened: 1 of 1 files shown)


src/stirling/source_connectors/socket_tracer/socket_trace_bpf_test.cc
```diff
@@ -569,6 +569,12 @@ TEST_F(SocketTraceBPFTest, SendFile) {
     std::string data;
 
     client.Send(kHTTPReqMsg1);
+
+    // This test case assumes that eBPF will pick up the client *after* the server.
+    // With some low probability, without sleeping here, that is not always true.
+    // Sleeping gives enough time for eBPF perf buffers to be populated in the correct sequence.
+    sleep(1);
```
ddelnano, line 576: Can you explain how this test flakiness is manifesting? I'm not understanding the "pick up the client after the server" part and why that indicates there isn't enough time before the assertions.

Since the qemu test suite seems sensitive to timing issues (#1245), it seems like we should verify that this doesn't make qemu's situation worse (assuming this timing delay is needed).

JamesMBartlett (changes requested):
QEMU is slower to execute than native code. As we try to transition to QEMU, I've been trying to reduce dependence on timing. I think we should avoid adding more timing dependence.

etep:
@ddelnano @JamesMBartlett 

The problem:
The client and server responses can be observed out-of-sequence. Looking specifically at the following two lines:
```
std::string server_body = records[kHTTPRespBodyIdx]->Get<types::StringValue>(0);
std::string client_body = records[kHTTPRespBodyIdx]->Get<types::StringValue>(1);
```
The `server_body` is sometimes found at index 1 and the `client_body` is (in those cases) found at index 0. Running on our `enigma` dev. host, with `--runs_per_test 16`, I found this in 4 out of 16 test runs.

There are several ways to fix this, here are three that I have tested:
1. [sleep](https://github.com/etep/pixie/commit/5151f470f729ffcbcabc33b0651469a292233bda)
2. [barriers ](https://github.com/etep/pixie/commit/d9645616d0df1209988f82bae558a735f0d64208), as suggested [1] by @JamesMBartlett
3. [contains](https://github.com/etep/pixie/commit/20437e82f23fbee10bf3e5698c440e0c021001e2)

We could also create a sub-process so that we can disambiguate the server/client responses by pid. There may be another way to fix not contemplated here.

All the fixes (linked above) pass 16 out of 16.

Here are my comments for each:
`sleep`: the marginal cost here is not high. Is QEMU actually sensitive to a 1 second sleep?
`barriers`: less run time than `sleep`.
`contains`: simple, small loss of test coverage. Fastest run time.

Run times [2]:
```
SLEEP:      INFO: Elapsed time: 340.190s, Critical Path: 275.20s
BARRIERS:   INFO: Elapsed time: 320.343s, Critical Path: 318.38s
CONTAINS:   INFO: Elapsed time: 306.059s, Critical Path: 304.10s
```

[1] @JamesMBartlett maybe there is a better way to materialize your suggestion, not sure I got this version to its best possible state.
[2] Ran with all tests disabled (*except* for send file).

JamesMBartlett:
I'm fine with 2 or 3

ddelnano:
I'm also fine with 2 or 3 with a slight preference to 3 since I don't think we need strict ordering for this use case.

src/stirling/source_connectors/socket_tracer/socket_trace_bpf_test.cc
etep: As discussed in comment trail.

ddelnano (approved):

JamesMBartlett (approved):

vihangm:
@pixie-io-buildbot retest this please.

vihangm:

PR #1256, opened 2023-05-01
Add `format_duration` pxl function to aid duration calculations

ddelnano:
Summary: Add `format_duration` pxl function to aid duration calculations

While developing the `http_trace_id` pxl script in #1233, I had to perform some brittle math in order to parameterize the deep linked script ([source](https://github.com/pixie-io/pixie/blob/e6f7ad60905bf6a0b696cb74d2b91230017c764e/src/pxl_scripts/px/http_trace_id/script.pxl#L103-L111)). This function will simplify that logic and ensure that it works for all units (currently it would break if anything other than minutes were used).

Relevant Issues: N/A

Type of change: /kind new-pxl-script

Test Plan: Tests added for `format_duration` script pass

Changelog Message: Added the `format_duration` pxl function, which simplifies converting a duration integer to a string representation. This is useful for manipulating script time windows within pxl scripts (when deep linking pxl scripts via `px.script_reference`).

### DIFF
--- a/src/carnot/planner/objects/pixie_module.h
+++ b/src/carnot/planner/objects/pixie_module.h
@@ -200,6 +200,29 @@ class PixieModule : public QLObject {
   Returns:
     px.Time: The time value represented in the data.
   )doc";
+
+  inline static constexpr char kFormatDurationOpID[] = "format_duration";
+  inline static constexpr char kFormatDurationDocstring[] = R"doc(
+  Convert an integer in nanoseconds (-3,000,000,000) to a duration string ("-5m") while
+  preserving the sign. This function converts to whole milliseconds, seconds, minutes,
+  hours or days. This means it will round down to the nearest whole number time unit.
+
+  Examples:
+    # duration = "-5m"
+    duration = px.format_duration(-5 * 60 * 1000 * 1000 * 1000)
+    # duration = "5m" duration is rounded down to nearest whole minute.
+    duration = px.parse_duration(-5 * 60 * 1000 * 1000 * 1000 + 5)
+
+    # duration = "-5h"
+    duration = px.format_duration(-5 * 60 * 60 * 1000 * 1000 * 1000)
+
+  :topic: compile_time_fn
+
+  Args:
+    duration (int64): The duration in nanoseconds.
+  Returns:
+    string: The string representing the human readable duration (i.e. -5m, -10h).
+  )doc";
   inline static constexpr char kParseDurationOpID[] = "parse_duration";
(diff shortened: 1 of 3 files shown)


src/carnot/planner/objects/pixie_module.h
```diff
@@ -200,6 +200,29 @@ class PixieModule : public QLObject {
   Returns:
     px.Time: The time value represented in the data.
   )doc";
+
+  inline static constexpr char kFormatDurationOpID[] = "format_duration";
+  inline static constexpr char kFormatDurationDocstring[] = R"doc(
+  Convert an integer in nanoseconds (-3,000,000,000) to a duration string ("-5m") while
+  preserving the sign. This function converts to whole milliseconds, seconds, minutes,
+  hours or days. This means it will round down to the nearest whole number time unit.
```
JamesMBartlett, line 208: I'm not sure I agree with this behaviour. eg. I don't think `px.format_duration(int representing 5h59m) == '5h'` makes much sense

ddelnano: I should have explained this in the summary of the PR, but I chose to implement it this way because our UI only seems to supports a duration string with a single time unit.

You can see from the screenshots below that specifying `-1d12h` and `-36h` yield different results. The former seems to return results from the last 24 hours, which seems to indicate that the `12h` portion is ignored.

Screen Shot 2023-05-01 at 5 23 00 PM
Screen Shot 2023-05-01 at 5 23 13 PM

In addition to that, the parse_duration function will only parse strings that include a [single](https://github.com/pixie-io/pixie/blob/0072ce95cde9d6a8788163746a41e833482136be/src/carnot/planner/objects/pixie_module.cc#L371) time [unit](https://github.com/pixie-io/pixie/blob/0072ce95cde9d6a8788163746a41e833482136be/src/common/base/time.h#L42).

I essentially was aiming for compatibility with this change. Since the function's use case would be to craft other pxl script time windows, I wanted it to support the inputs our pxl scripts and related functions (`parse_duration`) expect.

I'm happy to change this but it feels like there is a potential to cause confusion. If we want to pursue that, I would want to evaluate how much work it would be to support that type of a duration string in the other aspects of Pixie (UI and other pxl script functions).

JamesMBartlett: Makes sense, I didn't realize that `parse_duration` doesn't handle this case either. We can maybe consider fixing this across the board later. But for now I think mirroring what `parse_duration` does is fine.

JamesMBartlett (approved):

PR #1257, opened 2023-05-01
Release changelogs should read release-notes from commit messages

aimichelle:
Summary: The current process for writing changelogs is manual. You manually compile the changelogs and pipe it to the `create_release_tag` script to be used as the tag's message.
We've now standardized our commit message formats, so we can parse commits to determine the type of commit + if it has any release notes to include. This PR updates `create_release_tag` so that it:
1. Finds all commits which touch a dependency of the artifact being released.
2. Parses the commit message to find the kind of commit (bug, feature request, cleanup).
3. Parses the commit message to get the release notes, if any.
4. Takes the commits and compiles it into a markdown-formatted changelog.

Here's an example changelog (there were no changelog updates since the last release so I set it to gather changelogs from a much earlier prev version):
```
tag release/vizier/v0.13.1-pre-changelog5.1
Tagger: Michelle Nguyen <michellenguyen@pixielabs.ai>

##Changes by Kind

###New Features
- New chart type: gauge
- New chart type: pie
- New chart type: Text. Literally just the text you type in the Vis spec, nothing more.
- New chart type: Stat. Shows a label, metric, and units for that metric.

###Bug Fixes
- Fixes a bug where postgres requests could be dropped causing missing protocol traces
update changelog generation

```

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Ran `create_release_tag`. See example output above.

### DIFF
--- a/scripts/create_release_tag.sh
+++ b/scripts/create_release_tag.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -ex
+#!/bin/bash
 
 # Copyright 2018- The Pixie Authors.
 #
@@ -19,7 +19,7 @@
 # This creates a new tag in git for the current commit.
 
 usage() {
-    echo "Usage: echo <changelog message> | $0 <artifact_type> [-p] [-r] [-m] [-n]"
+    echo "Usage: $0 <artifact_type> [-p] [-r] [-m] [-n]"
     echo " -p : Push the tag to Github."
     echo " -r : Create a release."
     echo " -m : Increment the major version."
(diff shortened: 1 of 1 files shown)


scripts/create_release_tag.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -ex
+#!/bin/bash
```
vihangm, line 1: We probably want the `-e` to stay so that the script fails on any intermediate errors.

scripts/create_release_tag.sh
```diff
@@ -112,16 +112,93 @@ function update_pre {
...
+                typeRe='Type of change: /kind ([A-Za-z]+)'
+                if [[ $log =~ $typeRe ]]
+                then
+                changeType=${BASH_REMATCH[1]}
+                fi
+
+                # Get release notes.
+                # shellcheck disable=SC2016
+                # The backticks make shellcheck think that the string should be evaluatable.
+                notesRe='```release-note\s((\S|\s)*)```'
```
vihangm, line 152: can't we use double quotes instead and not disable the check?

scripts/create_release_tag.sh
```diff
@@ -172,9 +249,12 @@ if [ "$RELEASE" != "true" ]; then
...
+changelog=$(generate_changelog "$prev_tag" "$BAZEL_TARGET")
+
 new_tag="release/$ARTIFACT_TYPE/v"$new_version_str
-git tag -a "$new_tag" -m "$(echo -e "$CHANGELOG")"
+echo -e "$changelog"
+git tag -a "$new_tag" -m "$changelog" --cleanup=whitespace
 
 if [ "$PUSH" = "true" ]; then
-  git push origin "$new_tag"
+   git push origin "$new_tag"
```
vihangm, line 259: Indentation seems off?

scripts/create_release_tag.sh
```diff
@@ -112,16 +112,93 @@ function update_pre {
...
+                releaseNote=''
+
+                log=$(git log --format=%B -n 1 "$commit")
+
+                # Get the type of change (cleanup|bug|feature).
+                typeRe='Type of change: /kind ([A-Za-z]+)'
+                if [[ $log =~ $typeRe ]]
+                then
+                changeType=${BASH_REMATCH[1]}
+                fi
```
vihangm, line 147: can we format these as the following to improve scanability
```
if [[condition]]; then
  expression
fi
```

scripts/create_release_tag.sh
```diff
@@ -112,16 +112,93 @@ function update_pre {
...
+
+                if [[ -n $releaseNote ]]
+                then
+                case $changeType in
+                "cleanup")
+                # shellcheck disable=SC2086
+                # Double-quoting adds an extra newline.
+                cleanup_changelog="${cleanup_changelog}- $(echo -e $releaseNote)\n"
+                ;;
```
vihangm, line 165: nit: re formatting to improve scanability
```
case EXPR in
  PATTERN1)
    STATEMENTS
  PATTERN2)
    STATEMENTS
esac
```

scripts/create_release_tag.sh
```diff
@@ -112,16 +112,93 @@ function update_pre {
...
+                "bug")
+                # shellcheck disable=SC2086
+                bug_changelog="${bug_changelog}- $(echo -e $releaseNote)\n"
+                ;;
+                "feature")
+                # shellcheck disable=SC2086
+                feature_changelog="${feature_changelog}- $(echo -e $releaseNote)\n"
+                ;;
```
vihangm, line 173: IMO this is a bit hard to grok. maybe we should just use arrays and then combine them later?

`feature_changelog=()` to initialize
`feature_changelog+=($releaseNote)` to append
`printf '- %s\n' "${feature_changelog[@]}"` to format+join.

vihangm (changes requested):

scripts/create_release_tag.sh
```diff
@@ -112,16 +112,93 @@ function update_pre {
...
+                esac
+                fi
+                break
+            fi
+        done
+    done
+
+    # Format changelog.
+    changelog="## Changes by Kind\n"
```
JamesMBartlett, line 184: imo I think the `Changes by Kind` header is unnecesary

scripts/create_release_tag.sh
aimichelle: I initially tried to do that, but iterating through the array to format the final output got a little tricky. You can set the IFS to `\n` but you can have releaseNotes that are multiline. I guess I could add some special char at the end of each release note and use that as the separator?

scripts/create_release_tag.sh
```diff
@@ -112,16 +112,91 @@ function update_pre {
...
+
+    bug_changelog=''
+    feature_changelog=''
+    cleanup_changelog=''
+
+    # Find all the commits between now and the last release.
+    commits=$(git log HEAD..."$prev_tag" --pretty=format:"%H")
+
+    # Find all file dependencies of the bazel target.
+    bazel query --noshow_progress 'kind("source file", deps('"$bazel_target"')) union buildfiles(deps('"$bazel_target"'))' | sed  -e 's/:/\//' -e 's/^\/\+//' > target_files.txt
```
vihangm, line 127: IMO this expression is brittle and doesn't handle many things such as third party repos etc, might be easier to map from changed files to affected targets like we do in `ci/bazel_build_deps.sh`
Or perhaps easier to just create a set of folders that should be watched based on release kind.

This is a ok starting point but we should revisit this in the future.

vihangm (approved):

aimichelle:
Tested out the latest changes. Looks good, so merging it in. We can tackle the build deps afterwards.

PR #1258, opened 2023-05-01
Tidy up & fix `EverythingLocalContext`.

etep:
Summary: In the previous impl. of `EverythingLocalContext`, we masked a base class instance of `upids_`. We fix that in this patch, and align usage of the `EverythingLocalContext` to better match the existing work in `socket_trace_bpf_test_fixture.h`.

Type of change: /kind cleanup

Test Plan: Tested locally with the next diff and along with work to de-flake socket tracer tests.


etep:
This patch, #1258, is in support of the following:
[stacked change in `unit_connector.h`](https://github.com/etep/pixie/compare/ddf-fix-everything-local-context...etep:ddg-unit-connector-fixes).
As soon as we can land this, the next (above) will pushed up as a PR.

JamesMBartlett (approved):

PR #1260, opened 2023-05-01
Add `pprof` as an external repo.

etep:
Summary: We add Google `pprof` as an external repository to support upcoming export of perf profiling in `pprof` format.

Type of change: /kind feature

Test Plan: Tested locally in a separate branch. At the moment, this additional bazel repo is not used.

### DIFF
--- a/bazel/external/pprof.BUILD
+++ b/bazel/external/pprof.BUILD
@@ -1,0 +1,31 @@
+# Copyright 2018- The Pixie Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


vihangm (changes requested):
This dep is already pulled in by our `go.mod` and the `proto_library` rule is available as `@com_github_google_pprof//proto:perftools_profiles_proto`

You should be able to create a `pl_cc_proto_library` with that rule and use it.

vihangm:
You can also directly access the `.proto` file as `@com_github_google_pprof//proto:profile.proto` incase you don't want to rely on the auto generated `proto_library` rule.

etep:
We won't need this diff.

PR #1262, opened 2023-05-02
[perf_tool] Specify experiments in terms of actions.

JamesMBartlett:
Summary: Instead of specifying experiment duration and pre-workload duration, explicitly specify the actions to run for the experiment and the order to run them in. Each action can optionally have a duration associated with it, for example, the `RUN` and `BURNIN` actions will sleep for the specified duration to allow the experiment to run/burnin. This allows for much more complicated experiments to be specified. For example, we can now have an experiment where for the first half we run workloads without vizier and the second half we run vizier, in order to measure overhead on the application caused by vizier.

Test Plan: Ran a full perf experiment locally, saw that it worked without issue.


src/e2e_test/perf_tool/pkg/run/run.go
```diff
@@ -279,13 +359,15 @@ func (r *Runner) prepareWorkloads(ctx context.Context, spec *experimentpb.Experi
 		if err := w.Prepare(); err != nil {
 			return err
 		}
-		workloads[i] = w
+		if r.workloadsBySelector[s.ActionSelector] == nil {
+			r.workloadsBySelector[s.ActionSelector] = make([]deploy.Workload, 0)
+		}
```
vihangm, line 364: this make is unnecessary, `append` works with nil slices.

src/e2e_test/perf_tool/pkg/run/run.go
```diff
@@ -104,139 +107,216 @@ func (r *Runner) RunExperiment(ctx context.Context, expID uuid.UUID, spec *exper
...
+			w.Close()
+		}
+	}()
+	for _, w := range ws {
+		// We return the first error, and the defer will cleanup the rest.
 		if err := w.Close(); err != nil {
 			return err
 		}
 	}
+	return nil
```
vihangm, line 303: cleaner to use `errors.Join` to just grab all errors. you can stick them in a slice and then call join.
https://pkg.go.dev/errors#Join

src/e2e_test/perf_tool/pkg/run/run.go
```diff
@@ -104,139 +107,216 @@ func (r *Runner) RunExperiment(ctx context.Context, expID uuid.UUID, spec *exper
...
+			mr.Close()
+		}
+	}()
+	for _, mr := range mrs {
+		// We return the first error, and the defer will cleanup the rest.
+		if err := mr.Close(); err != nil {
 			return err
 		}
 	}
+	return nil
```
vihangm, line 283: see note below re: `errors.Join`

vihangm (approved):

PR #1264, opened 2023-05-02
[perf_tool] Exit early on errors in metric recorders

JamesMBartlett:
Summary: Errors often occur in the metric recorders when a cluster goes unhealthy or a query fails for other reasons. Previously, these errors would only be surfaced after the experiment finished running. This PR adds a channel to send the errors to, so that the errors are caught as soon as they happen and the experiment can be abandoned earlier.

Test Plan: Tested that sending an error to the error channel leads to the experiment being aborted early.

### DIFF
--- a/src/e2e_test/perf_tool/pkg/run/run.go
+++ b/src/e2e_test/perf_tool/pkg/run/run.go
@@ -55,10 +55,14 @@ type Runner struct {
 	clusterCtx          *cluster.Context
 	clusterCleanup      func()
 	vizier              deploy.Workload
-	wg                  sync.WaitGroup
 	metricsResultCh     chan *metrics.ResultRow
 	metricsBySelector   map[string][]metrics.Recorder
 	workloadsBySelector map[string][]deploy.Workload
+
+	// wg is for goroutines that are unrelated to the main execution of the experiment.
+	wg sync.WaitGroup
+	// eg is for goroutines that should fail the experiment if they return an error.
+	eg *errgroup.Group
 }
 
 // NewRunner creates a new Runner for the given contexts.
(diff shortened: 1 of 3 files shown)


src/e2e_test/perf_tool/pkg/run/run.go
```diff
@@ -108,6 +112,43 @@ func (r *Runner) RunExperiment(ctx context.Context, expID uuid.UUID, spec *exper
 	defer r.clusterCleanup()
 	defer r.clusterCtx.Close()
 
+	var egCtx context.Context
+	r.eg, egCtx = errgroup.WithContext(ctx)
+	r.eg.Go(func() error {
+		return r.runActions(egCtx, spec)
+	})
```
vihangm, line 119: maybe write a small note here saying that you are relying on errgroup cancellation of context on first error.

src/e2e_test/perf_tool/pkg/run/run.go
```diff
@@ -55,10 +55,14 @@ type Runner struct {
 	clusterCtx          *cluster.Context
 	clusterCleanup      func()
 	vizier              deploy.Workload
-	wg                  sync.WaitGroup
 	metricsResultCh     chan *metrics.ResultRow
 	metricsBySelector   map[string][]metrics.Recorder
 	workloadsBySelector map[string][]deploy.Workload
+
+	// wg is for goroutines that are unrelated to the main execution of the experiment.
+	wg sync.WaitGroup
```
vihangm, line 63: don't they also include the experiment? seems like wg also includes stuff in eg.

vihangm (approved):

src/e2e_test/perf_tool/pkg/run/run.go
JamesMBartlett: The only thing the wait group is used for at the moment is the bigquery inserter which runs in the background and moves results from the result channel to bigquery.

PR #1270, opened 2023-05-02
Add dimensional metrics to record applications that violate tls tracing assumptions (mismatched fds)

ddelnano:
Summary: Add dimensional metrics to record applications that violate tls tracing assumptions (mismatched fds)

Relevant Issues: #692

Type of change: /kind feature

Test Plan: Verified that the metrics added do not clash with the non dimensional metrics. Please see https://github.com/pixie-io/pixie/commit/e83443a7ad555abe13d5d9bd0d28fc9084ffc427 for the code that was used to produce this output ([P363](https://phab.corp.pixielabs.ai/P363))

### DIFF
--- a/src/common/metrics/metrics.h
+++ b/src/common/metrics/metrics.h
@@ -30,6 +30,13 @@ prometheus::Registry& GetMetricsRegistry();
 // This function should only be called by testing code.
 void TestOnlyResetMetricsRegistry();
 
+// A convenience wrapper to return a counter with the specified name and help message when
+// dimensions aren't known at compile time. This should only be used when dimensional data is
+// very low cardinality.
+inline auto& BuildRawCounter(const std::string& name, const std::string& help_message) {
+  return prometheus::BuildCounter().Name(name).Help(help_message).Register(GetMetricsRegistry());
+}
+
 // A convenience wrapper to return a counter with the specified name and help message.
 inline auto& BuildCounter(const std::string& name, const std::string& help_message) {
   return prometheus::BuildCounter()
--- a/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
+++ b/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
@@ -17,7 +17,6 @@
  */
 
 #include "src/stirling/source_connectors/socket_tracer/socket_trace_connector.h"
-#include <prometheus/text_serializer.h>
 
 #include <sys/sysinfo.h>
 #include <sys/types.h>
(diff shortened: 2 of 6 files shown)


etep (approved):

src/common/metrics/metrics.h
```diff
@@ -30,6 +30,13 @@ prometheus::Registry& GetMetricsRegistry();
 // This function should only be called by testing code.
 void TestOnlyResetMetricsRegistry();
 
+// A convenience wrapper to return a counter with the specified name and help message when
+// dimensions aren't known at compile time. This should only be used when dimensional data is
+// very low cardinality.
+inline auto& BuildRawCounter(const std::string& name, const std::string& help_message) {
```
JamesMBartlett, line 36: nit: let's use the prometheus terminology of families, so maybe call it `BuildCounterFamily`

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -170,13 +170,16 @@ constexpr size_t kMaxHTTPHeadersBytes = 8192;
...
+
 SocketTraceConnector::SocketTraceConnector(std::string_view source_name)
     : SourceConnector(source_name, kTables),
       conn_stats_(&conn_trackers_mgr_),
-      openssl_trace_mismatched_fds_counter_(BuildCounter(
-          "openssl_trace_mismatched_fds",
-          "Count of the times a syscall's fd was mismatched when detecting fds from an "
-          "active user space call")),
+      openssl_trace_mismatched_fds_counter_(
+          BuildCounter(openssl_mismatched_fds_metric, openssl_mismatched_fds_help)),
```
JamesMBartlett, line 182: I think we should call `BuildCounterFamily` here, and then call `...family_.Add(...).Increment()` on the class member below

src/common/metrics/metrics.h
ddelnano: Sounds good and updated.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
ddelnano: Updated.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -648,7 +654,18 @@ void SocketTraceConnector::CheckTracerState() {
   openssl_trace_state_->get_value(kOpenSSLTraceStatusIdx, error_code);
 
   if (error_code == kOpenSSLMismatchedFDsDetected) {
-    openssl_trace_mismatched_fds_counter_.Increment();
+    openssl_trace_mismatched_fds_counter_family_.Add({{"name", openssl_mismatched_fds_metric}})
```
JamesMBartlett, line 657: do we still need this metric given its just the sum of all the different `exe` ones below.

JamesMBartlett (approved):

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
ddelnano: I agree that it is likely redundant, but out of caution I wanted to keep the existing metric. Any instance of this should be recorded and so I wanted to rule out that the additional loop and BPF hash contents may impact the signal we are already getting.

PR #1266, opened 2023-05-02
Upgrade golang

vihangm:
Summary: This upgrades our golang SDKs to the latest patch version
to pull in bug fixes and security fixes.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Existing tests should provide enough coverage.

PR #1272, opened 2023-05-02
[build] Allow building non-PIE go executables without explicitly setting no-pie

JamesMBartlett:
Summary: The golang linker command accepts `-extld` to specify what external linker to use. If this option is a relative path, things mostly work. However, the [`linkerFlagSupported`](https://github.com/golang/go/blob/0d347544cbca0f42b160424f6bc2458ebcc7b3fc/src/cmd/link/internal/ld/lib.go#L1994) function doesn't work properly with the relative path, so some linker flags are not set. `rules_go` has a wrapper go binary that eventually calls out to `go tool link`. We circumvent the issue by making sure this wrapper binary uses an absolute path for `-extld` instead of a relative one.

Type of change: /kind cleanup

Test Plan: Tested that `//:pl_go_base_image` still builds properly with the new go upgrade (#1266).

### DIFF
--- a/bazel/cc_toolchains/clang/toolchain.BUILD
+++ b/bazel/cc_toolchains/clang/toolchain.BUILD
@@ -88,7 +88,7 @@ cc_toolchain_config(
         "-Wl,-z,relro,-z,now",
         "-Bexternal/{this_repo}/{toolchain_path}/bin",
         "-lm",
-    ] + (["-no-pie"] if {use_for_host_tools} else []),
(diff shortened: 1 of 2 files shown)


vihangm (approved):

JamesMBartlett:
This doesn't work with sysroots

JamesMBartlett:
Closing this for now since there's some issues with getting this to work with sysroots

PR #1281, opened 2023-05-03
[bpf tests] Remove timing dependence from dynamic_bpftrace_connector_bpf_test

### LINKED ISSUE #1245: `dynamic_bpftrace_connector_bpf_test` shouldn't be timing dependent.

JamesMBartlett:
Summary: Removes dependence on sleeps/timing from the dynamic_bpftrace bpf test, by adding loops with timeouts to wait for records to show up.

Relevant Issues: Fixes #1245.

Type of change: /kind cleanup

Test Plan: Along with #1278, ran `bazel test --config=qemu-bpf //src/stirling/source_connectors/dynamic_bpftrace:dynamic_bpftrace_connector_bpf_test --runs_per_test=50` without any failures.

### DIFF
--- a/src/stirling/source_connectors/dynamic_bpftrace/dynamic_bpftrace_connector_bpf_test.cc
+++ b/src/stirling/source_connectors/dynamic_bpftrace/dynamic_bpftrace_connector_bpf_test.cc
@@ -39,6 +39,7 @@ using ::px::stirling::dynamic_tracing::ir::logical::TracepointDeployment_Tracepo
 using ::px::stirling::dynamic_tracing::ir::shared::DeploymentSpec;
 using ::px::stirling::testing::FindRecordsMatchingPID;
 using ::px::stirling::testing::RecordBatchSizeIs;
+using ::px::stirling::testing::Timeout;
 using ::px::testing::status::StatusIs;
 using ::testing::Gt;
 using ::testing::HasSubstr;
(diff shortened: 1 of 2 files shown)


etep (approved):

src/stirling/source_connectors/dynamic_bpftrace/dynamic_bpftrace_connector_bpf_test.cc
```diff
@@ -212,19 +219,19 @@ TEST(DynamicBPFTraceConnectorTest, BPFTraceBuiltins) {
...
     EXPECT_LE(nsecs, now_ns);
 
     int64_t elapsed = records[kElapsedIdx]->Get<types::Int64Value>(0).val;
     LOG(INFO) << absl::Substitute("elapsed: $0", elapsed);
-    // Expecting a value of close to 100 ms. Add +/- 50 ms to be generous.
+    // Expecting a value of close to 100 ms. Look for between 50 and 200 to be generous.
     constexpr int kNanosPerMilli = 1000 * 1000;
     EXPECT_GE(elapsed, 50 * kNanosPerMilli);
-    EXPECT_LE(elapsed, 150 * kNanosPerMilli);
+    EXPECT_LE(elapsed, 200 * kNanosPerMilli);
```
vihangm, line 234: This seems liable to cause flakes in the future, do we really need to have an expectation for the elapsed time?

vihangm (approved):

ddelnano (approved):

PR #1283, opened 2023-05-04
Update `perf_profiler_bpf_test` to use `UnitConnector`.

etep:
Summary: we convert `perf_profiler_bpf_test` to use `UnitConnector` to share code across different tests and binaries.

We also refactor the Java test case to separately test the expectation that Java symbolization artifacts are removed after processes stop. The main Java test case then conforms to the pattern established for Go and cpp.

We also update some of the helper functions in the test to tidy up the common test cases.

Type of change: /kind cleanup

Test Plan: existing tests & runs per test.
- [x] `bazel test //src/stirling/source_connectors/perf_profiler/... --config bpf --runs_per_test 16` ([all tests passed](https://phab.corp.pixielabs.ai/P372)).

### DIFF
--- a/src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
+++ b/src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
@@ -231,7 +231,7 @@ class PerfProfileBPFTest : public ::testing::TestWithParam<std::filesystem::path
       test_run_time_ = std::chrono::seconds{FLAGS_stirling_profiler_table_update_period_seconds};
     }
   }
-  
+
   void PopulateObservedStackTraces(const std::vector<size_t>& target_row_idxs) {
     auto stack_traces_column = columns_[kStackTraceStackTraceStrIdx];
     auto counts_column = columns_[kStackTraceCountIdx];
@@ -505,7 +505,7 @@ TEST_F(FastPerfProfileBPFTest, PerfProfilerJavaArtifactsCleanupTest) {
   std::set<int> pids_with_java_symbols;
   {
     Timeout timer;
-    while( !timer.TimedOut() && pids_with_java_symbols.size() < kNumSubProcs) {
+    while (!timer.TimedOut() && pids_with_java_symbols.size() < kNumSubProcs) {
       ASSERT_OK(source_.Flush());
       auto s = source_.ConsumeRecords(kProfilerTableNum);
       if (s.ok()) {
@@ -552,8 +552,8 @@ TEST_F(FastPerfProfileBPFTest, PerfProfilerJavaArtifactsCleanupTest) {
 
   // Need to pass in an invalid upid set to prevent picking up the default context that
   // accepts all pids.
-  const md::UPID invalid_upid = { 0, 0, 0 };
-  const absl::flat_hash_set<md::UPID> invalid_upid_set = { invalid_upid };
+  const md::UPID invalid_upid = {0, 0, 0};
+  const absl::flat_hash_set<md::UPID> invalid_upid_set = {invalid_upid};
   ASSERT_OK(source_.ResetConnectorContext(invalid_upid_set));
 
   // Now that the sub-processes are gone, iterate over the previously found artifact paths.
@@ -588,7 +588,7 @@ TEST_F(FastPerfProfileBPFTest, TestOutOfContext) {
 
   // The "out of context" test needs to handle init., param. setup, and start each individually,
   //  so that it can pass in an empty set of upids rather than the subprocess upids.
-  const md::UPID invalid_upid = { 0, 0, 0 };
+  const md::UPID invalid_upid = {0, 0, 0};
   const absl::flat_hash_set<md::UPID> invalid_upid_set = {invalid_upid};
   ASSERT_OK(source_.Init(invalid_upid_set));
   ASSERT_OK(source_.Start());
--- a/src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
+++ b/src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
@@ -232,10 +232,6 @@ class PerfProfileBPFTest : public ::testing::TestWithParam<std::filesystem::path
     }
   }
   
-  // void TearDown() override {
-  //   PX_UNUSED(source_.Stop());
-  // }
-
   void PopulateObservedStackTraces(const std::vector<size_t>& target_row_idxs) {
     auto stack_traces_column = columns_[kStackTraceStackTraceStrIdx];
     auto counts_column = columns_[kStackTraceCountIdx];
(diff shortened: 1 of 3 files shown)


src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -211,36 +209,36 @@ class ContainerSubProcesses final : public PerfProfilerTestSubProcesses {
...
-    source_ = PerfProfileConnector::Create("perf_profile_connector");
-    ASSERT_OK(source_->Init());
-    source_->set_data_tables({&data_table_});
-
-    // Immediately start the transfer data thread to continue draining perf buffers,
-    // i.e. to prevent dropping perf buffer entries.
-    StartTransferDataThread();
+    test_run_time_ = std::chrono::seconds{FLAGS_test_run_time};
+    const auto period = std::chrono::seconds{FLAGS_stirling_profiler_table_update_period_seconds};
+    source_.RawPtr()->sampling_freq_mgr().set_period(period);
```
ddelnano, line 231: I find it difficult to reason about these flags because we are introducing a new point in test execution where these flags can be modified.

While it seems safe to change the "fast params" after the unit connector is `Init`'ed, I think this introduces the possibility of experiencing stale reads (where a flag is read before its value is updated).

I would prefer if these test case variants were composed with Google test components (Type tests, parameterized tests. etc). That would allow these flags to be set before any application code runs and stays more consistent with our other test cases.

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -317,88 +314,52 @@ class PerfProfileBPFTest : public ::testing::TestWithParam<std::filesystem::path
...
-    source_->TransferData(ctx_.get());
+  Status InitAndStartSource(const bool fast_test) {
+    // These three commands need to run in sequence, i.e. because SetFastTestParameters() makes
+    // a call into the source connector freq. manager, it must be after init and before start.
+    PX_RETURN_IF_ERROR(source_.Init(sub_processes_->upids()));
+    if (fast_test) {
+      SetFastTestParameters();
+    }
+    PX_RETURN_IF_ERROR(source_.Start());
+    return Status::OK();
```
ddelnano, line 338: This can be simplified to the following:

```
return source_.Start()
```

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -317,88 +314,52 @@ class PerfProfileBPFTest : public ::testing::TestWithParam<std::filesystem::path
...
+  Status RunTestCore() {
     const auto start_time = std::chrono::steady_clock::now();
 
     // While we sleep (here), the transfer data thread is running.
     // In that thread, TransferData() is invoked periodically.
     std::this_thread::sleep_for(test_run_time_);
-    StopTransferDataThread();
 
     // We return the amount of time that we ran the test; it will be used to compute
     // the observed sample rate and the expected number of samples.
```
ddelnano, line 349: This comment is out of date (duration return value was removed).

On that note, I have a slight preference to keeping the assertion functions as taking a duration as an argument. While the current API this test class exposes makes it easy to write a new test, I find requiring an argument for a test assertion makes the API difficult to misuse (if you happened to call the elapsed time population and test assertion functions in the wrong order).

I'm ok with either option: removing this comment to make the new code consistent or keeping the previous test assertions / duration recording inputs. Just mentioning my slight preference this since it was related to the comment.

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -528,18 +499,12 @@ TEST_P(PerfProfileBPFTest, PerfProfilerJavaTest) {
...
-  // list of upids based on our subprocs.
-  // For that to happen, here, we reset the context so that it has no UPIDs.
-  // The ProcTracker (inside of PerfProfileConnector) will take the difference
-  // between the previous list of upids (our subprocs) and the current list of upids (empty)
-  // to find a list of deleted upids.
-  const absl::flat_hash_set<md::UPID> empty_upid_set;
-  RefreshContext(empty_upid_set);
+  // Wait for the context to refresh. After the context is refreshed, the perf profiler will
+  // detect processes that have stopped, and remove their symbol files.
+  sleep(1);
```
ddelnano, line 504: Is this to ensure the [artifacts path assertion](https://github.com/pixie-io/pixie/pull/1283/files#diff-81fb698ef6614ee6ef234ad5e7ce700d06ae8ffefa6f98cb67ea577065bc2707R510) is not flaky?

Can we leverage the Timeout class added in #1281 to remove the sleep?

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
etep: Split the test cases by `FastPerfProfileBPFTest` and `SlowPerfProfileBPFTest`.

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
etep: This test case was fully re-written. We use timeout to wait for things to happen (e.g. wait for symbolization artifacts).

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
etep: Would rather keep as is. The "main" test cases are all self contained now. The symbolization artifacts removal test case and the out-of-context test case do not use this at all.

ddelnano (approved):

src/stirling/source_connectors/perf_profiler/perf_profiler_bpf_test.cc
```diff
@@ -210,37 +209,37 @@ class ContainerSubProcesses final : public PerfProfilerTestSubProcesses {
   std::vector<std::unique_ptr<ContainerRunner>> sub_processes_;
 };
 
+template <bool FastTest>
```
JamesMBartlett, line 212: Could you explain fast tests vs slow tests? I don't really understand what we're trying to achieve with the distinction.

etep: The "slow" test accepts all the default profiler parameters, but with the 30 sec. data sampling periodicity, the test inherently runs slower. The "fast" test overrides the defaults, but the test case will complete in less time. At some point in the history, we had (or were going toward) all tests using the "fast" methodology but a reviewer suggested that we keep at least one test with the slow/default setup. This PR does not change the existing setup, but it does provide a nicer way to identify which test is in which category by naming the templated test fixtures accordingly.

JamesMBartlett (approved):

PR #1289, opened 2023-05-05
[qemu] Interactive qemu runner

JamesMBartlett:
Summary: Adds a rule that creates an executable that will run qemu in interactive mode. QEMU will be launched, and then ssh'd into so that you get a normal shell instead of a serial console. You can run `bazel run //bazel/test_runners/qemu_with_kernel:run_interactive` to get a blank qemu without any test runfiles. Alternatively, you can run `bazel run //<path/to/bpf/test>_qemu_interactive` for any bpf test to get a qemu environment that has the test and its runfiles, eg `bazel run //src/stirling/e2e_tests:probe_cleaner_bpf_test_qemu_interactive`. Note that you don't specify `--config=qemu-bpf`, because of how bazel's `--run_under` works, that config flag wont work.

Test Plan: Tested a couple of different bpf tests in the interactive qemu and they all worked. Also tested that qemu isn't zombied when you exit.

### DIFF
--- a/bazel/cc_toolchains/sysroots/sysroots.bzl
+++ b/bazel/cc_toolchains/sysroots/sysroots.bzl
@@ -34,9 +34,9 @@ SYSROOT_LOCATIONS = dict(
         urls = ["https://storage.googleapis.com/pixie-dev-public/sysroots/pl6/sysroot-amd64-test.tar.gz"],
     ),
     sysroot_x86_64_glibc2_36_debug = dict(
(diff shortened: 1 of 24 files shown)


JamesMBartlett:
Lint will pass after #1288 lands

JamesMBartlett:
I'm going to make the normal tests run through ssh also to cleanup the logs.

JamesMBartlett:
I'm going to make the normal tests run through ssh also to cleanup the logs.

PR #1290, opened 2023-05-05
Move `pprof` into `shared` sub-tree.

etep:
Summary: We meed to move this code into a location that can be shared between `stirling` and `carnot`.

Type of change: /kind cleanup

Test Plan: The pprof export code will be integrated into a future diff. We have tested it locally using the `stirling_profiler` binary. Its test case will live inside of the (future work) `carnot` UDA that generates a profiling pprof protobuf for export.


ddelnano (approved):

src/shared/pprof/pprof.cc
```diff
@@ -16,24 +16,21 @@
...
 
-DECLARE_uint32(stirling_profiler_stack_trace_sample_period_ms);
-
 namespace px {
-namespace stirling {
+namespace shared {
 
-PProfProfile CreatePProfProfile(const uint32_t num_cpus, const histo_t& histo) {
+PProfProfile CreatePProfProfile(const uint32_t num_cpus, const uint32_t period_ms,
+                                const histo_t& histo) {
```
JamesMBartlett, line 28: nit: should follow style guide for type names https://google.github.io/styleguide/cppguide.html#Type_Names. I know this isn't really followed in `src/stirling` but we do follow it in `src/shared`

etep: Removed the `histo_t` alias and replaced it directly with the underlying.

JamesMBartlett (approved):

Issue #1111, opened 2023-03-28
PEM crash with terminating due to uncaught exception of type std::bad_alloc: std::bad_alloc


caphrim007:
**Describe the bug**

On a subset of our nodes, we are receiving a crash and subsequent stack trace that begins with the following

```
libc++abi: terminating due to uncaught exception of type std::bad_alloc: std::bad_alloc
E20230328 21:00:40.293191 11213 signal_action.cc:63] Caught Aborted, suspect faulting address 0x26ae. Trace:
```

**To Reproduce**

at the moment we're not sure how to specifically reproduce it. we are deploying the px-operator with no specific configuration beyond the default

**Expected behavior**

we dont observe these crashes on all nodes, only a subset of them. when we delete the PEM from the node and wait for the DaemonSet to re-add it, it continues to CLB.

**Screenshots**
If applicable, add screenshots to help explain your problem. Please make sure the screenshot does not contain any sensitive information such as API keys or access tokens.

**Logs**
Please attach the logs by running the following command:

[pixie_logs_20230328140709.zip](https://github.com/pixie-io/pixie/files/11094217/pixie_logs_20230328140709.zip)

**App information (please complete the following information):**
- Pixie version: PEM version is 0.12.18, is that what is being asked for here?
- K8s cluster version: v1.23.13-eks-fb459a0
- Node Kernel version: 5.4.226-129.415.amzn2.x86_64
- Browser version: Chrome 111.0.5563.110 (Official Build) (x86_64)

**Additional context**

ddelnano:
Hi @caphrim007. I've copied the relevant logs from your zip file below, so we can more easily reference the failures.

<details>

<summary>Stack traces from failing pods</summary>

```
# pl__vizier-pem-6bn7r__pem.log
libc++abi: terminating due to uncaught exception of type std::bad_alloc: std::bad_alloc
E20230328 21:05:49.930624 21818 signal_action.cc:63] Caught Aborted, suspect faulting address 0x528b. Trace:
**************************
PC: @     0x7f4437e1f472  (unknown)  abort
    @     0x55efe62087b6  (unknown)  abort_message
    @     0x55efe61f265b  (unknown)  demangling_terminate_handler()
    @     0x55efe6208433  (unknown)  std::__terminate()
    @     0x55efe620a156  (unknown)  __cxxabiv1::failed_throw()
    @     0x55efe620a0ef  (unknown)  __cxa_throw
[...]

PR #1291, opened 2023-05-05
Add bounds checking to prevent bogus ELF bytecode parsing from causing allocation errors

ddelnano:
Summary: Add bounds checking to prevent bogus ELF bytecode parsing from causing allocation errors

#1111 documents a case where attempting to deploy Go uprobes to a particular application results in `std::bad_alloc` errors. The specific application that caused this behavior was a go 1.13 binary built using include GOOS=linux GOARCH=386 CGO_ENABLED=0.

The bytecode threshold was determined by inspecting Go's `crypto/tls.(*Conn).Write` and `crypto/tls.(*Conn).Read` symbol sizes with an additional 100x multiplier applied (see Test plan for more details).

Relevant Issues: #1111

Type of change: /kind bug

Test Plan: Created a [test program](https://github.com/pixie-io/pixie/commit/51bef3c04331d13e635532bf0dc9d23e082044fa) to isolate the ELF and DWARF parsing code to reproduce the crash and validated the following:
- [x] Verified that the test program no longer throws a `std::bad_alloc` error ([P369](https://phab.corp.pixielabs.ai/P369))
- [x] Searched for other usages of `px::utils::u8string` to determine if other usages needed to be protected
- [x] Verified that Go's `crypto/tls.(*Conn).Write` and `crypto/tls.(*Conn).Read` symbol bytecode is ~2 KiB in size ([P370](https://phab.corp.pixielabs.ai/P370)).

Changelog Message: Add bounds checking to prevent a situation where certain Go binaries can cause Pixie to crash with a memory allocation error.

### DIFF
--- a/src/stirling/obj_tools/elf_reader.cc
+++ b/src/stirling/obj_tools/elf_reader.cc
@@ -617,7 +617,7 @@ StatusOr<utils::u8string> ElfReader::SymbolByteCode(std::string_view section,
   // To protect against our ELF parsing logic locating bogus memory, set a bound on
   // how large of a string we will allocate. SymbolByteCode's main use case is to determine
   // return instructions for the crypto/tls.(*Conn).Write and crypto/tls.(*Conn).Read Go functions.
-  // These symbols' are roughly 2 KiB and were used to inform the threshold below. We apply
+  // These symbols are roughly 2 KiB and were used to inform the threshold below. We apply
   // an additional 100x multiplier for additional headroom.
   // See https://github.com/pixie-io/pixie/issues/1111 for more details.
   if (symbol.size > 100 * 2048) {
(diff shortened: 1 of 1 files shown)


src/stirling/obj_tools/elf_reader.cc
```diff
@@ -614,6 +614,15 @@ StatusOr<utils::u8string> ElfReader::SymbolByteCode(std::string_view section,
   if (!ifs.seekg(offset)) {
     return error::Internal("Failed to seek position=$0 in binary=$1", offset, binary_path_);
   }
+  // To protect against our ELF parsing logic locating bogus memory, set a bound on
+  // how large of a string we will allocate (100 MB). See
+  // https://github.com/pixie-io/pixie/issues/1111 for more details.
+  if (symbol.size > 100 * 1024 * 1024) {
```
JamesMBartlett, line 620: Do you think it makes sense to try and address the bug that causes the symbol size to be so large, rather than just protecting against it?

etep (approved):

JamesMBartlett (approved):

PR #1292, opened 2023-05-05
[qemu] Use ssh to run tests inside of qemu machine

JamesMBartlett:
Summary: QEMU tests now start a qemu instance in the background and then SSH into the machine to run tests. This makes the logs much cleaner and also enables running interactively in a normal terminal instead of a serial console.

Test Plan: Tested that `bazel test --config=qemu-bpf //...` works (apart from a flake in stirling error bpf test that I'll address separately).


src/stirling/source_connectors/stirling_error/BUILD.bazel
```diff
@@ -44,7 +44,6 @@ pl_cc_test(
         px_jattach,
         "//src/stirling/source_connectors/perf_profiler/testing/java:java_image_base-java-profiler-test-image-omit-frame-pointer.tar",
     ],
-    flaky = True,
```
vihangm, line 47: did you mean to remove this? if so perhaps pull into a separate PR

vihangm (approved):
LGTM but perhaps @zasgar should take a quick pass too.

PR #1294, opened 2023-05-05
[bpf tests] Remove timing dependence from stirling_error_bpf_test

JamesMBartlett:
Summary: `stirling_error_bpf_test` relied a lot on timing in order to make its assertions. This removes all timing dependence, and instead waits for the expectations or timesout.

Test Plan: Ran `bazel test --config=qemu-bpf //src/stirling/source_connectors/stirling_error:stirling_error_bpf_test --runs_per_test=50` without a single failure.

### DIFF
--- a/src/stirling/source_connectors/stirling_error/stirling_error_bpf_test.cc
+++ b/src/stirling/source_connectors/stirling_error/stirling_error_bpf_test.cc
@@ -18,6 +18,7 @@
 
 #include <absl/functional/bind_front.h>
 #include <absl/strings/substitute.h>
+#include <type_traits>
 
 #include "src/carnot/planner/probes/tracepoint_generator.h"
 #include "src/common/base/base.h"
@@ -56,6 +57,7 @@ using ::testing::UnorderedElementsAre;
 
 using ::px::stirling::profiler::testing::GetAgentLibsFlagValueForTesting;
 using ::px::stirling::profiler::testing::GetPxJattachFlagValueForTesting;
+using ::px::stirling::testing::WaitAndExpectRecords;
 using ::px::testing::BazelRunfilePath;
 using ::px::types::ColumnWrapperRecordBatch;
 using ::px::types::Int64Value;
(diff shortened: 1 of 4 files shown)


src/stirling/testing/common.h
```diff
@@ -180,6 +186,41 @@ class Timeout {
...
+std::vector<TRecord> WaitAndExpectRecords(
+    TGetRecords get_records, std::vector<TRecord> expected,
+    std::chrono::nanoseconds sleep_time = std::chrono::milliseconds{200}) {
+  std::vector<TRecord> records;
+  Timeout t;
+  while (!RecordsContains(records, expected) && !t.TimedOut()) {
+    records = get_records();
+    std::this_thread::sleep_for(sleep_time);
+  }
+  EXPECT_THAT(records, IsSupersetOf(expected));
```
ddelnano, line 220: The new assertions are no longer strictly matching, correct? Was there flakiness with strict matching or did it just better match with this new style of testing (looping until all the records exist)?

src/stirling/source_connectors/stirling_error/stirling_error_bpf_test.cc
```diff
@@ -479,6 +471,11 @@ std::filesystem::path BazelJavaTestAppPath(const std::string_view app_name) {
 }  // namespace
 
 TEST_F(StirlingErrorTest, PerfProfilerNoPreserveFramePointer) {
+  if (std::getenv("TESTING_UNDER_QEMU") != nullptr) {
+    // This test crashes under qemu for unknown reasons, with a dcheck failure of:
+    // error_code == kPerfProfilerStatusOk (1 vs 0)
+    GTEST_SKIP() << "Skipping this test under qemu";
```
ddelnano, line 477: This likely means that the records aren't being processed quickly enough. We can take a deeper look into why that might be.

src/stirling/testing/common.h
JamesMBartlett: There are some tests that weren't doing strict matches and some that were, I felt like standardizing on this made sense, but open to opinions

ddelnano: Sounds good, just wanted to understand the rationale.

I agree with opting for the less strict matches when it's not crucial to the specific test scenario. When we deflaked a bunch of trace bpf tests back in Jan/Feb, moving to the less strict matching helped in a few scenarios.

JamesMBartlett: FWIW the size checks are still there for the ones that had them

ddelnano (approved):

src/stirling/source_connectors/stirling_error/stirling_error_bpf_test.cc
vihangm: Leave a TODO here for someone to investigate.

vihangm (approved):

PR #1295, opened 2023-05-05
[releases] Don't create github releases for CLI rcs

JamesMBartlett:
Summary: Don't create github releases for CLI RCs.

Type of change: /kind cleanup

Test Plan: N/A


JamesMBartlett:
Going to land this despite the cli-release failing, since the failure is on the notarization/apple side and unrelated to these changes. Will address that issue separately.

PR #1302, opened 2023-05-08
[qemu] Add interactive runner using ssh

JamesMBartlett:
Summary: Adds a way to get a shell inside qemu using SSH. To run the interactive qemu session without any test runfiles, run: `bazel run //bazel/test_runners/qemu_with_kernel:run_interactive`.
To run with a test's runfiles, run `bazel run //<path/to/test>_qemu_interactive`, eg. `bazel run //src/stirling/obj_tools:elf_reader_symbolizer_bpf_test_qemu_interactive`. Note it won't work if you add `--config=qemu-bpf` because of the way bazel's `run_under` flag works, so just leave out the `--config`.

Type of change: /kind test-infra.

Test Plan: Tested that the interactive runner opens a shell with the expected runfiles. Also tested that closing the shell at any point doesn't leave zombie qemu processes.


JamesMBartlett:

zasgar (approved):

PR #1308, opened 2023-05-09
Update `tcp_stats_bpf_test` to work with qemu

ddelnano:
Summary: Update `tcp_stats_bpf_test` to work with qemu

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Verified that the test ran successfully for our various test runners
- [x] Running test multiple times within docker to verify that adding the `nc` server doesn't cause flakiness ([P376](https://phab.corp.pixielabs.ai/P376)). Also verified that each test is in its own network namespace so that the servers can't conflict with each other ([P378](https://phab.corp.pixielabs.ai/P378))
- [x] Running test within qemu passes ([P377](https://phab.corp.pixielabs.ai/P377))
- [x] Verified that when a test times out that the test prints the accumulated records found (otherwise it's difficult to see why the assertion failed) - [P379](https://phab.corp.pixielabs.ai/P379)

### DIFF
--- a/src/stirling/source_connectors/tcp_stats/tcp_stats_bpf_test.cc
+++ b/src/stirling/source_connectors/tcp_stats/tcp_stats_bpf_test.cc
@@ -23,7 +23,9 @@
 
 #include "src/common/base/base.h"
 #include "src/common/exec/exec.h"
+#include "src/common/exec/subprocess.h"
 #include "src/common/testing/testing.h"
+#include "src/shared/types/column_wrapper.h"
 #include "src/shared/types/types.h"
 #include "src/stirling/core/data_table.h"
 #include "src/stirling/source_connectors/tcp_stats/tcp_stats_connector.h"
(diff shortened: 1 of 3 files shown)


src/stirling/source_connectors/tcp_stats/tcp_stats_bpf_test.cc
```diff
@@ -46,26 +46,19 @@ class TcpTraceTest : public TcpTraceBPFTestFixture {};
...
-
-  StartTransferDataThread();
-
-  // Send "This is sample test" to 127.0.0.1, total 20 bytes of data
-  std::string cmd2 = "echo \"This is sample test\" | nc 127.0.0.1 22";
-  ASSERT_OK(px::Exec(cmd2));
-
-  StopTransferDataThread();
-  auto records2 = testing::ExtractToString(tcp_stats::kTCPStatsTable, tcp_stats_table_);
-  EXPECT_THAT(records2, HasSubstr("remote_addr:[127.0.0.1] remote_port:[22] tx:[20]"));
```
ddelnano, line 68: I decided to remove this assertion since it didn't provide any more coverage than the one earlier in the test.

src/stirling/source_connectors/tcp_stats/tcp_stats_bpf_test.cc
```diff
@@ -46,26 +46,19 @@ class TcpTraceTest : public TcpTraceBPFTestFixture {};
...
+  EXPECT_THAT(records, HasSubstr("remote_addr:[127.0.0.1] remote_port:[12345] tx:[6]"));
```
JamesMBartlett, line 61: Using `Timeout` here and looping until the assertion would reduce flakiness I think

ddelnano: Updated in latest revision. I also went through the work to populate the tcp stats data into a record struct (I thought that was a prerequisite in order to use the `Timeout`, but looking back I don't think it was necessary). In any case, it is more aligned to our other tests and makes more targeted assertions easier.

etep (approved):

PR #1316, opened 2023-05-09
[build-and-test] Remove custom name from matrix checks.

JamesMBartlett:
Summary: When there are no matrix configs to run, without this change, a check is added with an ugly ${{ matrix.name }}. This PR removes the custom naming, which removes the ugly ${{matrix.name}}

Type of change: /kind cleanup

Test Plan: Testing with this PR, since there shouldn't be any matrix configs to run.


vihangm (approved):

vihangm (approved):
sigh

PR #1324, opened 2023-05-10
Fix analytics for sign up page view

NickLanam:
Summary: TSIA.

Type of change: /kind bugfix

Test Plan: View signup page (through any means). Event should now be tracked when permitted to do so.

### DIFF
--- a/src/ui/src/components/auth/auth-box.tsx
+++ b/src/ui/src/components/auth/auth-box.tsx
@@ -23,6 +23,7 @@ import { Button, Link, Typography } from '@mui/material';
 import { Theme } from '@mui/material/styles';
 import { createStyles, makeStyles } from '@mui/styles';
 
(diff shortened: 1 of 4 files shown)


aimichelle:
How come this doesn't get tracked by the default page tracking that gets triggered whenever you go to a new URL?

NickLanam:
> How come this doesn't get tracked by the default page tracking that gets triggered whenever you go to a new URL?

From what I can tell, it's because it triggers a page refresh (though I'm not sure why it's doing that so consistently). Then, the first history event happens before the tracking is initialized. It's probably possible to fix that by buffering the event in a blocking script that can be picked up after initialization, but this band-aid works too.

aimichelle:
If we're already adding a custom event for this, can we add some analytics around whether the buttons are actually clicked?

PR #1326, opened 2023-05-10
Load cron scripts from multiple sources

m25n:
Summary: For our usage of Pixie, it's convenient to exclude Pixie's dependency on the control plane. This commit allows the query broker:
1. to load cron scripts from config maps in the same namespace
2. to turn on and off loading cron scripts from the cloud or config maps

Type of change: /kind feature

Changes: This commit implements the functionality described above.

* added a `scriptrunner.Source` type that encapsulates how cron scripts are initialized/updated.
* moved existing logic for fetching cron scripts to `scriptrunner.CloudSource`
* added `scriptrunner.ConfigMapSource`
* Modified `scriptrunner.ScriptRunner` to use `scriptrunner.Source`s.
* Added a configuration flag (`--cron_script_sources` or or `PL_CRON_SCRIPT_SOURCES`) that allows a customer to choose which sources cron scripts can be loaded from. By default, it only loads cron scripts from the control plane. If you wanted to use both, for example, you could set `PL_CRON_SCRIPT_SOURCES="cloud configmaps"`.
* Moved many of the fakes/mocks/test helpers to `helper_test.go`

Test Plan: All of these changes have been tested at the unit level in addition to manual testing in all relevant configurations. We did our manual testing on GKE, EKS, and Kind with Kubernetes version 1.25. We don't anticipate other Kubernetes platforms or versions to be a problem since the config map API is stable and consistent across those dimensions. Moving the logic to load scripts from the control plane was probably the riskiest change and might warrant some extra attention.


src/vizier/services/query_broker/script_runner/config_map_source.go
```diff
@@ -0,0 +1,124 @@
...
+
+func logCronScriptParseError(err error) {
+	log.WithError(err).Error("Failed to parse cron.yaml from configmap cron script")
+}
+
+func configmapToCronScript(configmap *v1.ConfigMap) (string, *cvmsgspb.CronScript, error) {
+	id := string(configmap.UID)
+	cronScript := &cvmsgspb.CronScript{
+		ID:      utils.ProtoFromUUIDStrOrNil(id),
+		Script:  configmap.Data["script.pxl"],
```
aimichelle, line 110: What does an example configmap look like?

m25n: Do you have a preferred location to document this?

src/vizier/services/query_broker/query_broker_server.go
```diff
@@ -205,11 +209,22 @@ func main() {
 	}()
 	defer ptProxy.Close()
 
-	// Start cron script runner.
-	sr, err := scriptrunner.New(natsConn, csClient, vzServiceClient, viper.GetString("jwt_signing_key"))
+	kubeConfig, err := rest.InClusterConfig()
```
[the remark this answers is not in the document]
m25n: And we would just pass in the namespace string? Or would we include the `viper.GetString("pod_namespace")`?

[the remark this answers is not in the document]
m25n: If it were entirely up to me, I would lean towards passing in the namespace string just to avoid propagating knowledge that there is a config option named "pod_namespace" to more than one module.

src/vizier/services/query_broker/script_runner/config_map_source.go
aimichelle: We should probably add this new mechanism + an example configmap to our docs. Probably on this page: https://docs.px.dev/tutorials/integrations/otel/ 

But for now, just for code review purposes you can just send an example as a comment here.

m25n: I also added a description to the doc string for the `ConfigMapSource` function.

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: wavefront-span-script
  namespace: pl
  labels:
    purpose: cron-script
data:
  script.pxl: |
    px.export(...)
  configs.yaml: |
    otelEndpointConfig:
      insecure: true
      url: someurl:999
  cron.yaml: |
    frequency_s: 15
```

src/vizier/services/query_broker/script_runner/script_source.go
```diff
@@ -0,0 +1,64 @@
...
+const (
+	CloudSourceName     = "cloud"
+	ConfigMapSourceName = "configmaps"
+)
+
+// DefaultSources is a list of sources enabled by default
+var DefaultSources = []string{CloudSourceName}
+
+// A Source provides an initial set of cron scripts and sends incremental updates to that set
+type Source func(ctx context.Context, updateCb func(*cvmsgspb.CronScriptUpdate)) (initial map[string]*cvmsgspb.CronScript, unsubscribe func(), err error)
```
aimichelle, line 43: Sorry for the delay! I had some more high level thoughts while I was reviewing this:
The `Source` type got a little confusing (lots of nested funcs, funcs that take funcs as args, etc).
I'm thinking this sort of architecture might look cleaner/readable:

```
type interface Source {
    GetInitialScripts() map[string]*cvmsgspb.CronScript
    Stop() error
    Start() error // does setupSource
}
```

which `CloudSource` and `ConfigMapSource` then implement. Then script runner can just iterate through the sources and call `source.GetInitialScripts()`, `source.Stop()`, etc when needed.

And rather than having an `updateCb`, which the sources call with any cronscript updates, what do you think about instead passing an `updateCh` channel to `Sources` here?
And then CloudSource and ConfigMapSource would add any new updates to `updateCh`, which `scriptrunner` can listen on.

Let me know what you think!

warshawd: This makes sense to me, we'll work on the refactor.

src/vizier/services/query_broker/script_runner/cloud_source.go
```diff
@@ -0,0 +1,275 @@
...
+	log "github.com/sirupsen/logrus"
+	"google.golang.org/grpc/metadata"
+
+	"px.dev/pixie/src/shared/cvmsgspb"
+	"px.dev/pixie/src/shared/scripts"
+	svcutils "px.dev/pixie/src/shared/services/utils"
+	"px.dev/pixie/src/vizier/services/metadata/metadatapb"
+)
+
+// CloudSource is a Source that pulls cron scripts form the cloud
```
aimichelle, line 40: Nit: Typo, `form` -> `from`. Also as per the styleguide, we like to add periods at the end of these comments.

src/vizier/services/query_broker/script_runner/cloud_source.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	source.scripts = initialScripts
+	source.stop = unsubscribe
+	return nil
+}
+
+// GetInitialScripts returns the initial set of scripts that all updates will be based on
+func (source *CloudSource) GetInitialScripts() map[string]*cvmsgspb.CronScript {
+	return source.scripts
```
aimichelle, line 85: Rather than storing `source.scripts`, since we don't keep `source.scripts` up-to-date as new scripts roll in, why don't we call `fetchInitialScripts` in here?

Alternatively, if we always call `GetInitialScripts` right after `Start`, maybe start should just return the initialScripts.

src/vizier/services/query_broker/script_runner/config_map_source.go
```diff
@@ -0,0 +1,145 @@
...
+//   - a configs.yaml which will be stored in the Configs field of [cvmsgspb.CronScript]
+//   - a cron.yaml that contains a "frequency_s" key
+func NewConfigMapSource(client clientv1.ConfigMapInterface) *ConfigMapSource {
+	return &ConfigMapSource{client: client}
+}
+
+// Start watches for updates to matching configmaps and sends resulting updates on updatesCh.
+func (source *ConfigMapSource) Start(baseCtx context.Context, updatesCh chan<- *cvmsgspb.CronScriptUpdate) (map[string]*cvmsgspb.CronScript, error) {
+	options := metav1.ListOptions{LabelSelector: "purpose=cron-script"}
+	watcher, err := source.client.Watch(baseCtx, options)
```
aimichelle, line 54: Sorry, just realized I should have called this our earlier. But we switched our k8s API watchers in the other parts of our code to use the informer model, since we've seen that it handles retries/disconnects much better. Here's an example of how we set up our informers: https://github.com/pixie-io/pixie/blob/4ee33417a5a839417f02e0faa8dada2ecee21a1d/src/vizier/services/metadata/controllers/k8smeta/k8s_metadata_utils.go#L77

m25n: We learned that adding a `ResourceEventHandler` to the informer produces add events for every initial item in the list (regardless of when you add the event handler). In an effort to satisfy the API contract of `Source`, we needed to discard the add events for the initial items in the list so that we can return the initial scripts and only send changes to that initial list.

src/vizier/services/query_broker/script_runner/config_map_source.go
```diff
@@ -20,111 +20,161 @@ package scriptrunner
...
-	watcher, err := source.client.Watch(baseCtx, options)
+func (source *ConfigMapSource) Start(_ context.Context, updatesCh chan<- *cvmsgspb.CronScriptUpdate) (map[string]*cvmsgspb.CronScript, error) {
+	stopCh := make(chan struct{})
+	isInitialized := &atomic.Bool{}
+	_, err := source.informer.Informer().AddEventHandler(configMapEventHandlers(isInitialized, updatesCh))
+	if err != nil {
+		return nil, err
+	}
+	go source.informer.Informer().Run(stopCh)
+	waitUntilSynced(source.informer.Informer())
```
aimichelle, line 76: Following https://blog.dsb.dev/posts/creating-dynamic-informers/, it may be a little nicer to use `cache.WaitForCacheSync(ctx.Done(), informer.HasSynced)` which seems more conventional.

src/vizier/services/query_broker/script_runner/config_map_source.go
```diff
@@ -20,111 +20,161 @@ package scriptrunner
...
-						Script: script,
-					},
-				},
-				RequestID: id,
-				Timestamp: time.Now().Unix(),
+func configMapEventHandlers(isInitialized *atomic.Bool, updatesCh chan<- *cvmsgspb.CronScriptUpdate) cache.ResourceEventHandlerFuncs {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			configmap := obj.(*corev1.ConfigMap)
+			if !isInitialized.Load() {
```
aimichelle, line 108: Do we need to check `isInitialized` for delete and update funcs too?

m25n: It will only send add events to build the initial list.

m25n: So what lead me to the idea is that in k8s 1.27, the [`ResourceEventHandler.OnAdd`](https://pkg.go.dev/k8s.io/client-go@v0.27.2/tools/cache#ResourceEventHandler) method has an extra argument `isInInitialList` while the other methods do not.

aimichelle (approved):

aimichelle:
The lint in the checks is timing out, but I pulled the PR manually and ran the linter myself. Lint seems fine.

PR #1336, opened 2023-05-12
Add a UDA for converting profiling data to pprof protobuf format.

etep:
Summary: We add a new UDA that converts profiling data to pprof protobuf format.

Long term, we see this as useful in several scenarios:
1. For OTel export (it seems likely that OTel will add profiling data and include pprof as a supported wire format).
2. For general purpose download, e.g. a user wants to directly download profiling data in pprof format.

Type of change: /kind feature

Test Plan: Added a new test `pprof_test` into `src/carnot/funcs/builtins`. The new test covers both the UDA and its underlying serialize and deserialize methods found in `src/shared/pprof`.

### DIFF
--- a/src/carnot/funcs/builtins/pprof_ops.h
+++ b/src/carnot/funcs/builtins/pprof_ops.h
@@ -87,10 +87,10 @@ class CreatePProfRowAggregate : public udf::UDA {
     return output;
   }
 
-  Status Deserialize(FunctionContext*, const StringValue& upstream_pprof_str) {
-    // Parse upstream data into a pprof proto object.
+  Status Deserialize(FunctionContext*, const StringValue& pprof_str) {
+    // Parse serialized input a pprof proto object.
     PProfProfile pprof;
-    if (!pprof.ParseFromString(upstream_pprof_str)) {
+    if (!pprof.ParseFromString(pprof_str)) {
       return error::Internal("Could not parse input string into a pprof proto.");
     }
 
--- a/src/carnot/funcs/builtins/pprof_ops.h
+++ b/src/carnot/funcs/builtins/pprof_ops.h
@@ -97,10 +97,10 @@ class CreatePProfRowAggregate : public udf::UDA {
     UpdateOrCheckSamplingPeriod(pprof.period() / 1000 / 1000);
 
     // Deserialize into a map from stack_trace string to count.
-    const auto upstream_histo = ::px::shared::DeserializePProfProfile(pprof);
+    const auto merge_histo = ::px::shared::DeserializePProfProfile(pprof);
 
     // Incorporate the deserialized result into our histo_.
-    for (const auto& [stack_trace, count] : upstream_histo) {
+    for (const auto& [stack_trace, count] : merge_histo) {
       histo_[stack_trace] += count;
     }
     return Status::OK();
--- a/src/carnot/funcs/builtins/pprof_ops.h
+++ b/src/carnot/funcs/builtins/pprof_ops.h
@@ -38,8 +38,11 @@ class CreatePProfRowAggregate : public udf::UDA {
     return udf::UDADocBuilder("Convert perf profiling data to pprof format.")
         .Details("Converts perf profiling stack traces into pprof format.")
         .Example(
-            R"example(df = px.DataFrame(table='stack_traces.beta', start_time='-1m')
-df = df.agg(pprof=('stack_trace', 'count', px.pprof)))example")
+            R"example(stack_traces = px.DataFrame(table='stack_traces.beta', start_time='-1m')
+sample_period = px.GetProfilerSamplingPeriodMS()
+stack_traces.asid = px.asid()
+df = stack_traces.merge(sample_period, how='inner', left_on=['asid'], right_on=['asid'])df = df.groupby(['profiler_sampling_period_ms'])
+df = df.agg(pprof=('stack_trace', 'count', 'profiler_sampling_period_ms', px.pprof)))example")
         .Arg("stack_trace", "Stack trace string.")
         .Arg("count", "Count of the stack trace string.")
         .Arg("profiler_period_ms", "Profiler stack trace sampling period in ms.")
--- a/src/carnot/funcs/builtins/pprof_ops.h
+++ b/src/carnot/funcs/builtins/pprof_ops.h
@@ -51,25 +51,16 @@ df = df.agg(pprof=('stack_trace', 'count', 'profiler_sampling_period_ms', px.ppr
 
   void Update(FunctionContext*, const StringValue stack_trace, const Int64Value count,
               const Int64Value profiler_period_ms) {
+    UpdateOrCheckSamplingPeriod(profiler_period_ms.val);
+
     histo_[stack_trace] += count.val;
-
-    // Initialize profiler_period_ms_.
-    if (profiler_period_ms_ == -1) {
-      profiler_period_ms_ = profiler_period_ms.val;
-    }
-
-    // If any inconsistent profiler period is observed, set the error flag.
-    if (profiler_period_ms_ != profiler_period_ms.val) {
-      multiple_profiler_periods_found_ = true;
-    }
   }
 
   void Merge(FunctionContext*, const CreatePProfRowAggregate& other) {
+    UpdateOrCheckSamplingPeriod(other.profiler_period_ms_);
+
     for (const auto& [stack_trace, count] : other.histo_) {
       histo_[stack_trace] += count;
-    }
-    if (profiler_period_ms_ != other.profiler_period_ms_) {
-      multiple_profiler_periods_found_ = true;
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/carnot/funcs/builtins/pprof.h
```diff
@@ -0,0 +1,98 @@
...
+namespace builtins {
+
+using px::shared::PProfProfile;
+
+class CreatePProfRowAggregate : public udf::UDA {
+ public:
+  static udf::UDADocBuilder Doc() {
+    return udf::UDADocBuilder("Convert perf profiling data to pprof format.")
+        .Details("Converts perf profiling stack traces into pprof format.")
+        .Example("df = df.agg(pprof=('stack_trace', 'count', px.pprof))")
```
ddelnano, line 43: I think it would be helpful to provide a slightly more complete example (referencing the `stack_traces.beta` data table).

As an aside, do we have plans to rename that table? It wasn't until I looked up the name of it that I realized it was suffixed with "beta".

src/carnot/funcs/builtins/pprof.h
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  void Merge(FunctionContext*, const CreatePProfRowAggregate& other) {
+    for (const auto& [stack_trace, count] : other.histo_) {
+      histo_[stack_trace] += count;
+    }
+  }
+
+  StringValue Serialize(FunctionContext*) {
+    static constexpr uint32_t stack_trace_period_ms = 11;
```
ddelnano, line 60: Should this be `FLAGS_stirling_profiler_stack_trace_sample_period_ms` rather than being hard coded?

src/carnot/funcs/builtins/pprof.h
```diff
@@ -0,0 +1,98 @@
...
+    const uint32_t ncpus = get_nprocs_conf();
```
ddelnano, line 61: Are UDAs guaranteed to execute on the instance that collected the data? I don't have a great mental model of what happens at the kelvin vs the PEM level.

etep: No, that is not guaranteed. Agreed that we are not doing the full right thing here, but this is the same as what we have for the pxl script for flamegraphs right now. I wonder if we should add an issue to extend the perf profile source connector to collect this info inline with stack trace collection.

ddelnano: Thanks for clarifying. Creating an issue for fixing this in the future sounds good to me. Could we link that issue here with a TODO comment?

src/carnot/funcs/builtins/pprof.h
etep: Expanded the example and opened issue #1351 to move the perf profiler table out of beta.

src/carnot/funcs/builtins/pprof.h
etep: We removed the cpu count as an input to pprof.

src/carnot/funcs/builtins/pprof.h
etep: Currently keeping this as hard coded. Added issue #1352. Open to further discussion.

src/carnot/funcs/builtins/pprof_ops.h
```diff
@@ -0,0 +1,99 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <sys/sysinfo.h>
+
```
vihangm, line 22: seems unused?

src/carnot/funcs/builtins/pprof_ops.h
```diff
@@ -0,0 +1,99 @@
...
+#include <absl/container/flat_hash_map.h>
+
+#include <string>
+
+#include "src/shared/pprof/pprof.h"
+
+#include "src/carnot/udf/registry.h"
+#include "src/carnot/udf/udf.h"
```
vihangm, line 30: Fix include order to conform to styleguide https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

src/carnot/funcs/builtins/pprof_ops.h
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  void Merge(FunctionContext*, const CreatePProfRowAggregate& other) {
+    for (const auto& [stack_trace, count] : other.histo_) {
+      histo_[stack_trace] += count;
+    }
+  }
+
+  StringValue Serialize(FunctionContext*) {
+    static constexpr uint32_t stack_trace_period_ms = 11;
```
vihangm, line 62: I assume this is expected to line up with the flag `stirling_profiler_stack_trace_sample_period_ms`?
Seems strange and incorrect to hard code it here. What happens if someone overrides the flag?

etep: Agreed, there is a potential mismatch. We added issue #1352 for this. Curious to get feedback, should we:
1. Pursue #1352 first (and mark this as draft) or,
2. Just land this and triage #1352 instead?
i.e. because although it is possible for the parameters to be out-of-sync we don't think this is a common setup (and since the use case for export isn't fully end-to-end setup yet, we have time also).

src/carnot/funcs/builtins/pprof_ops_test.cc
```diff
@@ -0,0 +1,138 @@
...
+
+#include <google/protobuf/text_format.h>
+#include <gtest/gtest.h>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "src/carnot/funcs/builtins/pprof_ops.h"
+#include "src/carnot/udf/test_utils.h"
+#include "src/shared/pprof/pprof.h"
```
vihangm, line 27: nit: include orders need to be fixed in all files.

src/shared/pprof/pprof.cc
```diff
@@ -16,11 +16,13 @@
...
  */
 
-#include "src/shared/pprof/pprof.h"
-
+#include <absl/strings/str_join.h>
 #include <absl/strings/str_split.h>
 #include <vector>
 
+#include "src/common/base/base.h"
+#include "src/shared/pprof/pprof.h"
```
vihangm, line 24: nit: fix according to style guide

src/carnot/funcs/builtins/pprof_ops.h
vihangm: IMO it's a bit strange to land this PR first. Even though something is unlikely to happen, we should avoid leaving the codebase in a state where one could encounter unexpected behavior.
My vote would be option 1.

etep: Converted to draft. Will pursue #1352 first.

vihangm (changes requested):
Tests seem to be failing.

src/carnot/funcs/builtins/pprof_ops.h
```diff
@@ -0,0 +1,124 @@
...
+class CreatePProfRowAggregate : public udf::UDA {
+ public:
+  static udf::UDADocBuilder Doc() {
+    return udf::UDADocBuilder("Convert perf profiling data to pprof format.")
+        .Details("Converts perf profiling stack traces into pprof format.")
+        .Example(
+            R"example(stack_traces = px.DataFrame(table='stack_traces.beta', start_time='-1m')
+sample_period = px.GetProfilerSamplingPeriodMS()
+stack_traces.asid = px.asid()
+df = stack_traces.merge(sample_period, how='inner', left_on=['asid'], right_on=['asid'])df = df.groupby(['profiler_sampling_period_ms'])
```
vihangm, line 44: missing a newline between the merge and the groupby.

also we canonically chain the groupby and agg statements, i.e. `df = df.groupby(...).agg(...)`

vihangm: Also the example string is processed and the formatting is changed.
I believe you need to include the pipe char to include newlines in the generated doc, see `QuantilesUDA::Doc` for example.

src/carnot/funcs/builtins/pprof_ops.h
```diff
@@ -0,0 +1,124 @@
...
+    const auto pprof = px::shared::CreatePProfProfile(profiler_period_ms_, histo_);
+    std::string output;
+    const bool ok = pprof.SerializeToString(&output);
+    if (!ok) {
+      return "Protobuf `SerializeToString` failed.";
+    }
+    return output;
+  }
+
+  Status Deserialize(FunctionContext*, const StringValue& upstream_pprof_str) {
```
vihangm, line 81: minor nit: I don't like the name `upstream_pprof_str`. `upstream` is kinda confusing.
rename to `pprof_str`

vihangm (approved):

src/carnot/funcs/builtins/pprof_ops.h
etep: > also we canonically chain the groupby and agg statements, i.e. `df = df.groupby(...).agg(...)`

I had them split to avoid a long line. Will experiment with the formatting using the chained groupby + agg.

etep: > missing a newline between the merge and the groupby.

yikes, good catch, ty.

vihangm (approved):

PR #1337, opened 2023-05-15
Enable TLS tracing for applications using dynamically linked OpenSSL v3

ddelnano:
Summary: Enable TLS tracing for applications using dynamically linked OpenSSL v3

Relevant Issues: #692

Type of change: /kind feature

Test Plan: Existing tests provide the necessary coverage

Changelog Message:
```release-note
TLS tracing now supports applications using OpenSSL v3
```

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
+++ b/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
@@ -149,7 +149,7 @@ DEFINE_uint64(max_body_bytes, gflags::Uint64FromEnv("PL_STIRLING_MAX_BODY_BYTES"
 
 DEFINE_bool(
     access_tls_socket_fd_via_syscall,
-    gflags::BoolFromEnv("PL_ACCESS_TLS_SOCKET_FD_VIA_SYSCALL", false),
+    gflags::BoolFromEnv("PL_ACCESS_TLS_SOCKET_FD_VIA_SYSCALL", true),
     "If true, stirling will identify a socket's fd based on the underlying syscall (read, write, "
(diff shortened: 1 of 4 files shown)


src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -149,7 +149,7 @@ DEFINE_uint64(max_body_bytes, gflags::Uint64FromEnv("PL_STIRLING_MAX_BODY_BYTES"
 
 DEFINE_bool(
     access_tls_socket_fd_via_syscall,
-    gflags::BoolFromEnv("PL_ACCESS_TLS_SOCKET_FD_VIA_SYSCALL", false),
+    gflags::BoolFromEnv("PL_ACCESS_TLS_SOCKET_FD_VIA_SYSCALL", true),
```
etep, line 152: Questions about this bool flag:
1. Would it make sense to rename as "disable nested syscall fd" or similar, and have the default as false?
2. Do we really need it at all?

etep (approved):

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
ddelnano: 1. Renaming it means that our Launch Darkly feature flag will need to changed as well (if we were to use it).

For 2, my thinking was that we can clean up this feature flag after one more week with the new tracing implementation active. Our verification has been running in the background for close to two months now, but I thought I would give this one more week since we are expanding its surface area with this change.

etep: ok w/ me to leave "as is."

aimichelle (approved):

PR #1354, opened 2023-05-18
During deploy time operator should check for CSI Driver and switch to etcd

aimichelle:
Summary: We currently decide whether or not to deploy an etcd-backed metadata store or pv-backed metadata store if a user doesn't have any default storage classes set.
However, we've been seeing instances where the storageclass exists, but persistent volumes still can't be used unless the CSI driver is enabled. This leads to many failed deployments. Instead, we can autodetect this case since it is fairly common.

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Skaffold deploy operator to a cluster

### DIFF
--- a/src/operator/controllers/vizier_controller.go
+++ b/src/operator/controllers/vizier_controller.go
@@ -127,8 +127,8 @@ func getLatestVizierVersion(ctx context.Context, client cloudpb.ArtifactTrackerC
 	return resp.Artifact[0].VersionStr, nil
 }
 
-// missingNecessaryCSIDriver checks if the user is running an EKS cluster, and if so, whether they have
-// the CSIDriver enabled such that persistent volumes can be deployed.
+// missingNecessaryCSIDriver checks if the user is running an EKS cluster, and if so, whether they are
+// missing the CSIDriver. Without the CSI driver, persistent volumes may not be able to be deployed.
 func missingNecessaryCSIDriver(clientset *kubernetes.Clientset, k8sVersion string) bool {
 	// This check only needs to be done for eks clusters with K8s version > 1.22.0.
 	if !strings.Contains(k8sVersion, "-eks-") {
--- a/src/operator/controllers/vizier_controller.go
+++ b/src/operator/controllers/vizier_controller.go
@@ -127,32 +127,32 @@ func getLatestVizierVersion(ctx context.Context, client cloudpb.ArtifactTrackerC
 	return resp.Artifact[0].VersionStr, nil
 }
 
-// validCSIDriver checks if the user is running an EKS cluster, and if so, whether they have
+// missingNecessaryCSIDriver checks if the user is running an EKS cluster, and if so, whether they have
 // the CSIDriver enabled such that persistent volumes can be deployed.
-func validCSIDriver(clientset *kubernetes.Clientset, k8sVersion string) (bool, error) {
+func missingNecessaryCSIDriver(clientset *kubernetes.Clientset, k8sVersion string) bool {
 	// This check only needs to be done for eks clusters with K8s version > 1.22.0.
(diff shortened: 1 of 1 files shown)


src/operator/controllers/vizier_controller.go
```diff
@@ -126,6 +127,34 @@ func getLatestVizierVersion(ctx context.Context, client cloudpb.ArtifactTrackerC
...
 
+// validCSIDriver checks if the user is running an EKS cluster, and if so, whether they have
+// the CSIDriver enabled such that persistent volumes can be deployed.
+func validCSIDriver(clientset *kubernetes.Clientset, k8sVersion string) (bool, error) {
+	// This check only needs to be done for eks clusters with K8s version > 1.22.0.
+	if !strings.Contains(k8sVersion, "-eks-") {
+		return true, nil
+	}
+
+	parsedVersion, err := semver.Parse(strings.TrimPrefix(k8sVersion, "v"))
```
vihangm, line 138: should we use `semver.ParseTolerant` here instead?

src/operator/controllers/vizier_controller.go
```diff
@@ -126,6 +127,34 @@ func getLatestVizierVersion(ctx context.Context, client cloudpb.ArtifactTrackerC
 	return resp.Artifact[0].VersionStr, nil
 }
 
+// validCSIDriver checks if the user is running an EKS cluster, and if so, whether they have
+// the CSIDriver enabled such that persistent volumes can be deployed.
+func validCSIDriver(clientset *kubernetes.Clientset, k8sVersion string) (bool, error) {
+	// This check only needs to be done for eks clusters with K8s version > 1.22.0.
+	if !strings.Contains(k8sVersion, "-eks-") {
+		return true, nil
+	}
```
vihangm, line 136: maybe as a followup we should consider pulling this detection logic and using something that we share with https://github.com/pixie-io/pixie/blob/main/src/pixie_cli/pkg/utils/checks.go#L130-L154

src/operator/controllers/vizier_controller.go
```diff
@@ -126,6 +127,34 @@ func getLatestVizierVersion(ctx context.Context, client cloudpb.ArtifactTrackerC
 	return resp.Artifact[0].VersionStr, nil
 }
 
+// validCSIDriver checks if the user is running an EKS cluster, and if so, whether they have
+// the CSIDriver enabled such that persistent volumes can be deployed.
+func validCSIDriver(clientset *kubernetes.Clientset, k8sVersion string) (bool, error) {
```
vihangm, line 132: I think returning an error from this func doesn't make much sense since we log it in this func and from the caller too, we should do one or the other.

vihangm: Also consider renaming this func, logically it's "(doesn't need CSI) or (needs CSI and has CSI)"
perhaps `validCSIDriverIfRequired` or something?
or perhaps a negation is better. i.e. a func that detects `missingNecessaryCSIDriver`

src/operator/controllers/vizier_controller.go
```diff
@@ -127,32 +127,32 @@ func getLatestVizierVersion(ctx context.Context, client cloudpb.ArtifactTrackerC
 	return resp.Artifact[0].VersionStr, nil
 }
 
-// validCSIDriver checks if the user is running an EKS cluster, and if so, whether they have
+// missingNecessaryCSIDriver checks if the user is running an EKS cluster, and if so, whether they have
 // the CSIDriver enabled such that persistent volumes can be deployed.
```
vihangm: update comment since we negated the function. so detects missing driver.

vihangm (approved):
LGTM, just need to update the func docstring.

vihangm (approved):

Issue #634, opened 2022-10-24
The Pixie Operator creates non-existent apiGroup/resource permissions combinations


benjdewan:
**Describe the bug**
The pixie-operator creates Roles that provide blanket access to manipulate non-existent combinations of apiGroups and resources. Due to the hierarchical nature of RBAC in Kubernetes this means deploying Pixie at all requires granting the operator those same permissions, even though neither the operator, nor the services it manages can make use of them.

For example, the [`pl-updater-role`](https://github.com/pixie-io/pixie/blob/main/k8s/vizier/bootstrap/updater_role.yaml#L38-L64) grants `*` permission to manage `poddisruptionbudget` resources in the `rbac.authorization.k8s.io` apiGroup even though that resource is not defined within that apiGroup.

(Another example, the [`pl-cloud-connector-ns-role`](https://github.com/pixie-io/pixie/blob/main/k8s/vizier/bootstrap/cloud_connector_role.yaml#L41-L71) grants `*` permission to manage `secret` objects in the `px.dev` apiGroup)

**To Reproduce**
Steps to reproduce the behavior in GitHub:
1. Click [here](https://github.com/pixie-io/pixie/blob/main/k8s/vizier/bootstrap/updater_role.yaml#L38-L64) or [here](https://github.com/pixie-io/pixie/blob/main/k8s/vizier/bootstrap/cloud_connector_role.yaml#L41-L71)

Steps to reproduce the behavior in Kubernetes
1. Install `Pixie`
2. Run `kubectl describe role pl-updater-role --namespace pl`
```

% kubectl describe role/pl-updater-role -n pl
Name:         pl-updater-role
Labels:       app=pl-monitoring
              component=vizier
              vizier-bootstrap=true
              vizier-name=pixie
Annotations:  vizier-name: pixie
PolicyRule:
  Resources                                    Non-Resource URLs  Resource Names  Verbs
  ---------                                    -----------------  --------------  -----
  configmaps                                   []                 []              [*]
  cronjobs                                     []                 []              [*]
  daemonsets                                   []                 []              [*]
  deployments                                  []                 []              [*]
  jobs                                         []                 []              [*]
  persistentvolumes                            []                 []              [*]
  pods                                         []                 []              [*]
  rolebindings                                 []                 []              [*]
  roles                                        []                 []              [*]
  secrets                                      []                 []              [*]
  serviceaccounts                              []                 []              [*]
  services                                     []                 []              [*]
[...]

Issue #1353, opened 2023-05-18
Unused ClusterRole and ClusterRoleBinding in Vizier?


rosenhouse:
I’m trying to make sense of some of the RBAC for Vizier:
1. A `ClusterRoleBinding` that references a `ClusterRole` named `pl-updater-role`: https://github.com/pixie-io/pixie/blob/8111b1c558a41dfb20454c8dfbfb5f61538c5114/k8s/vizier/bootstrap/updater_role.yaml#L8-L14
2. There is a `Role` with the name `pl-updater-role`, but not a `ClusterRole` with that name
3. There is a `ClusterRole` named `pl-updater-cluster-role` https://github.com/pixie-io/pixie/blob/8111b1c558a41dfb20454c8dfbfb5f61538c5114/k8s/vizier/bootstrap/updater_role.yaml#L20-L23 but [there are no references to it](https://github.com/search?q=repo%3Apixie-io/pixie%20pl-updater-cluster-role&type=code)

Based on this, it appears that `pl-updater-cluster-binding` and `pl-updater-cluster-role` are both non-functional.

If so, I can PR to remove them.

rosenhouse:
Fixed by #1356

PR #1356, opened 2023-05-21
Reduce privileges required to install Vizier

rosenhouse:
Summary: Cleans up RBAC rules for Vizier in order to reduce the privileges required to install it.  Removes `pl-updater` service account and its RBAC, which included non-functional `ClusterRole` and `ClusterRoleBinding` and doesn't appear to be used by the current operator-based upgrade workflow.

Type of change: /kind cleanup

Test Plan: Tested new yamls on a local minikube using `skaffold` as suggested in DEVELOPMENT.md, deleted `proxy-tls-certs` manually and re-ran the cert-provisioner-job successfully.  Tested change to vizier_yaml.go by running the template generator against a local build of the base yaml tar, observed that the new binding's subject namespace is templated as expected.

Fixes: #634 and #1353

### DIFF
--- a/k8s/vizier/base/metadata_role.yaml
+++ b/k8s/vizier/base/metadata_role.yaml
@@ -11,18 +11,24 @@ metadata:
 rules:
 - apiGroups:
   - ""
-  - "apps"
   resources:
   - pods
   - services
   - endpoints
   - namespaces
-  - replicasets
-  - deployments
   verbs:
   - "watch"
   - "get"
   - "list"
+- apiGroups:
+  - apps
+  resources:
+  - replicasets
+  - deployments
+  verbs:
+  - watch
+  - get
+  - list
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -72,7 +78,10 @@ rules:
   - ""
   resources:
   - endpoints
-  verbs: ["*"]
+  verbs:
+  - get
+  - list
+  - watch
 - apiGroups:
   - coordination.k8s.io
   resources:
--- a/k8s/vizier/bootstrap/cert_provisioner_role.yaml
+++ b/k8s/vizier/bootstrap/cert_provisioner_role.yaml
@@ -1,0 +1,36 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: pl-cert-provisioner-service-account
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: pl-cert-provisioner-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: pl-cert-provisioner-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: pl-cert-provisioner-role
+subjects:
+- kind: ServiceAccount
+  name: pl-cert-provisioner-service-account
+  namespace: pl
--- a/k8s/vizier/bootstrap/updater_role.yaml
+++ b/k8s/vizier/bootstrap/updater_role.yaml
@@ -5,63 +5,67 @@ metadata:
   name: pl-updater-service-account
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: pl-updater-cluster-binding
-roleRef:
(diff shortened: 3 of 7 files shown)


rosenhouse:
One thing I want to highlight is there are still some vestigial `pl-updater` bits, e.g. https://github.com/pixie-io/pixie/blob/8111b1c558a41dfb20454c8dfbfb5f61538c5114/src/vizier/services/cloud_connector/bridge/server.go#L68-L82

Because the RBAC for that service account has been broken for a while (#1353), I'm guessing that's dead code too.   I'm ok to also remove that code path through `bridge/server.go` here, or in a follow-up PR.

k8s/vizier/base/metadata_role.yaml
```diff
@@ -11,7 +11,6 @@ metadata:
 rules:
 - apiGroups:
   - ""
-  - "apps"
   resources:
```
vihangm, line 14: Can we remove `replicasets` and `deployments` from this group?

rosenhouse: Whoops! yes of course 😄 .  Fixed in latest.

k8s/vizier/bootstrap/updater_role.yaml
```diff
@@ -1,77 +0,0 @@
...
-metadata:
-  name: pl-updater-binding
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: Role
-  name: pl-updater-role
-subjects:
-- kind: ServiceAccount
-  name: pl-updater-service-account
-  namespace: pl
```
vihangm, line 77: I believe these are still used, see this inlined template: https://github.com/pixie-io/pixie/blob/main/src/vizier/services/cloud_connector/bridge/server.go#L69

This is used in non-operator deploys of Pixie where we create a update job to update pixie.

k8s/vizier/bootstrap/updater_role.yaml
```diff
@@ -1,77 +0,0 @@
----
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: pl-updater-service-account
```
vihangm, line 5: keep, see below.

vihangm (changes requested):

k8s/vizier/bootstrap/updater_role.yaml
rosenhouse: Ok... but it is almost certgainly broken in its current state: https://github.com/pixie-io/pixie/issues/1353

Do we have any tests to cover that non-operator deploy of Pixie?

Should I leave it, even though it won't work?

rosenhouse: Other alternatives that I'm open to:

1. attempt to _fix_ the ClusterRole naming issue in 1353, but again I'd be looking for guidance on how to test that change.
2. remove only the broken ClusterRole and ClusterRoleBindings identified in #1353 , but leave the Role, RoleBinding and ServiceAccount.
3. Entirely remove the non-operator update functionality??

Let me know what would be best.

vihangm: I was suggesting option 2.
i.e. keep `pl-updater-binding`, `pl-updater-role`, `pl-updater-service-account` and remove the rest.

Your assessment that the ClusterRole and the ClusterRoleBindings are incorrect seems accurate and I believe that it's a leftover from some old refactoring to scope down these roles.

vihangm: Re: non-operator deploys, they are here to stay for now and we might invest more into providing those as an explicit deployment option for pixie users. For now, we test both operator and non-operator releases/updates as part of our pre-release process. :)

vihangm (approved):

k8s/vizier/bootstrap/cert_provisioner_role.yaml
```diff
@@ -0,0 +1,36 @@
...
+metadata:
+  name: pl-cert-provisioner-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: pl-cert-provisioner-role
+subjects:
+- kind: ServiceAccount
+  name: pl-cert-provisioner-service-account
+  namespace: pl
```
aimichelle, line 36: We support Pixie deploys in namespaces other than pl. How we do that is that we actually generate templates from these base kustomize YAMLs using: https://github.com/pixie-io/pixie/blob/main/src/utils/template_generator/vizier_yamls/vizier_yamls.go

You will need to add something like this: https://github.com/pixie-io/pixie/blob/main/src/utils/template_generator/vizier_yamls/vizier_yamls.go#L404-L413
for this new rolebinding.

rosenhouse: Fixed on latest!

aimichelle (approved):

PR #1362, opened 2023-05-23
[releases] Automatically create PR to update readme on vizier release.

JamesMBartlett:
Summary: Creates a PR from the pixie-io-buildbot account to update the README with the a link to the latest vizier release. The action will be reused for the other releases in future PRs.

Type of change: /kind cleanup

Test Plan: Tested on my fork, saw that the PR was created: https://github.com/JamesMBartlett/pixie/pull/15

### DIFF
--- a/.github/workflows/vizier_release.yaml
+++ b/.github/workflows/vizier_release.yaml
@@ -66,12 +66,15 @@ jobs:
     needs: build-release
     permissions:
       contents: write
+    outputs:
+      release-url: ${{ steps.release.outputs.url }}
     steps:
     - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3  # v3.5.0
       with:
         fetch-depth: 0
     - uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a  # v3.0.2
     - name: Create Release
+      id: release
       env:
         REF: ${{ github.event.ref }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/vizier_release.yaml
```diff
@@ -87,4 +90,27 @@ jobs:
         gh release create "${TAG_NAME}" --title "Vizier ${TAG_NAME#release/vizier/}" \
           --notes $'Pixie Vizier Release:\n'"${changelog}"
         gh release upload "${TAG_NAME}" vizier-artifacts/*
+        url="$(gh release view "${TAG_NAME}" -q '.url' --json url)"
+        echo "url=${url}" >> $GITHUB_OUTPUT
       # yamllint enable rule:indentation
+  update-readme:
+    name: Update README with latest release
```
vihangm, line 97: should we gate this on non-RC builds? `if: ${{ !contains(github.event.ref, '-') }}`

ci/update_readme_with_latest_release.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -ex
```
vihangm, line 1: Remove the `-x`

.github/workflows/vizier_release.yaml
```diff
@@ -87,4 +90,27 @@ jobs:
...
+        ref: main
+    - uses: ./.github/actions/update_readme_releases
+      with:
+        component-name: "vizier"
+        release-url: ${{ needs.create-github-release.outputs.release-url }}
+        ref: ${{ github.event.ref }}
+        BUILDBOT_GH_API_TOKEN: ${{ secrets.BUILDBOT_GH_API_TOKEN }}
+        BUILDBOT_GPG_KEY_B64: ${{ secrets.BUILDBOT_GPG_KEY_B64 }}
+        BUILDBOT_GPG_KEY_ID: ${{ secrets.BUILDBOT_GPG_KEY_ID }}
+        BUILDBOT_SSH_KEY_B64: ${{ secrets.BUILDBOT_SSH_KEY_B64 }}
```
vihangm, line 116: I'm unsure if it's cleaner to do this or to instead just create another workflow that triggers on:
```
release:
    types: [published]
```

I think the latter might be cleaner so that we don't have to update every `{vizier/cli/operator/cloud}_release` workflow. And then you could just collapse all of the update_readme_releases logic into that workflow.

.github/workflows/vizier_release.yaml
JamesMBartlett: needs takes care of it.

.github/workflows/vizier_release.yaml
JamesMBartlett: yeah fair enough

PR #1365, opened 2023-05-24
Refactor analytics init behavior

NickLanam:
Summary: Makes analytics behave more predictably.
- There's a new `PixieAPIManager` that takes over most of the behavior of `PixieAPIContext`.
This is so auth and preferences can be checked from outside of React's scope.
`PixieAPIContext` is now mostly a convenience wrapper around this.
- Using this, analytics now always check opt-out state immediately when the page loads.
Previously, it wouldn't initialize unless the user visited a Live view at some point.
- Analytics now IMMEDIATELY turn on/off to match opt-out state instead of waiting for a page refresh.
Previously, the page needed to be reloaded for the change to take effect.
Once "Opt Out" is clicked, the door is slammed shut. Zero events get logged from that moment.
Several events that previously never got reported (in the auth flow) due to the above work now.
- They still only trigger if you're opted in (or not logged in yet).

Type of change: /kind cleanup

Test Plan: This one's a bit gnarly to test.
- Log out. While logged out and visiting auth pages, analytics should trigger normally.
- Log in, and opt in to analytics if not already. An `identify` call should trigger within a second or two.
- While logged in, visit `/login`.
Try switching between login/signup with the button. Relevant events should be logged.
- Opt out of analytics, stay logged in, and try the above again.
Absolutely no analytics should trigger.
- Go directly to `/admin` (with a refresh).
If opted out, it should never fire a single analytics call
If opted in, it should immediately identify, and any typical events like switching Admin tabs should be logged.
- Analytics still never trigger (even when opted in) if the current environment is not configured with keys to do so.
- Likewise for the Live views.


src/ui/src/utils/analytics.ts
```diff
@@ -31,12 +36,103 @@ declare global {
...
+      }
+    } catch {
+      // No-op: this happens when the user is not logged in; it's not an error.
+    }
+  }
+
+  private async checkOptOut() {
+    // To check if analytics are permitted to run, we test if there is a logged in user. If there is, check if they've
+    // opted out of analytics. If there isn't, they're permitted until that changes (for things like login/signup flow).
+    this.waiting = true;
```
aimichelle, line 112: Should `this.waiting` be true even when we're waiting for `autoIdentify` to run?

NickLanam: I think not, because `waiting` is about "waiting to learn if we have permission to run" rather than "waiting to make sure we're set up".

If we wanted to guarantee the `identify` call happens first in this sort of situation, we'd probably want to do something like buffering events that come in while `waiting` is true and dealing with the added complexity of that. However, we don't log events at a high enough frequency for that to be worth it I think.

PR #1366, opened 2023-05-24
[releases] Automatically create PR to update readme on releases.

JamesMBartlett:
Summary: Adds a github workflow to create a PR to update the release links in the README. Note that github won't recursively run workflows, so all of the release workflows need to use the pixie-io-buildbot token to create the releases instead of the default github action token.

Type of change: /kind cleanup

Test Plan: Tested on my fork, saw that the PR gets created as expected: https://github.com/JamesMBartlett/pixie/pull/18.

### DIFF
--- a/.github/workflows/release_update_readme.yaml
+++ b/.github/workflows/release_update_readme.yaml
@@ -1,0 +1,67 @@
+---
+name: update-readme-on-release
+on:
+  release:
+    types: [published]
+permissions:
+  contents: read
+jobs:
+  update-readme:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+    - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3  # v3.5.0
+      with:
+        ref: main
+    - name: Import GPG key
+      shell: bash
+      env:
+        BUILDBOT_GPG_KEY_B64: ${{ secrets.BUILDBOT_GPG_KEY_B64 }}
+        BUILDBOT_GPG_KEY_ID: ${{ secrets.BUILDBOT_GPG_KEY_ID }}
+      run: |
+        echo "${BUILDBOT_GPG_KEY_B64}" | base64 --decode | gpg --no-tty --batch --import
+        git config --global user.signingkey "${BUILDBOT_GPG_KEY_ID}"
+        git config --global commit.gpgsign true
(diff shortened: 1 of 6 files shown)


.github/workflows/release_update_readme.yaml
```diff
@@ -0,0 +1,68 @@
...
+        git remote add fork git@github.com:pixie-io-buildbot/pixie.git
+    - name: Update readme locally
+      shell: bash
+      env:
+        TAG_NAME: ${{ github.event.release.tag_name }}
+        URL: ${{ github.event.release.html_url }}
+      run: |
+        export COMPONENT_VERSION="${TAG_NAME#release/}"
+        export COMPONENT="${COMPONENT_VERSION%%/*}"
+        export VERSION="${COMPONENT_VERSION#*/}"
```
vihangm, line 49: I wonder if using `cut -d "/"` is more readable than bash string manipulation ...

.github/workflows/release_update_readme.yaml
```diff
@@ -0,0 +1,68 @@
...
+        BUILDBOT_GPG_KEY_B64: ${{ secrets.BUILDBOT_GPG_KEY_B64 }}
+        BUILDBOT_GPG_KEY_ID: ${{ secrets.BUILDBOT_GPG_KEY_ID }}
+      run: |
+        echo "${BUILDBOT_GPG_KEY_B64}" | base64 --decode | gpg --no-tty --batch --import
+        git config --global user.signingkey "${BUILDBOT_GPG_KEY_ID}"
+        git config --global commit.gpgsign true
+    - name: Import SSH key
+      shell: bash
+      env:
+        BUILDBOT_SSH_KEY_B64: ${{ secrets.BUILDBOT_SSH_KEY_B64 }}
```
vihangm, line 29: I think you can rely on the GH_TOKEN instead of needing a SSH_KEY if you clone over `https` instead of using the `git@` remote.
Feels marginally cleaner than doing the ssh key setup but this is fine too.

JamesMBartlett: I spent like 20min trying to get it to work with `https` but git was pushing as user `github-actions[bot]` instead of as `pixie-io-buildbot` and I tried a variety of things none of which worked, so just going to leave it with the SSH_KEY

vihangm (approved):

Issue #680, opened 2023-01-05
Add k8ssandra to the demos


philkuz:
**Is your feature request related to a problem? Please describe.**
k8ssandra is a great way to demo cassandra data. We should add it to the demos and our release test process.
https://docs-v2.k8ssandra.io/install/local/single-cluster-helm/

**Describe the solution you'd like**
```
px demo deploy k8ssandra
```

ddelnano:
I've been working on this for the Cassandra Town hall I'll be presenting at. The difficult part is allowing the `px demo` command to support the more complex k8ssandra (requires installing crds before hand or integrating with helm).

ddelnano:
With #1376 merged the demo is fully functional. We just need to upload the demo manifests to the production gcs bucket and release the `px` cli for this to be complete.

PR #1367, opened 2023-05-24
Add k8ssandra demo support to CLI

aimichelle:
Summary:
We want to add a K8ssandra demo to our list of demos. A few changes needed to be made to ensure the demo can be deployed properly:
- We should check for cert-manager as a required dep. We decided not to deploy this with the demo, since deploying 2 cert-managers can break things. If we deploy cert-manager for the user, there's a bit of bookkeeping we'd need to do if the user runs `px demo delete` to decide whether or not we should also clean up cert-manager. Instead, we'll leave this up to the user for the time-being.
- The demo deploys some CRDs. This can lead to race conditions if deploying everything in the manifest at once. To handle this, added a retry in the deploy block.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Push to dev demo manifest, `px demo deploy` with and without cert-manager installed.

### DIFF
--- a/demos/BUILD.bazel
+++ b/demos/BUILD.bazel
@@ -45,11 +45,19 @@ pkg_tar(
     strip_prefix = "online-boutique",
 )
 
+pkg_tar(
+    name = "px-k8ssandra",
+    srcs = glob(["k8ssandra/*"]),
+    extension = "tar.gz",
+    strip_prefix = "k8ssandra",
+)
+
 ARCHIVES = [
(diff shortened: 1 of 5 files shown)


vihangm (approved):

ddelnano:
Can we add https://github.com/pixie-io/pixie/issues/680 to the relevant issue section in the description?

ddelnano (approved):

Issue #1364, opened 2023-05-23
Make metadata service track the events/changes for a specific set of namespaces in multi-tenant clusters


MrAta:
**Is your feature request related to a problem? Please describe.**
Currently, metadata service keeps track of the events/changes in the entire cluster. However, the number and amount of events in big and multi-tenant clusters can make the metadata service run out of memory. As such it's desired to limit the scope of the events to a subset of the namespaces that a tenant cares about to be tracked by metadata service.

**Describe the solution you'd like**
A solution would be making the metadata service configurable to provide a list of namespaces to track the events in.

**Describe alternatives you've considered**
There's no clear alternative as of now and the metadata service runs out of memory after a while. This however seems to be a harmless event as the Pod restarts and doesn't affect the functionality of Pixie as a whole.

**Additional context**
N/A

vihangm:
This was added and released as `release/vizier/v0.14.3`
Thanks for the contribution @MrAta

PR #1368, opened 2023-05-25
Enable metadata service to track the events from a subset of namespaces

MrAta:
Summary: Enable metadata service to track the events from a subset of namespaces when desired, instead of tracking the events from the entire cluster.

Relevant Issues: #1364

Type of change: /kind feature

Test Plan: Added unit tests for the new behaviour. @JamesMBartlett tested by skaffolding a dev vizier version to a cluster both with and without `metadata_namespaces` selected, and both behaved as expected.

### DIFF
--- a/src/vizier/services/metadata/controllers/k8smeta/k8s_metadata_utils.go
+++ b/src/vizier/services/metadata/controllers/k8smeta/k8s_metadata_utils.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"time"
 
+	log "github.com/sirupsen/logrus"
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -99,13 +100,18 @@ func podWatcher(resource string, namespaces []string, ch chan *K8sResourceMessag
 	}
 
 	init := func() error {
-		// We initialize ch with the current pods to handle cold start race conditions.
-		list, err := clientset.CoreV1().Pods(v1.NamespaceAll).List(context.Background(), metav1.ListOptions{})
-		if err != nil {
-			return err
+		var podList []v1.Pod
+		// We initialize ch with the current Pods to handle cold start race conditions.
+		for _, ns := range namespaces {
+			list, err := clientset.CoreV1().Pods(ns).List(context.Background(), metav1.ListOptions{})
+			if err != nil {
+				log.WithError(err).Errorf("Failed to init %s in %s namespace.", resource, ns)
+				return err
+			}
+			podList = append(podList, list.Items...)
 		}
 
-		for _, obj := range list.Items {
+		for _, obj := range podList {
 			item := obj
 			msg := iw.convert(&item)
 			if msg != nil {
--- a/src/vizier/services/metadata/controllers/k8smeta/k8s_metadata_utils.go
+++ b/src/vizier/services/metadata/controllers/k8smeta/k8s_metadata_utils.go
@@ -35,11 +35,11 @@ import (
 )
 
 type informerWatcher struct {
-	convert func(obj interface{}) *K8sResourceMessage
-	objType string
-	ch      chan *K8sResourceMessage
-	inf     cache.SharedIndexInformer
-	init    func() error
+	convert   func(obj interface{}) *K8sResourceMessage
+	objType   string
+	ch        chan *K8sResourceMessage
+	init      func() error
+	informers []cache.SharedIndexInformer
 }
 
 func (i *informerWatcher) send(msg *K8sResourceMessage, et watch.EventType) {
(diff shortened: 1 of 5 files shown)


src/vizier/services/metadata/metadata_server.go
```diff
@@ -235,7 +236,11 @@ func main() {
 	updateCh := make(chan *k8smeta.K8sResourceMessage)
 	mdh := k8smeta.NewHandler(updateCh, k8sMds, k8sMds, nc)
 
-	k8sMc, err := k8smeta.NewController(updateCh)
+	namespaces := []string{v1.NamespaceAll}
+	if viper.IsSet("METADATA_NAMESPACES") {
+		namespaces = strings.Split(viper.GetString("METADATA_NAMESPACES"), ",")
+	}
```
vihangm, line 242: Can you explicitly declare this as a pflag in `init` and use that instead? Then you can make it a `StringSlice`.

src/vizier/services/metadata/metadata_server.go
```diff
@@ -36,6 +36,7 @@ import (
 	"go.etcd.io/etcd/client/pkg/v3/transport"
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"google.golang.org/grpc"
+	v1 "k8s.io/api/core/v1"
```
vihangm, line 39: Looks like you need to run `gazelle` to update the build files to add this dep.
If you have `bazel` and the golang toolchain setup `make gazelle` at top of tree should do the trick.

vihangm (changes requested):

src/vizier/services/metadata/metadata_server.go
```diff
@@ -73,13 +74,15 @@ func init() {
...
 	pflag.Bool("use_etcd_operator", false, "Whether the etcd operator should be used instead of the persistent version.")
+	pflag.StringSlice("metadata_namespaces", []string{v1.NamespaceAll}, "The list of namespaces to watch for metadata.")
 
 	// Metadata flags are set using the env vars in pl-cluster-config.
 	// We historically set PL_ETCD_OPERATOR_ENABLED but not PL_USE_ETCD_OPERATOR in the configmap.
 	// We also don't have a clean way to update  configmaps for existing deploys.
 	// So instead just map PL_ETCD_OPERATOR_ENABLED to use_etcd_operator to make it work.
 	// TODO: We should clean this up in the future and make these flags consistent.
 	viper.BindEnv("use_etcd_operator", "PL_ETCD_OPERATOR_ENABLED")
+	viper.BindEnv("metadata_namespaces")
```
vihangm, line 85: This isn't necessary. Our shared service setup code will bind all flags to corresponding env vars with a `PL_` prefix.

src/vizier/services/metadata/controllers/k8smeta/k8s_metadata_utils.go
```diff
@@ -51,38 +51,46 @@ func (i *informerWatcher) send(msg *K8sResourceMessage, et watch.EventType) {
...
-	// We initialize ch with the current nodes to handle cold start race conditions.
+
+	for _, ns := range namespaces {
+		factory := informers.NewSharedInformerFactoryWithOptions(clientset, 12*time.Hour, informers.WithNamespace(ns))
+		inf := factory.Core().V1().Pods().Informer()
+		iw.informers = append(iw.informers, inf)
+	}
+
+	// We initialize ch with the current Pods to handle cold start race conditions.
 	list, err := clientset.CoreV1().Pods(v1.NamespaceAll).List(context.Background(), metav1.ListOptions{})
```
vihangm, line 94: the namespaces here need to also be restricted.

vihangm:
This looks mostly ok, but I think #1595 will actually make this easier by allowing us to filter in only one location (i.e. at the SharedIndexInformerFactory creation).

Also we should see if it's possible to add some testing coverage to ensure that the namespace restriction works as expected.

MrAta:
Thanks for the insights, @vihangm!
Re #1595, either of the PRs will need a rebase and conflict resolution once the other one is merged :D I'm not sure which way is more convenient, up to you.

Re the test case, I just added a simple test case for Pod creation only. Please let me know if it sounds right to you and I can add other resource types and events too.

vihangm:
Test case looks good. Adding a little bit of simple coverage on all resources types would be nice.

MrAta:
@vihangm add more test cases for other resource types.

JamesMBartlett (approved):
LGTM, discussed with @vihangm who's out at the moment and he said its good to go from his end. So if we can get the merge conflicts fixed we should be good to go. Sorry for the delay

vihangm:
Also can you update the test plan in the PR since this now has unit tests (and also should be tested with a dev deploy of vizier before merging).

JamesMBartlett:
I tested this out on a cluster, both with all namespaces and with a subset of namespaces and both seemed to work as expected.

JamesMBartlett (approved):

PR #1374, opened 2023-05-25
Add k8s manifests for the k8ssandra px cli demo

ddelnano:
Summary: Add k8s manifests for k8ssandra px cli demo

Relevant Issues: #680

Type of change: /kind feature

Test Plan: Verify the demo works by deploying it as described below:
- [ ] Deploying the demo to minikube is successful: expected data is visible in the `px/cql_data` and `px/http_data` scripts and `px demo delete` is successful
Screenshot 2023-05-26 at 9 00 31 AM
Screenshot 2023-05-26 at 9 00 40 AM

Changelog Message:
```release-note
Add a k8ssandra demo to the `px` cli
```

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -51,6 +51,39 @@ file describes the demo scenarios.
 
 4. Copy the yaml file to `pixie/demos/kafka`.
 
+## Updating the `px-k8ssandra` demo
+
+1. Clone `https://github.com/pixie-io/spring-petclinic-reactive`
+
+2. Build the k8s manifests with kustomize
+
+    ```shell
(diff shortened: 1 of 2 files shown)


demos/px-k8ssandra.tar.gz
aimichelle, line 1: We don't need to check in the tar.

ddelnano: Thanks for catching. I accidentally committed this since I was using a local http server rather than `https://storage.googleapis.com/pl-infra-dev-artifacts/dev-demo-app ` for the demo artifacts.

aimichelle (approved):

PR #1376, opened 2023-05-26
Refine `px demo delete` logic for more complex demo applications

aimichelle:
Summary: We are working on adding the px-k8ssandra demo in our demo apps. However, this demo app is more complex. It requires cluster-scoped resources such as mutating webhooks, and proper cleanup requires the K8ssandraCluster to be deleted first or else things will hang.
We can control the delete process through labels. Adding labels to the demo resources allow us to target cluster-scoped resources that wouldn't be deleted from just deleting the namespace. We can use an additional label for resources that need to be cleaned up first.

Relevant Issues: #680

Type of change: /kind cleanup

Test Plan: Deploy px-k8ssandra demo app and delete. Make sure resources all get cleaned up and delete runs without error. Also make sure delete still works on existing demo-apps, such as px-sock-shop.

Changelog Message:
```release-note
Enhanced `px demo` to support more complex applications which require CRDs
```

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -70,9 +70,15 @@ file describes the demo scenarios.
     # Your kustomize version must be v4.5.7 or newer for these configs
 
     k8ssandra-operator-root-dir $ cd config/crd
+    # Edit config/crd/kustomization.yaml to append the following:
+    # commonLabels:
+    #   pixie-demo: px-k8ssandra
     kustomize build . > crd.yaml
 
     k8ssandra-operator-root-dir $ cd config/deployments/control-plane/cluster-scope
+    # Edit config/deployments/control-plane/cluster-scope/kustomization.yaml to append the following:
+    # commonLabels:
+    #   pixie-demo: px-k8ssandra
     kustomize build . > k8ssandra-control-plane.yaml
     ```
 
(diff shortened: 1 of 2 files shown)


vihangm (approved):
This doesn't delete cert-manager right? Since we don't track if it's installed by pixie or not?

demos/k8ssandra/k8ssandra.yaml
```diff
@@ -37326,6 +37349,9 @@ metadata:
   annotations:
     cert-manager.io/inject-ca-from: px-k8ssandra/cass-operator-serving-cert
     controller-gen.kubebuilder.io/version: v0.10.0
+  creationTimestamp: null
```
ddelnano, line 37352: What's the purpose of ensuring this is set? It seems like it was added for the `MutatingWebhookConfiguration` as well.

ddelnano (approved):
Just one minor question about why `creationTimestamp` was added in a few places.

In addition to that, maybe it's worth adding a release note for this? Something like the following:

```release-note
- Enhanced the `px demo` cli command to support more complex applications which require CRDs
```

demos/k8ssandra/k8ssandra.yaml
aimichelle: That actually got set when I ran the `kustomize`... Maybe it happened because of differing kustomize versions since generating the YAMLs for this required a newer version of kustomize.

I'll add a separate PR for upgrading kustomize, just so that's consistent.

aimichelle:
> This doesn't delete cert-manager right? Since we don't track if it's installed by pixie or not?

That is correct, we error at `px demo deploy` time if they don't have cert-manager installed already. So users are required to install cert-manager themselves, and remove it themselves.

Issue #1375, opened 2023-05-26
Large CQL `Result` frames (> 1MiB) cause poor CQL protocol stitching


ddelnano:
**High Level Plan**
- [x] Prototype new stitcher implementation that accurately matches requests and responses -- prototype developed [here](https://github.com/ddelnano/pixie/tree/ddelnano/record-and-replay-with-cql-stitcher-change)
- [ ] Convert all stitching implementations to use new interface that simplifies handling protocols with streams
- [ ] Use the new interface to productionize the CQL stitcher implementation prototyped in the first bullet.

**Background**

While creating the k8ssandra demo (#680) for the recent Cassandra Town Hall, I noticed that large CQL `Result` frames cause poor protocol stitching behavior. This manifests as `Result` frames that have incorrect latency values -- requests that take tens of milliseconds to complete can be reported on the order of minutes in some cases (as seen in the reproduction below).

**To Reproduce**
1. Deploy the `px-k8ssandra` via the px cli -- `px demo deploy px-k8ssandra`. Note: this demo isn't released at this time, but will be ~1 week from the time of this writing
2. Use the `cqlsh` cli within the cassandra container to load the [following CSV file](https://github.com/pixie-io/pixie/files/11578468/petclinic_visit_by_pet.csv) into the `petclinic_visit_by_pet` table
```
# Copy CSV file containing ~8k rows into the pod
kubectl -n px-k8ssandra cp petclinic_visit_by_pet.csv  demo-dc1-default-sts-0:/tmp/

kubectl -n px-k8ssandra exec demo-dc1-default-sts-0 -it -- cqlsh -u cassandra -p cassandra -e "COPY spring_petclinic.petclinic_visit_by_pet FROM '/tmp/petclinic_visit_by_pet.csv'"
Defaulted container "cassandra" out of: cassandra, server-system-logger, server-config-init (init)
Using 3 child processes

Starting copy of spring_petclinic.petclinic_visit_by_pet with columns [pet_id, visit_id, description, visit_date].
Processed: 8511 rows; Rate:    5921 rows/s; Avg. rate:    8434 rows/s
8511 rows imported from 1 files in 0 day, 0 hour, 0 minute, and 1.009 seconds (0 skipped).
```
3. Load the `px/cql_data` script and see that the `Result` rows show traces that report seconds of latency when `curl`ing that endpoint manually shows ~100ms response times. The `px/http_data` script correctly shows the ~100ms of latency.
```
[...]

PR #1380, opened 2023-05-30
Fix test data used in CQL parser tests

ddelnano:
Summary: Fix test data used in CQL parser tests

The CQL protocol defines that messages have a [256 MiB size limit](https://github.com/apache/cassandra/blob/fad1f7457032544ab6a7b40c5d38ecb8b25899bb/doc/native_protocol_v3.spec#LL199C24-L199C24). The CQL parser test cases used a 16 MiB message size (0x1000000) rather than one that was greater than 256 MiB. The inconsistencies in this test were identified as I've been investigating #1375.

Relevant Issues: #1375

Test Plan: Verified with gdb that each test case returns the `kInvalid` from its specific branch and existing tests pass

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/cql/parse_test.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/cql/parse_test.cc
@@ -103,8 +103,8 @@ TEST_F(CQLParserTest, BadOpcode) {
 }
 
 TEST_F(CQLParserTest, LengthTooLarge) {
-  // Length is 0x01000000
-  constexpr uint8_t kBadLengthFrame[] = {0x04, 0x00, 0x00, 0x06, 0xff, 0x01,
+  // Length is 0x020000000
+  constexpr uint8_t kBadLengthFrame[] = {0x04, 0x00, 0x00, 0x06, 0x07, 0x20,
                                          0x00, 0x00, 0x00, 0x00, 0x00};
 
   std::string_view frame_view =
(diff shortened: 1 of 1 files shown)


src/stirling/source_connectors/socket_tracer/protocols/cql/parse_test.cc
```diff
@@ -103,8 +103,8 @@ TEST_F(CQLParserTest, BadOpcode) {
 }
 
 TEST_F(CQLParserTest, LengthTooLarge) {
-  // Length is 0x01000000
-  constexpr uint8_t kBadLengthFrame[] = {0x04, 0x00, 0x00, 0x06, 0xff, 0x01,
+  // Length is 0x020000000
+  constexpr uint8_t kBadLengthFrame[] = {0x04, 0x00, 0x00, 0x06, 0x07, 0x20,
                                          0x00, 0x00, 0x00, 0x00, 0x00};
```
ddelnano, line 108: See the [CQL spec](https://github.com/apache/cassandra/blob/4d2ee09ffabf7fb086fb9e2bd042fc946e72c3ec/doc/native_protocol_v4.spec#L68-L73) for how this should be laid out. Bytes 6-9 will be be a 32 bit signed integer representing the length.

Previously the 5th byte (0xff) was set to an [invalid opcode](https://github.com/pixie-io/pixie/blob/401c92c1ed086eaa35dbe17b4fbcfd96e7b1ca4f/src/stirling/source_connectors/socket_tracer/protocols/cql/types.h#L41-L58), so [this logic](https://github.com/pixie-io/pixie/blob/401c92c1ed086eaa35dbe17b4fbcfd96e7b1ca4f/src/stirling/source_connectors/socket_tracer/protocols/cql/parse.cc#L51) was triggering instead.

etep (approved):

vihangm (approved):

Issue #1174, opened 2023-04-05
UI Skin Refresh 2023


NickLanam:
We have a refreshed design to use for Pixie's UI (almost entirely CSS - not changing features).
This issue tracks the various tasks involved.

- [x] Simplified color palette (dark mode). Colors are similar to before, but contrast and consistency is adjusted to use fewer shades overall.
- [x] Simplified color palette (light mode). Same deal, brighter look.
- [x] Widget titlebars. Move controls into them. This makes graphs, tables, and charts more consistent with one another, and easier to read.
- [x] Full-width widgets. By tweaking the use of negative space and visible borders, we can use more of the available space to show data. The titlebars above make this work.
- [x] Improved table readability.
- [x] Updated dropdown styling. This is for a more modern and consistent look. Functionality unchanged.
- [ ] Admin Views
  - [x] Admin tabs to sidebar
  - [ ] Keys tables (sticky heading, count in tab bar, clean up spacing, etc)
  - [x] Cluster overview and details
  - [ ] Overview for plugins
  - [ ] Overview for retention scripts
  - [ ] Editing retention scripts

PR #1383, opened 2023-05-30
Move admin tab headers into sidebar

NickLanam:
Summary: Puts new buttons in the sidebar when in the Admin views.
These buttons replace the tab bar for Admin, except that API Keys and Deployment keys nest in a shared tab now.
This changes how the sidebar works a little bit, but should not impact the live views.

Relevant Issues: #1174

Type of change: /kind cleanup

Test Plan: Load admin pages. The tabs should now be in the sidebar rather than below the topbar.
For the Keys tab specifically, it should still have a tab bar, and it should be sticky.
Buttons that appear on various tabs should still appear (adding users, adding keys, etc).


NickLanam:
All of the new icons, with sidebar expanded:
image

PR #1386, opened 2023-05-31
[chore] Update TypeScript to 5.0.4, fix breakage

NickLanam:
Summary: Does the following:
- Updates Yarn, eslint, esbuild, and Webpack to versions that support TypeScript 5.x.
- Re-runs `yarn dlx @yarnpkg/sdks` so that VS Code and eslint can still resolve dependencies normally.
- Fixes newly-surfaced lint errors (new version of eslint is better at finding them).
- Sets required Node version to the latest LTS as of this writing, 18.16.0, to fix Jest hanging.
- Happy side effect, building and running the UI is now a tiny bit faster just by updating the toolchain!

Type of change: /type cleanup

Test Plan: Run `yarn lint`, `yarn test`, `yarn run build_prod`, and `yarn dev`. All should still function normally, though a hair faster.


NickLanam:
Note: dependency license check is flagging dependencies that we already use and simply updated.

vihangm (changes requested):
There's a lot happening in this PR, can we split it into 3 separate PRs?
- Fix lint issues
- Upgrade yarn
- Finally upgrade all our npm deps

NickLanam:
> There's a lot happening in this PR, can we split it into 3 separate PRs?
> 
>     * Fix lint issues
> 
>     * Upgrade yarn
> 
>     * Finally upgrade all our npm deps

Done, this PR now only does the final dep update and the tooling update (those have to happen at the same time, updating TypeScript breaks the tooling but the tooling won't update without the new version of TypeScript).

vihangm (approved):

PR #1395, opened 2023-05-31
[equinix] Add k3s setup script

JamesMBartlett:
Summary: Adds a script to setup a k3s cluster on equinix. The script assumes you have nodes already provisioned with public IPs, and that your ssh key has access to those nodes.

Test Plan: Tested that the script creates a working equinix cluster. The teardown side has some fragility around the EIP being set as an addr on the loopback device, but I added a `ip addr del` that should take care of that.

### DIFF
--- a/k8s/devinfra/equinix_k3s/README.md
+++ b/k8s/devinfra/equinix_k3s/README.md
@@ -1,0 +1,29 @@
+- Install k3sup (this installs k3sup to `~/bin` if that's not in your path choose somewhere that is)
+```
+curl -sLS https://get.k3sup.dev | sh
+install k3sup ~/bin
+```
+- Make sure your ssh key has access to all of the machines you are going to use.
+- Provision an EIP. Make sure the EIP is tagged with a unique identifier.
+- Run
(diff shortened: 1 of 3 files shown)


k8s/devinfra/equinix_k3s/README.md
```diff
@@ -0,0 +1,29 @@
...
+curl -sLS https://get.k3sup.dev | sh
+install k3sup ~/bin
+```
+- Make sure your ssh key has access to all of the machines you are going to use.
+- Provision an EIP. Make sure the EIP is tagged with a unique identifier.
+- Run
+```
+EQUINIX_API_KEY=<api_key> \
+EQUINIX_PROJECT_ID=<project_id> \
+EQUINIX_FACILITY=<equinix facility of machines, eg. la4> \
```
vihangm: let's switch to metro since facilities are "deprecated"
see https://feedback.equinixmetal.com/changelog/bye-facilities-hello-again-metros

vihangm (approved):

PR #1400, opened 2023-05-31
[equinix] Add raid setup script

JamesMBartlett:
Summary: Adds a script to setup a RAID0 device of the two large SSDs on the equinix machines.

Type of change: /kind cleanup

Test Plan: Tested that the script creates a RAID0 device and mounts it at `/data`, and that it's persistent across reboots.

### DIFF
--- a/k8s/devinfra/equinix_k3s/README.md
+++ b/k8s/devinfra/equinix_k3s/README.md
@@ -1,3 +1,7 @@
+- For each server, run the following only once:
+```
+./setup_raid.sh <SERVER_IP>
+```
 - Install k3sup (this installs k3sup to `~/bin` if that's not in your path choose somewhere that is)
 ```
 curl -sLS https://get.k3sup.dev | sh
--- a/k8s/devinfra/equinix_k3s/setup_raid.sh
+++ b/k8s/devinfra/equinix_k3s/setup_raid.sh
@@ -1,0 +1,42 @@
+#!/bin/bash -e
+
+# Copyright 2018- The Pixie Authors.
+#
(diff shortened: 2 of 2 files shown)


vihangm (approved):

k8s/devinfra/equinix_k3s/README.md
```diff
@@ -1,3 +1,7 @@
+- For each server, run the following only once:
+```
+./setup_raid.sh <SERVER_IP>
```
ddelnano, line 3: @JamesMBartlett are we using Equinix reserved hardware? I know you investigated the terraform provider, but it seems like these manual steps could be pushed into terraform completely (if we were to consider using it).

The `equinix_metal_device` resource allows you to specify [custom raid and disk partitioning](https://deploy.equinix.com/developers/docs/metal/storage/custom-partitioning-raid/) with the [storage argument](https://registry.terraform.io/providers/equinix/equinix/latest/docs/resources/equinix_metal_device#storage).

Unrelated to this specific change, if the steps for setting up instances becomes cumbersome I think we would should seriously consider Terraform provisioning paired with configuration management (and potentially packer image builds). We should be able to automate these bootstrapping steps to streamline k3s cluster management (cluster creation, instance replacement, etc).

JamesMBartlett: Yeah I think at some point we'll want to extend this beyond just this equinix setup (maybe to run on tanzu, zain's garage, etc), so I do think it would be nice to put these things in some sort of chef/terraform/ansible setup.

PR #1403, opened 2023-06-01
`perf_profiler.cc|h`: Implment better metrics for overrun of expected number of stack trace samples.

etep:
Summary: We implement some additional metrics for case where the profiler detects an overrun vs. the number of expected stack trace samples. Although this overrun may not cause issues (because of over-provisioning), we want to understand the magnitude and frequency of the issue so that we can re-assess our provisioning.

Type of change: /kind feature

Test Plan: Existing tests. We know these metrics are reporting infrequent overruns, we expect to see more granular data after these new metrics roll out. Also, added some [verbose logging](https://phab.corp.pixielabs.ai/P400) to capture both expected values and actual values in the metrics. For this experiment, we lowered the provisioned number of stack traces to ensure that the overflow condition was triggered.

### DIFF
--- a/src/stirling/source_connectors/perf_profiler/perf_profile_connector.cc
+++ b/src/stirling/source_connectors/perf_profiler/perf_profile_connector.cc
@@ -376,12 +376,19 @@ void PerfProfileConnector::CheckProfilerState(const uint64_t num_stack_traces) {
 
   switch (error_code) {
     case kOverflowError: {
+      // overflow_ratio is actual:expected. That is, the actual number of stack traces sampled
+      // vs. the expected number of stack traces.
       const double overflow_ratio =
           static_cast<double>(num_stack_traces) / static_cast<double>(expected_stack_traces_);
+      profiler_state_overflow_gauge_.Set(overflow_ratio);
+
+      // We compute the increment to profiler_transfer_data_counter_ such that the counter value is
+      // equal to the total number of transfer data invocations.
       const double current_transfer_counter = profiler_transfer_data_counter_.Value();
       const double transfer_count_increment = transfer_count_ - current_transfer_counter;
+
+      // Track the total number of overflows and the total number of transfer data invocations.
       profiler_transfer_data_counter_.Increment(transfer_count_increment);
-      profiler_state_overflow_gauge_.Set(overflow_ratio);
       profiler_state_overflow_counter_.Increment();
       break;
     }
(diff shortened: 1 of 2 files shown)


src/stirling/source_connectors/perf_profiler/perf_profile_connector.cc
```diff
@@ -354,23 +360,42 @@ void PerfProfileConnector::ProcessBPFStackTraces(ConnectorContext* ctx, DataTabl
...
   DCHECK_EQ(error_code, kPerfProfilerStatusOk);
 
   switch (error_code) {
-    case kOverflowError:
+    case kOverflowError: {
+      // overflow_ratio is actual:expected. That is, the actual number of stack traces sampled
+      // vs. the expected number of stack traces.
+      const double overflow_ratio =
+          static_cast<double>(num_stack_traces) / static_cast<double>(expected_stack_traces_);
+      profiler_state_overflow_gauge_.Set(overflow_ratio);
```
ddelnano, line 383: Our metric scraping interval was changed to 15 mins (https://github.com/pixie-io/pixie/pull/1358), so we will only record the value that happens during that instant. Even with the previous 1 minute interval, there is still a risk that the gauge value fluctuates between sampling and that we are blind to its variance (possible to miss a really high, brief value).

What if we recorded the worst case overflow rather than each iteration's overflow instead? While that doesn't inform us how frequently it happens, I think that's a good indicator for if the perf profiler is performing well or not.

src/stirling/source_connectors/perf_profiler/perf_profile_connector.cc
```diff
@@ -61,6 +61,12 @@ PerfProfileConnector::PerfProfileConnector(std::string_view source_name)
       sampling_period_(
           std::chrono::milliseconds{1000 * FLAGS_stirling_profiler_table_update_period_seconds}),
       push_period_(sampling_period_ / 2),
+      profiler_state_overflow_gauge_(
+          BuildGauge("perf_profiler_overflow_gauge",
+                     "Fraction of the overflow, e.g. overflowed by 1.25x the expected amount.")),
```
ddelnano, line 66: Shouldn't this value should be < 1? Since we are tracking `actual / expected` if there is overflow I would assume actual would be less than expected.

etep: The "error" condition is triggered when the actual count exceeds the threshold amount. The threshold amount is set 1/2 way between the "expected" value and the "over-provisioned" max value. Our current over-provision factor is 3x. That gives us: expected=1x, error-threshold=2x, configured-max=3x. So when we see this ratio reported, hopefully we see a number that is closer to 2x and not exceeding 3x. Will update the help message to be more precise.

ddelnano (approved):

vihangm (approved):

PR #1408, opened 2023-06-01
Add github action for updating autogenerated pxl and python api docs

ddelnano:
Summary: Add github action for updating autogenerated pxl and python api docs

Relevant Issues: N/A

Type of change: /kind feature

Test Plan: The following PR was successfully created with the missing docs https://github.com/pixie-io/docs.px.dev/pull/261 and verified the following:
- [x] When run against a non `release/vizier/*` ref it is skipped ([build](https://github.com/pixie-io/pixie/actions/runs/5160662692))

Changelog Message:
```release-notes
Automatically create pull requests to update the pxl and python api documentation on docs.px.dev 
```

### DIFF
--- a/.github/workflows/release_update_docs_px_dev.yaml
+++ b/.github/workflows/release_update_docs_px_dev.yaml
@@ -2,17 +2,13 @@
 name: update-docs-px-dev-on-release
 on:
   push:
-    # TODO(ddelnano): Remove this in favor of vizier release events only once testing is complete
-    branches:
-    - test-create-docs-px-dev-prs-on-documentation-updates
   release:
     types: [published]
 permissions:
   contents: read
 jobs:
-  # TODO(ddelnano): Uncomment this and retest after job is verified fully
   get-dev-image:
-    if: startsWith(github.head_ref, 'release/vizier/')
+    if: startsWith(github.ref, 'release/vizier/')
     uses: ./.github/workflows/get_image.yaml
     with:
       image-base-name: "dev_image_with_extras"
--- a/.github/workflows/release_update_docs_px_dev.yaml
+++ b/.github/workflows/release_update_docs_px_dev.yaml
@@ -11,8 +11,8 @@ permissions:
   contents: read
 jobs:
   # TODO(ddelnano): Uncomment this and retest after job is verified fully
-  # if: startsWith(github.head_ref, 'release/vizier/')
   get-dev-image:
+    if: startsWith(github.head_ref, 'release/vizier/')
     uses: ./.github/workflows/get_image.yaml
     with:
       image-base-name: "dev_image_with_extras"
--- a/.github/workflows/release_update_docs_px_dev.yaml
+++ b/.github/workflows/release_update_docs_px_dev.yaml
@@ -2,18 +2,23 @@
 name: update-docs-px-dev-on-release
 on:
   push:
-    tags:
-    - release/vizier/**
+    # TODO(ddelnano): Remove this in favor of vizier release events only once testing is complete
+    branches:
+    - test-create-docs-px-dev-prs-on-documentation-updates
+  release:
+    types: [published]
 permissions:
   contents: read
 jobs:
+  # TODO(ddelnano): Uncomment this and retest after job is verified fully
+  # if: startsWith(github.head_ref, 'release/vizier/')
   get-dev-image:
     uses: ./.github/workflows/get_image.yaml
     with:
       image-base-name: "dev_image_with_extras"
   generate-docs:
     needs: get-dev-image
-    runs-on: [self-hosted, nokvm]
+    runs-on: ubuntu-latest-8-cores
     permissions:
       contents: read
     container:
@@ -25,10 +30,12 @@ jobs:
     - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3  # v3.5.0
     - name: Add pwd to git safe dir
       run: git config --global --add safe.directory `pwd`
-    - name: get bazel config
+    - name: get dev bazel config
       uses: ./.github/actions/bazelrc
       with:
         dev: 'true'
+        use_remote_exec: 'true'
+        BB_API_KEY: ${{ secrets.BB_IO_API_KEY }}
     - name: Generate documentation files
       shell: bash
       run: |
@@ -48,7 +55,6 @@ jobs:
       with:
         repository: pixie-io/docs.px.dev
         ref: main
-        fetch-depth: 0
     - name: Import GPG key
       shell: bash
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release_update_docs_px_dev.yaml
```diff
@@ -0,0 +1,106 @@
...
+      with:
+        ref: main
+    - name: Detect changes to documentation
+      id: detect-changes
+      shell: bash
+      run: |
+        STATUS="changed"
+        echo "status=${STATUS}" >> $GITHUB_OUTPUT
+
+  update-docs_px_dev:
```
JamesMBartlett, line 37: nit: just use `-` instead of `_` for consistency with other job names

.github/workflows/release_update_docs_px_dev.yaml
```diff
@@ -0,0 +1,102 @@
+---
+name: update-docs-px-dev-on-release
+on:
+  push:
```
JamesMBartlett, line 4: we should do this on release published instead of on push to the tag. See: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

.github/workflows/release_update_docs_px_dev.yaml
```diff
@@ -0,0 +1,102 @@
...
+        GIT_SSH_COMMAND: "ssh -i /tmp/ssh.key"
+      # yamllint disable rule:indentation
+      run: |
+        if [[ $(git status --porcelain=v1 | wc -l) -eq 0 ]]; then
+          echo "No updates to the documentation detected, exiting."
+          exit 0
+        fi
+
+        DATE=$(date +%Y-%m-%d)
+        PR_TITLE="Update pxl script documentation ${DATE}"
```
JamesMBartlett, line 93: nit: can you prefix the title with `[bot]`

.github/workflows/release_update_docs_px_dev.yaml
```diff
@@ -0,0 +1,102 @@
...
+permissions:
+  contents: read
+jobs:
+  get-dev-image:
+    uses: ./.github/workflows/get_image.yaml
+    with:
+      image-base-name: "dev_image_with_extras"
+  generate-docs:
+    needs: get-dev-image
+    runs-on: [self-hosted, nokvm]
```
JamesMBartlett, line 16: can you take a look at the new `build-and-test` and follow that for removing `self-hosted`? We're transitioning away from `self-hosted` so it would be nice to not add a new one.

.github/workflows/release_update_docs_px_dev.yaml
```diff
@@ -0,0 +1,102 @@
...
+  update-docs-px-dev:
+    needs: [get-dev-image, generate-docs]
+    runs-on: ubuntu-latest
+    steps:
+    - name: Clone docs.px.dev repo
+      uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3  # v3.5.0
+      with:
+        repository: pixie-io/docs.px.dev
+        ref: main
+        fetch-depth: 0
```
JamesMBartlett, line 51: nit: probably don't need fetch-depth 0 here, but also probably doesn't matter

.github/workflows/release_update_docs_px_dev.yaml
JamesMBartlett: Since this is not run on PRs, you don't need any of the authorize, or env-protect-setup stuff, just the `use_remote_exec` and `BB_API_KEY` in the bazelrc action, as well as running on `ubuntu-latest-{8,16}-cores` instead of self-hosted

.github/workflows/release_update_docs_px_dev.yaml
ddelnano: Removed.

.github/workflows/release_update_docs_px_dev.yaml
ddelnano: Sounds good. I've updated this in the latest version and went with 8 cores since this job doesn't need to run as fast.

.github/workflows/release_update_docs_px_dev.yaml
ddelnano: I wasn't sure how to scope the release publish event to just the vizier releases, which is why I went for that. My latest change should hopefully test that the `release/vizier/*` git ref must match while using the publish event.

.github/workflows/release_update_docs_px_dev.yaml
```diff
@@ -0,0 +1,105 @@
...
+    # TODO(ddelnano): Remove this in favor of vizier release events only once testing is complete
+    branches:
+    - test-create-docs-px-dev-prs-on-documentation-updates
+  release:
+    types: [published]
+permissions:
+  contents: read
+jobs:
+  # TODO(ddelnano): Uncomment this and retest after job is verified fully
+  # if: startsWith(github.head_ref, 'release/vizier/')
```
JamesMBartlett, line 14: the `if` should be inside the job

ddelnano: Yea, I noticed that when I was prepping the final change for this PR. However, I still need to test that the job's previous functionality works after the most recent refactor.

ddelnano: This will be tested in my most recent update. I moved it into the `get-dev-image` job since all of the other ones depend on it.

.github/workflows/release_update_docs_px_dev.yaml
```diff
@@ -0,0 +1,105 @@
...
+    branches:
+    - test-create-docs-px-dev-prs-on-documentation-updates
+  release:
+    types: [published]
+permissions:
+  contents: read
+jobs:
+  # TODO(ddelnano): Uncomment this and retest after job is verified fully
+  get-dev-image:
+    if: startsWith(github.head_ref, 'release/vizier/')
```
JamesMBartlett, line 15: you should use `github.ref` because `head_ref` only exists on `pull_request` and `pull_request_target` events

.github/workflows/release_update_docs_px_dev.yaml
```diff
@@ -0,0 +1,101 @@
+---
+name: update-docs-px-dev-on-release
+on:
+  push:
```
JamesMBartlett, line 4: should remove this `push`

.github/workflows/release_update_docs_px_dev.yaml
ddelnano: Updated

.github/workflows/release_update_docs_px_dev.yaml
ddelnano: ah, I missed removing that. Updated in latest revision.

ddelnano:
@JamesMBartlett this should be ready for a final review. While I made a minor fix ([0dd627b](https://github.com/pixie-io/pixie/pull/1408/commits/0dd627bd58a5c91437b31f2702a4acf423088b6b)) after the most recent round of testing, I think it should be functional in its current state.

JamesMBartlett (approved):

vihangm (approved):

PR #1424, opened 2023-06-03
Split container_images into separate headers per image

### LINKED ISSUE #729: Split dependency graph of `src/stirling/ ... /:container_images` to improve BPF test builds
The BPF end to end tests rely on [docker images](https://github.com/pixie-io/pixie/blob/06d3de4120e4e4ba6091c8e77c96f3be227e2066/src/stirling/source_connectors/socket_tracer/testing/container_images.h) in order to perform their testing. All of these tests depend on the `src/stirling/source_connectors/socket_tracer/testing:container_images` target which defines all the container images that are available. The issue with this is that in order to run the [redis_trace_bpf_test](https://github.com/pixie-io/pixie/blob/06d3de4120e4e4ba6091c8e77c96f3be227e2066/src/stirling/source_connectors/socket_tracer/BUILD.bazel#L397), you have to build the images for the unrelated bpf tests.

We should reconsider this decision to speed up builds for a smaller subset of the tests.

vihangm:
Summary: The massive container_images.h header and it's data dependency
inclusion meant that anything that depended on container_image got
~8GB worth of runfiles even if they didn't need all the containers.

This split up means that consumers can include just the containers
they need and causes less data to be transferred to the QEMU fs.

This also makes it so that changing a single image doesn't trigger
builds for all the tests.

Relevant Issues: Fixes #729

Type of change: /kind cleanup

Test Plan: Everything builds locally and all the tests pass (including
BPF tests under QEMU).


src/stirling/utils/detect_application_bpf_test.cc
```diff
@@ -31,32 +31,10 @@ namespace stirling {
...
-  ContainerRunner node_server(px::testing::BazelRunfilePath(kNode15_0ImageTar), "node_server", "");
-  ASSERT_OK_AND_ASSIGN(std::string output, node_server.Run(std::chrono::seconds{60}));
-
-  ProcParser proc_parser;
-
-  ASSERT_OK_AND_ASSIGN(const auto proc_exe_path, proc_parser.GetExePath(node_server.process_pid()));
-  ASSERT_OK_AND_THAT(px::Exec(absl::StrCat(proc_exe_path.string(), " --version")),
-                     StrEq("v15.0.1\n"));
-}
-
```
vihangm, line 52: this was removed since 1) the test is disabled and 2) the image referenced by the test no longer exists

src/stirling/source_connectors/socket_tracer/testing/container_images/BUILD.bazel
```diff
@@ -0,0 +1,368 @@
...
+        "nginx_worker_pid.h",
+    ],
+    data = [
+        "//src/stirling/source_connectors/socket_tracer/testing/containers:nginx_openssl_1_1_1_image.tar",
+    ],
+    deps = ["//src/common/testing/test_utils:cc_library"],
+)
+
+pl_cc_test_library(
+    name = "nginx_open_ssl_3_0_7_container",
```
ddelnano, line 188: We have a precedent for naming openssl as one word. I would prefer we keep that consistency.

src/stirling/source_connectors/socket_tracer/kafka_trace_bpf_test.cc
```diff
@@ -28,7 +28,8 @@
 #include "src/common/testing/test_utils/container_runner.h"
 #include "src/common/testing/testing.h"
 #include "src/stirling/source_connectors/socket_tracer/protocols/kafka/common/types.h"
-#include "src/stirling/source_connectors/socket_tracer/testing/container_images.h"
+#include "src/stirling/source_connectors/socket_tracer/testing/container_images/kafka_container.h"
+#include "src/stirling/source_connectors/socket_tracer/testing/container_images/zookeeper_container.h"
```
ddelnano, line 32: This test is [disabled](https://github.com/pixie-io/pixie/blob/f4f4ec2d16cc66306ac6d35c0bf6a34ab528aab3/src/stirling/source_connectors/socket_tracer/BUILD.bazel#L416) under qemu. Can you force it to run locally to verify that even though it fails the containers run for the test?

src/stirling/source_connectors/socket_tracer/testing/container_images/BUILD.bazel
vihangm: Done.
I also renamed the nginx_open_ssl_3_0_7 container/headers to `nginx_openssl_3_0_8` since when I inspected the container, it was running openssl 3.0.8

```
$ docker run --rm -it --entrypoint /bin/sh gcr.io/pixie-oss/pixie-dev-public/docker-deps/library/nginx@sha256:3eb380b81387e9f2a49cb6e5e18db016e33d62c37ea0e9be2339e9f0b3e26170
/ # strings /lib/libcrypto.so.3 | grep -o "OpenSSL [[:digit:]][^ ]*"
OpenSSL 3.0.8
```

src/stirling/source_connectors/socket_tracer/kafka_trace_bpf_test.cc
vihangm: Ran with the `no_qemu` tag removed and the containers spin up properly even though some of the assertions fail.

ddelnano (approved):

ddelnano (approved):

aimichelle (approved):

Issue #655, opened 2022-11-21
User deletion and Organization selection


sgewirtz1:
**Is your feature request related to a problem? Please describe.**
When I have a collection of resources that I'm managing and want to add users to my organization for those that sign up in a timely fashion with the invite link it works fine, but for users that sign up either directly or with the incorrect or an old link the system creates their user account with their own default organization and from then on it's impossible for them to delete their user account or to join a new/different organization.

**Describe the solution you'd like**
1. A user administration capability to delete the user's account from the system so they can have a fresh start
2. A user organization management selector so that users can be part of more than one organization

**Describe alternatives you've considered**
As an alternative to 2 providing the user the ability to either leave their organization to move to another or to delete the user's organization so they can join another
**Additional context**
Slack link: https://pixie-community.slack.com/archives/CQ63KEVFY/p1668936141076099
Question:
is there a way to add a user to my org if they've already logged in to their own org? alternatively can we somehow delete his org so he can join our org?

Answer:
currently there isn’t a way to switch a user’s org. The best way to handle this right now is to delete their Pixie account and have them sign up again. Can you have the user email me their account information and ask to have their account deleted and we can look at it on Monday?

sgewirtz1:
Depending on the scope and feedback I could attempt a PR for this

htroisi:
For this specific use case, I think the simplest solution would be to add the ability for a user to delete their own account. 

According to @aimichelle we have the capability to remove other users from your org. However, if you’re the last user in the org (i.e. you accidentally created your own org and you're the only person in it), we don’t allow you to. So the improvement here would be to implement the following: if you are the last user in the org, and you try to remove yourself, it will tell you that the org will also be deleted.

sgewirtz1:
Sounds appropriate, thanks!

VishwasSomasekhariah:
Any Updates on this? Or workaround for now, until the issue is addressed.

htroisi:
@VishwasSomasekhariah the current workaround is to have your account deleted by reaching out to info(at)pixielabs.ai.

PR #1439, opened 2023-06-06
Add user deletion to API service

aimichelle:
Summary: We are adding user deletion to the API. #1434 added user deletion to the internal profile service. This PR adds it to the external-facing gRPC API. A followup diff will later add it to gql.

Relevant Issues: #655

Type of change: /kind feature

Test Plan: unit tests


JamesMBartlett:
need to rebase

src/api/proto/cloudpb/cloudapi.proto
```diff
@@ -76,6 +77,16 @@ message UpdateUserRequest {
   google.protobuf.BoolValue is_approved = 3;
 }
 
+// DeleteUserRequest is a request to delete a user. This request must be made with the user's
+// credentials.
```
JamesMBartlett, line 81: maybe add a comment that it deletes the org if the user is the last one in the org

JamesMBartlett (approved):

Issue #1352, opened 2023-05-18
Expose Stirling parameters to Carnot.


etep:
**Expose Stirling parameters to Carnot.**
Certain UDFs or UDAs may need access to parameters in Stirling. For example, the UDA to convert stack trace profiling information to pprof protobuf format needs to know the underlying stack trace sampling period. At this time, there is no obvious way to expose access this information in the UDA. Using `DECLARE_uint32` results in a linking error (and fixing this linking error likely requires carnot to depend on stirling).

Some options to consider:
1. Add the same command line flag to carnot (may require updates to how we set the command line flags when deploying).
2. Implement a "stirling parameters" table that carnot can access.

etep:
Closed by #1448.

PR #1448, opened 2023-06-07
Extend `AgentInfo` with `AgentParameters` to expose the profiler sampling period in the metadata service.

etep:
Summary: Motivated by the need to include profiler sampling period into the pprof format, we extend `AgentInfo` to include `AgentParameters` such that the profiler sampling period is available in the metadata service. We add a new UDTF `GetProfilerSamplingPeriodMS` (similar to agent status) that creates a table with columns `asid` and `profiler_sampling_period_ms`. This table can be joined with a stack traces table and the profiler sampling period column can be input to the pprof UDA.

Relevant Issues: https://github.com/pixie-io/pixie/issues/1352

Type of change: /kind feature

Test Plan: Tested end-to-end locally with a prototype. PR https://github.com/pixie-io/pixie/pull/1336 will include an updated unit test.


src/vizier/services/shared/agentpb/agent.proto
```diff
@@ -30,12 +30,17 @@ message AgentCapabilities {
   bool collects_data = 1;
 }
 
+message AgentParameters {
+  int32 profiler_stack_trace_sample_period_ms = 1;
```
aimichelle, line 34: Nit: We probably want `[ (gogoproto.customname) = "ProfilerStackTraceSamplePeriodMS" ];`

src/vizier/services/agent/shared/manager/registration.cc
```diff
@@ -80,6 +80,7 @@ Status RegistrationHandler::DispatchRegistration() {
   host_info->set_pod_name(agent_info()->pod_name);
   host_info->set_host_ip(agent_info()->host_ip);
   *req_info->mutable_capabilities() = agent_info()->capabilities;
+  *req_info->mutable_parameters() = agent_info()->parameters;
```
ddelnano, line 83: I take it that kelvin will register itself also since it seems this manager code is shared?

ddelnano (approved):

src/vizier/services/agent/shared/manager/registration.cc
aimichelle: Yup, both Kelvin and PEMs use this registration code to register with the metadata service. And the metadata service distinguishes between the Kelvin/PEM using the capabilities.

aimichelle (approved):

Issue #597, opened 2022-09-19
stirling support for boringcrypto as used by the 'boringcrypto' version of Golang


kay-ge:
It would be great if stirling were able to decode TLS flows from the 'boringcrypto' versions of Golang.

Stirling would need to be modified to have uprobes for this version of Golang: https://go.googlesource.com/go/+/dev.boringcrypto

This version of Golang is used in order to have FIPS compliant Go programs.  We at VMware use it extensively.

Alternatives:
Don't use the boringcrypto version of Golang.  This is not an option in cases where one wants to claim FIPS compliance for their Go program.

ddelnano:
@kay-ge while working on #692, I realized that our existing Go TLS tracing already covers Go applications using boringcrypto (as long as DWARF debug information is available).

Can you confirm if your boringcrypto applications had missing traffic or was it because our documentation only states Go TLS and OpenSSL support?

I was able to update Pixie's go tls tracing tests to use binaries opted into `GOEXPERIMENT=boringcrypto` and the tests pass ([commit](https://github.com/pixie-io/pixie/commit/17bc2dcf4d7ca398ab374b88bd193412da7a54ae)). This is because the functions we hook into (`crypto/tls.(*Conn) Write` and `Read`) are triggered for both the standard crypto and boringcrypto case. I'm going to add a boringcrypto acceptance test to the Pixie repo so we are regularly exercising this.

<details>
<summary>Test output</summary>

```
# Verify that boringcrypto is active

ddelnano@vigenere:~/code/pixie (main) $ bazel build //src/stirling/testing/demo_apps/go_grpc_tls_pl/server:golang_1_19_grpc_tls_server_binary
ddelnano@vigenere:~/code/pixie (main) $ bazel build //src/stirling/testing/demo_apps/go_grpc_tls_pl/server:golang_1_20_grpc_tls_server_binary

ddelnano@vigenere:~/code/pixie (main) $ ~/go/bin/goversion -crypto bazel-bin/src/stirling/testing/demo_apps/go_grpc_tls_pl/server/golang_1_19_grpc_tls_server_binary
bazel-bin/src/stirling/testing/demo_apps/go_grpc_tls_pl/server/golang_1_19_grpc_tls_server_binary go1.19.9 X:boringcrypto (boring crypto)

[...]

PR #1449, opened 2023-06-07
Refactor Go SDK label templating to support future boringcrypto SDK

ddelnano:
Summary: Refactor Go SDK label templating to support future boringcrypto SDK

This PR adds the scaffolding needed to add a boringcrypto go SDK. This SDK will be used in a future change to add TLS tracing tests for binaries using boringcrypto, which addresses #597. It wasn't known that boringcrypto was supported at the time, but we should still validate that it is functional.

`rules_go` does not support go SDKs that use the same version with different `GOEXPERIMENT`s enabled (will be following up to create a GitHub issue on the project). This is an issue because boringcrypto is enabled by setting `GOEXPERIMENT=boringcrypto` as mentioned [here](https://go.googlesource.com/go/+/refs/heads/dev.boringcrypto/README.boringcrypto.md). Until `rules_go` supports this, the proposed plan is to maintain a previous patch version of our latest supported version of go as the "boringcrypto go SDK". The description below should explain the process:

```
# rules_go doesn't support using multiple SDKs with the same version and differing
# GOEXPERIMENTs. Until this is addressed, go_sdk_boringcrypto is meant to be 1 bug fix
# version behind our latest go release. In the event our primary toolchain is upgraded
# to the first release of a new major version (i.e. 1.20.0) an rc suffixed build should
# be used for go_sdk_boringcrypto (1.20rcX) until the first minor release is available (1.20.1).
```

Relevant Issues: #597 #692

Test Plan: Existing tests pass and verified this supports the boringcrypto tests on a branch with the full set of changes

### DIFF
--- a/bazel/pl_build_system.bzl
+++ b/bazel/pl_build_system.bzl
@@ -32,7 +32,7 @@ pl_all_supported_go_sdk_versions = pl_supported_go_sdk_versions + pl_boringcrypt
 def pl_go_sdk_version_template_to_label(tpl, version):
     # If version matches sdk configured to use boringcrypto
     # the label name should not contain the sdk version string
-    if version == pl_all_supported_go_sdk_versions[-1]:
+    if version in pl_boringcrypto_go_sdk:
         return tpl % "boringcrypto"
     return tpl % version.replace(".", "_")
 


bazel/pl_build_system.bzl
```diff
@@ -23,8 +23,19 @@ load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
 load("@rules_python//python:defs.bzl", "py_test")
 load("//bazel:toolchain_transitions.bzl", "qemu_interactive_runner")
 
+pl_boringcrypto_go_sdk = ["1.20.4"]
 pl_supported_go_sdk_versions = ["1.16", "1.17", "1.18", "1.19", "1.20"]
 
+# The last version in this list corresponds to the boringcrypto go sdk version.
+pl_all_supported_go_sdk_versions = pl_supported_go_sdk_versions + pl_boringcrypto_go_sdk
```
ddelnano, line 30: In the next change I will replace `pl_supported_go_sdk_versions` with `pl_all_supported_go_sdk_versions` in order to create the container targets used in the TLS tracing tests.

bazel/pl_build_system.bzl
```diff
@@ -23,8 +23,19 @@ load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
...
+
+def pl_go_sdk_version_template_to_label(tpl, version):
+    # If version matches sdk configured to use boringcrypto
+    # the label name should not contain the sdk version string
+    if version == pl_all_supported_go_sdk_versions[-1]:
+        return tpl % "boringcrypto"
+    return tpl % version.replace(".", "_")
```
ddelnano, line 37: The goal of this function was to make provide stable names for boringcrypto targets. Whenever we upgrade our latest version of go and the boringcrypto version, I wanted to avoid having to change version strings across all the boringcrypto targets.

ddelnano:
@vihangm @JamesMBartlett could you give this a review before the Stirling team review? Since this is solely a build change, I wanted to make sure this was in line with what we previously discussed.

vihangm (commented):
Can you pull out the version bump? When bumping versions we also need to update it in chef and push new images.
Also go typically releases patch fixes for both the current version and the previous version, so we should also bump to `1.19.10` when we bump to `1.20.5`

ddelnano:
@vihangm sounds good. I've split out that change in #1451 and included bumping go 1.19.

ddelnano:
@vihangm this has been rebased now that #1451 is merged.

bazel/pl_build_system.bzl
```diff
@@ -23,8 +23,19 @@ load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_test")
...
 pl_supported_go_sdk_versions = ["1.16", "1.17", "1.18", "1.19", "1.20"]
 
+# The last version in this list corresponds to the boringcrypto go sdk version.
+pl_all_supported_go_sdk_versions = pl_supported_go_sdk_versions + pl_boringcrypto_go_sdk
+
+def pl_go_sdk_version_template_to_label(tpl, version):
+    # If version matches sdk configured to use boringcrypto
+    # the label name should not contain the sdk version string
+    if version == pl_all_supported_go_sdk_versions[-1]:
+        return tpl % "boringcrypto"
```
vihangm, line 36: nit: I think we should compare against `pl_boringcrypto_go_sdk` instead of `pl_all_supported_go_sdk_versions[-1]`

bazel/pl_build_system.bzl
vihangm: > The goal of this function was to make provide stable names for boringcrypto targets. Whenever we upgrade our latest version of go and the boringcrypto version, I wanted to avoid having to change version strings across all the boringcrypto targets.

This seems fine but also "Whenever we upgrade our latest version of go" we do have to change version strings across all the tests that consume version specific go binaries. So IMO having to change the boringcrypto target deps at the same time seems fine (though I thing having to rev at patch versions could get annoying, but if you dropped the patch version but made it an versioned string, it seems ok)

vihangm (approved):

bazel/pl_build_system.bzl
ddelnano: I think I'd prefer to keep it this way for now. If we need to test against boringcrypto enabled binaries across multiple go versions, then I think including the major and minor (similar to how the existing labels work) makes sense.

PR #1455, opened 2023-06-08
Add more debug info to failed UI builds and propagate tags

vihangm:
Summary: Propagating tags makes the build respect `no-remote-exec`.
The additional debug info can be useful if the build fails.

Type of change: /kind cleanup

Test Plan: Check the build results on this PR.


aimichelle (approved):

bazel/ui.bzl
```diff
@@ -76,15 +78,21 @@ def _pl_webpack_library_impl(ctx):
         all_files.append(ctx.info_file)
         all_files.append(ctx.version_file)
 
-    cmd = env_cmds + ui_shared_cmds_start + [
+    cp_cmds = ["cp -aL --parents {} $TMPPATH".format(file.path) for file in all_files]
```
JamesMBartlett, line 81: do we not need to copy the files from the deps and licenses attributes?

vihangm: those are handled in line 85, 86 since those are outputs and not in the src tree and need to be untar-ed and moved etc.

JamesMBartlett (approved):

PR #1456, opened 2023-06-08
Fetch license file from GCS instead of bundling it into cloud

vihangm:
Summary: We currently bundle the license file into the cloud assets.
However we also publish the file on GCS. Fetch it from GCS instead
of bundling it to make the UI build a bit less brittle.

Type of change: /kind cleanup

Test Plan: Deployed a skaffold cloud and tested that the credits page
still works (albeit with a slight delay now).


src/ui/src/pages/credits/credits.tsx
```diff
@@ -21,11 +21,14 @@ import * as React from 'react';
...
 import { createStyles, makeStyles } from '@mui/styles';
+import Axios from 'axios';
 
 import { buildClass, scrollbarStyles, Footer } from 'app/components';
+import { PIXIE_CLOUD_VERSION } from 'app/utils/env';
 import { WithChildren } from 'app/utils/react-boilerplate';
 import { Copyright } from 'configurable/copyright';
-import licenseJson from 'configurable/licenses.json';
+
+const licenseURL = `https://storage.googleapis.com/pixie-dev-public/oss-licenses/${PIXIE_CLOUD_VERSION.tag}.json`;
```
NickLanam, line 31: This will fail for airgapped users, leaving them without license information. Is that acceptable?

src/ui/src/pages/credits/credits.tsx
```diff
@@ -144,11 +147,16 @@ const LicenseEntryRow = React.memo<LicenseEntry>(({ name, url, licenseText }) =>
 LicenseEntryRow.displayName = 'LicenseEntryRow';
 
 // eslint-disable-next-line react-memo/require-memo
-const Credits: React.FC<{ licenses: LicenseEntry[] }> = ({ licenses }) => {
+const Credits: React.FC = () => {
   const classes = useStyles();
+  const [licenses, setLicenses] = React.useState<LicenseEntry[]>(null);
+
+  React.useEffect(() => {
+    Axios.get(licenseURL).then((response) => { setLicenses(response.data); });
```
NickLanam, line 155: Need a `.catch` on this.

src/ui/src/pages/credits/credits.tsx
vihangm: Yeah I thought about this issue already ... I have a different approach I'm going to try, hold on.

PR #1458, opened 2023-06-08
Don't use configurables for licenses

vihangm:
Summary: This ensures that all cloud builds have a working
credits page with a somewhat graceful though suboptimal
fallback.

Type of change: /kind cleanup

Test Plan: Check the credits page for an OSS cloud build.


NickLanam (approved):
I like this approach much better!

JamesMBartlett (approved):

PR #1460, opened 2023-06-08
[build] Disable tcmalloc for GCC build

JamesMBartlett:
Summary: tcmalloc was causing strange linker errors with CGO builds when linker with gold. Not really sure what's going on, since lld can link these things without issue. We should probably investigate this more, but for now this is a stop-gap that will unblock the coverage build. Since we don't ever deploy GCC built binaries, removing tcmalloc shouldn't matter too much for now.

Type of change: /kind cleanup

Test Plan: Ran `bazel build //src/vizier/services/query_broker:query_broker --config=gcc` without error, this was failing on main before this PR.


vihangm (approved):
Can you leave a TODO and link to and issue to investigate later?

PR #1468, opened 2023-06-12
Fix an update issue caused by React being out of scope at a bad time

NickLanam:
Summary: Moving PixieAPIManager out of React's scope was a risky move.
As it turns out, PixieAPIContext wasn't catching one of the updates during the embed authentication procedure because of this.
By implementing an unholy hack to tell React when this happens, the `authorized` network call still fires with a bearer token.

Type of change: /kind bugfix

Test Plan: Try to embed Pixie using `embedPixieToken` to authenticate. Before, it would give up trying right before the `postMessage` comes through. After, it should try once more as soon as the auth token is provided to Pixie.

### DIFF
--- a/src/ui/src/api/api-context.tsx
+++ b/src/ui/src/api/api-context.tsx
@@ -35,7 +35,7 @@ export type PixieAPIContextProviderProps = WithChildren<PixieAPIClientOptions>;
 
 declare global {
   interface Window {
-    apiContextUpdates?: SetStateFunc<number>;
+    setApiContextUpdatesFromOutsideReact?: SetStateFunc<number>;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/ui/src/api/api-context.tsx
```diff
@@ -21,27 +21,50 @@ import * as React from 'react';
...
+  interface Window {
+    setApiContextUpdatesFromOutsideReact?: SetStateFunc<number>;
+  }
+}
+
 export const PixieAPIContextProvider: React.FC<PixieAPIContextProviderProps> = React.memo(({ children, ...opts }) => {
   const { authToken } = React.useContext(AuthContext);
 
+  // PixieAPIManager exists outside of React's scope. If it replaces its instance, we'll never know.
+  // By putting a setState function somwhere that PixieAPIManager can reach, we can force the update from there.
```
aimichelle, line 46: Typo: somewhere

aimichelle (approved):

PR #1472, opened 2023-06-13
Cleanup privy deps

vihangm:
Summary: This splits of the direct deps for privy from the
indirect deps making it easier to understand and test
dep upgrades that address CVEs.

Type of change: /kind cleanup

Test Plan: bazel build and tests inside privy work. datagen works.


JamesMBartlett (approved):

JamesMBartlett (changes requested):
Should fix the protobuf version's cve

PR #1482, opened 2023-06-14
Use ghcr.io for build images

vihangm:
Summary: This switches to using ghcr.io for our build images.
We will continue to maintain a copy on gcr.io for now and also
sync historic images to ghcr.io.

Type of change: /kind cleanup

Test Plan: All github actions on this PR should work.


tools/chef/image_sync.yaml
```diff
@@ -0,0 +1,36 @@
+---
+version: 1
+creds:
+- registry: gcr.io
+- registry: ghcr.io
+  user: "{{env \"GHCR_USER\"}}"
```
JamesMBartlett, line 6: Does regsync get these from the docker config?

JamesMBartlett (approved):

tools/chef/image_sync.yaml
vihangm: If you are already auth-ed but these are not set, it still works.

Perhaps I should remove them from this config but I imagine it will exist for other sync configs that might be run via gh-actions.

aimichelle (approved):

PR #1485, opened 2023-06-14
[cloud] Use two services for NATS

JamesMBartlett:
Summary: Since NATS uses the service dns to join the cluster, we're required to set `publishNotReadyAddresses` on the service. This was causing issues where other pods, eg vzmgr, were connecting to a not-ready NATS pod. This would occur, for example, during cluster upgrades where the NATS pods would be moved to new nodes. This was causing vzmgr and other services to crash. This PR renames the current NATS service to `pl-nats-internal` and creates a new service for other services to connect to called `pl-nats`. The new service doesn't have `publishNotReadyAddresses` set, so the above issue should be avoided. Note that we'll need to update all the cluster certs after this change.

Type of change: /kind cleanup

Test Plan: Tested by deleting NATS pods and seeing that they were no longer in the routes for the `pl-nats` service, but still in the routes for the `pl-nats-internal` service. Also tested a k8s upgrade on testing cloud, and didn't see any crashes from `vzmgr` or others, but this could be due to the low load on testing.

### DIFF
--- a/k8s/cloud_deps/base/nats/statefulset.yaml
+++ b/k8s/cloud_deps/base/nats/statefulset.yaml
@@ -43,11 +43,11 @@ data:
       name: pl-nats
       port: 6222
       routes [
-        nats://pl-nats-0.pl-nats:6222
-        nats://pl-nats-1.pl-nats:6222
-        nats://pl-nats-2.pl-nats:6222
-        nats://pl-nats-3.pl-nats:6222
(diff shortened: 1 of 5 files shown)


JamesMBartlett:
I don't have time to test this out and go through the long process of changing the certs safely, so going to close for now. Someone else is welcome to pick this up if nats stability becomes a concern.

PR #1516, opened 2023-06-16
Update build an CI tooling to allow arbitrary container image registries

vihangm:
Summary: This updates our config flags and ci scripts to allow our
various container artifacts to be pushed to arbitrary image registies
via a config flag.

This doesn't yet support arbitrary registries since most registries don't
allow you to treat image paths like nested dirs, so
`$(IMAGE_PREFIX)/cloud/plugin/load_db` for example won't work.

However this is a stepping stone along the way to supporting other
registries without changing where we currently push images.

Type of change: /kind cleanup

Test Plan: Ran builds locally to test that image paths were sane.
More testing will require creating RCs and deploying those.

### DIFF
--- a/bazel/images.bzl
+++ b/bazel/images.bzl
@@ -16,16 +16,12 @@
 
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 
-PROPRIETARY_PREFIX = "gcr.io/pl-dev-infra/"
-PUBLIC_PREFIX = "gcr.io/pixie-oss/pixie-prod/"
-DEV_PREFIX = "gcr.io/pixie-oss/pixie-dev/"
-
-def image_replacements(image_map, existing_prefix, new_prefix):
+def image_replacements(image_map):
     replacements = {}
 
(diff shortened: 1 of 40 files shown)


src/vizier/services/cloud_connector/bridge/server.go
```diff
@@ -81,7 +81,7 @@ spec:
       serviceAccountName: pl-updater-service-account
       containers:
       - name: updater
-        image: gcr.io/pixie-oss/pixie-dev/vizier/vizier_updater_image:__VIZIER_UPDATER_IMAGE_TAG__
+        image: gcr.io/pixie-oss/pixie-prod/vizier-vizier_updater_image:__VIZIER_UPDATER_IMAGE_TAG__
```
aimichelle, line 84: I think you'll also need to update `ParseJobYAML` in `cloud_connector/bridge/vzinfo.go`, since it tries to do some image path replacement there and assumes you have `/vizier` in the image path.

aimichelle (approved):

PR #1517, opened 2023-06-16
Remove one off container push targets

vihangm:
Summary: We build and push most of our distributed images using
container bundles. We use skaffold to test local dev builds. So
these one off push targets are unused and increase maintenance
burden. Instead just remove them.

Type of change: /kind cleanup

Test Plan: All existing builds and deploys work.


JamesMBartlett (approved):
:100: :100:

aimichelle (approved):

PR #1521, opened 2023-06-16
Remove some more container_push targets

vihangm:
Summary: The container push targets for the e2e tests rely on a hardcoded
registry that folks might not have access to push to. Instead we can
switch to relying on skaffold to build the images, replace them in the k8s
yamls and deploy them as needed.

While this makes the yamls for these k8s tests not deployable via just
kustomize (since the referenced images don't exist), IMO it cleans things
up by relying on just having one way to run the tests (via skaffold).

Type of change: /kind cleanup

Test Plan: Skaffolded an e2e test and ensured that it was built
and deployed to my cluster.


JamesMBartlett (approved):
The yamls look a little weird now, but I think its fine

PR #1530, opened 2023-06-20
Remove legacy TLS tracing feature toggle and transitional code

ddelnano:
Summary: Remove legacy TLS tracing feature toggle and transitional code

The new style of TLS tracing has been rolled out since April 11th (vizier release v0.12.19). We believe that it is performing well and it has already allowed Pixie's TLS tracing to cover more libraries (OpenSSL v3) with BoringSSL coming soon.

This includes changes from #1518 and must be rebased once that is merged.

Relevant Issues: #692

Type of change: /kind cleanup

Test Plan: Existing test coverage


src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
```diff
@@ -296,23 +297,30 @@ static constexpr const auto kLibSSLMatchers = MakeArray<SSLLibMatcher>({
...
         // (e.g. libpython3.10.so.0.1).
         .libssl = kLibPython,
         .libcrypto = kLibPython,
         .search_type = HostPathForPIDPathSearchType::kSearchTypeContains,
+        .probe_suffix = "_syscall_fd_access",
     },
+    // non BIO native TLS applications cannot be probed by accessing the socket fd
+    // within the underlying syscall. Specifying an empty probe_suffix results in TLS
+    // tracing via the legacy mechanism (finding socket fd with user space memory offsets).
     SSLLibMatcher{
```
etep, line 319: The concept of a `probe_suffix` is not self documenting and although there is a comment here (enough of a clue to go on) there may be a better way to directly describe what is going on here. Possibly something like this:
1. An struct member (enum perhaps) that identifies the SSL technique (e.g. nested, legacy).
2. one of the following:
... a. Function that returns the probe fn. name based on the relevant parameters.
... b. A map or array that stores the probe fn. suffix based on the enum.

ddelnano: I've refactored this to use 1). Let me know what you think.

etep (approved):

src/cloud/config_manager/controllers/vizier_feature_flags.go
```diff
@@ -42,11 +42,6 @@ var availableFeatureFlags = []*featureFlag{
 		VizierFlagName:  "PL_PROFILER_JAVA_SYMBOLS",
 		DefaultValue:    true,
 	},
-	{
-		FeatureFlagName: "access-tls-socket-fd-via-syscall",
-		VizierFlagName:  "PL_ACCESS_TLS_SOCKET_FD_VIA_SYSCALL",
-		DefaultValue:    false,
-	},
```
ddelnano, line 49: cc @aimichelle FYI that we should keep this flag set to `true` on LaunchDarkly, but we will be able to clean it up once we believe all clusters have upgraded.

vihangm (approved):

src/cloud/config_manager/controllers/vizier_feature_flags.go
aimichelle: Sounds good! I think there are always a few stragglers who somehow get stuck on older Vizier versions, so we can leave that on. There's no cost in having the flag around, I'll just leave a note on LaunchDarkly with the context.

PR #1532, opened 2023-06-20
Update non operator upgrade to be a bit more robust to image paths

vihangm:
Summary: This allows us to make changes to the image path for the updater
more easily.

Type of change: /kind cleanup

Test Plan: Will test with a RC build.

### DIFF
--- a/src/vizier/services/cloud_connector/bridge/vzinfo.go
+++ b/src/vizier/services/cloud_connector/bridge/vzinfo.go
@@ -512,12 +512,9 @@ func (v *K8sVizierInfo) ParseJobYAML(yamlStr string, imageTag map[string]string,
 			imgPath := imgTag[0]
 
 			tagVers := semver.MustParse(val)
-
-			repoPath := publicImageRepo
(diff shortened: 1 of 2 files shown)


src/vizier/services/cloud_connector/bridge/vzinfo.go
```diff
@@ -512,12 +512,9 @@ func (v *K8sVizierInfo) ParseJobYAML(yamlStr string, imageTag map[string]string,
 			imgPath := imgTag[0]
 
 			tagVers := semver.MustParse(val)
-
-			repoPath := publicImageRepo
-			if len(tagVers.Pre) > 0 {
-				repoPath = privateImageRepo
+			if len(tagVers.Pre) > 0 && strings.HasPrefix(imgPath, publicImageRepo) {
+				imgPath = strings.Replace(imgPath, publicImageRepo, privateImageRepo, 1)
```
JamesMBartlett, line 516: I don't think RCs even get pushed to the private repo anymore. So maybe we can just do away with this?

aimichelle (approved):

PR #1552, opened 2023-06-21
Add github workflow to sync released container images

vihangm:
Summary: This cleans up the regbot script to make it easier to use for
other images and also sets up a github action to sync released images
when a release occurs.

Type of change: /kind cleanup

Test Plan: Triggered the action via a branch push to test
https://github.com/pixie-io/pixie/actions/runs/5338969676/jobs/9677065378


scripts/regclient/utils.lua
```diff
@@ -0,0 +1,36 @@
...
+  return string.match(v, "^(%d+)%.(%d+)%.(%d+)(.-)$")
+end
+
+function utils.combine(a, b)
+  -- create a path that looks like a directory, this is supported by gcr
+  local dirPath = a .. "/" .. b
+  if string.sub(dirPath, 1, #gcr) == gcr then
+    return dirPath
+  end
+
```
JamesMBartlett, line 15: why can't the rest of this just be `return a .. "/" .. string.gsub(b, "/", "-")`

JamesMBartlett (approved):

scripts/regclient/utils.lua
vihangm: because registries allow namespaces, this will be used later for copying from say `docker.io` + `curlimages/curl` which should map to `docker.io/curlimages/curl` not `docker.io/curlimages-curl`

aimichelle (approved):

PR #1553, opened 2023-06-21
[dev artifacts] Create releases for development artifacts.

JamesMBartlett:
Summary: We need to store our development artifacts somewhere other than gcs. This PR creates releases on the `pixie-io/dev-artifacts` repo, storing the artifacts in these releases. This prevents having random releases on the main repo, but still gives us a stable location for the dev artifacts.

Type of change: /kind cleanup

Test Plan: Ran `bazel clean --expunge && bazel build //...`. Also relying existing tests.

### DIFF
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -283,79 +283,118 @@ REPOSITORY_LOCATIONS = dict(
     ),
     com_llvm_clang_15 = dict(
         sha256 = "a275cde426a790cdf6eaecc2a33bac320ee61ceb4f5e76d2c73d3c555fdcedfc",
-        urls = ["https://storage.googleapis.com/pixie-dev-public/clang/15.0-pl11/clang-min-15.0-pl11.tar.gz"],
+        urls = [
+            "https://github.com/pixie-io/dev-artifacts/releases/download/clang%2F15.0-pl11/clang-min-15.0-pl11.tar.gz",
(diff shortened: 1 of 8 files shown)


bazel/repository_locations.bzl
```diff
@@ -283,79 +283,118 @@ REPOSITORY_LOCATIONS = dict(
     ),
     com_llvm_clang_15 = dict(
         sha256 = "a275cde426a790cdf6eaecc2a33bac320ee61ceb4f5e76d2c73d3c555fdcedfc",
-        urls = ["https://storage.googleapis.com/pixie-dev-public/clang/15.0-pl11/clang-min-15.0-pl11.tar.gz"],
+        urls = [
+            "https://github.com/pixie-io/dev-artifacts/releases/download/clang/15.0-pl11/clang-min-15.0-pl11.tar.gz",
+            "https://storage.googleapis.com/pixie-dev-public/clang/15.0-pl11/clang-min-15.0-pl11.tar.gz",
```
vihangm, line 288: shouldn't these be `clang%2F15.0-pl11`?
i.e. escaped `/` in the release name?

vihangm (commented):
I think the `/` in the release tag should be escaped.

bazel/repository_locations.bzl
JamesMBartlett: Both links work, and imo its nicer to not worry about the escape, but if you think they'll support the escaped link longer than the unescaped one, I can change it.

JamesMBartlett: Will escape

vihangm (approved):

PR #1556, opened 2023-06-21
Update install.sh to point to GH for CLI release artifacts

aimichelle:
Summary: This script used to point to GCS for downloading artifacts for the CLI. This updates the script to instead use the artifacts from our GH releases.
Since we no longer tag/store the CLI under latest, this pulls from our README to get the latest version.
Also removed `px auth login`, since this can be confusing for users who have not yet input their `PX_CLOUD_ADDR`.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Run install.sh on Linux + Mac, to ensure we properly select the right artifact.

### DIFF
--- a/src/pixie_cli/install.sh
+++ b/src/pixie_cli/install.sh
@@ -22,17 +22,10 @@
 ################################################################
 set -u
 
-CLOUD_ADDR=${PL_CLOUD_ADDR:-"work.withpixie.ai"}
 DEFAULT_INSTALL_PATH=/usr/local/bin
 ARTIFACT_NAME=cli_darwin_universal
-USE_VERSION=${PL_CLI_VERSION:-latest}
 USER_INSTALL_PATH="$HOME/bin"
-ARTIFACT_BUCKET="pixie-dev-public"
-if [[ $USE_VERSION == *"-"* ]]; then
-  ARTIFACT_BUCKET="pixie-prod-artifacts"
-fi
-ARTIFACT_BASE_PATH="https://storage.googleapis.com/${ARTIFACT_BUCKET}/cli"
-
+GITHUB_REPO="pixie-io/pixie"
 
 PIXIE_BANNER="
   ___  _       _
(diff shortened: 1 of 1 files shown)


src/pixie_cli/install.sh
```diff
@@ -41,6 +34,17 @@ PIXIE_BANNER="
...
+ARTIFACT_BASE_PATH="https://github.com/${GITHUB_REPO}/releases/download/release%2Fcli%2Fv"
+
+# Fetch latest version.
+LATEST_VERSION=""
+regex='\[CLI v(([0-9]+)\.([0-9]+)\.([0-9]+))\]'
+if [[ "$(curl -fsSL https://raw.githubusercontent.com/pixie-io/pixie/main/README.md)" =~ $regex ]]; then
+    LATEST_VERSION=${BASH_REMATCH[1]}
+fi
+
+USE_VERSION=${PX_CLI_VERSION:=${LATEST_VERSION}}
```
JamesMBartlett, line 46: We probably still want to support `PL_CLI_VERSION`. Also we previously didn't set `PL_CLI_VERSION` if it wasn't set already but this will set it (`:=` vs `:-`), I don't know if that was intentional but I don't really understand the change if it was

src/pixie_cli/install.sh
```diff
@@ -41,6 +34,17 @@ PIXIE_BANNER="
  |_|  |_|/_\_\|_|\___|
 "
 
+ARTIFACT_BASE_PATH="https://github.com/${GITHUB_REPO}/releases/download/release%2Fcli%2Fv"
+
+# Fetch latest version.
+LATEST_VERSION=""
+regex='\[CLI v(([0-9]+)\.([0-9]+)\.([0-9]+))\]'
+if [[ "$(curl -fsSL https://raw.githubusercontent.com/pixie-io/pixie/main/README.md)" =~ $regex ]]; then
```
JamesMBartlett, line 42: maybe grep for `<!--cli-latest-release-->` first, in case the pattern appears elsewhere in the readme. Also, could consider parsing the manifest instead of the readme, but that might be hard without `jq`

src/pixie_cli/install.sh
aimichelle: Whoops `:=` was unintentional.
I have updated it now to accept both `PL_CLI_VERSION` and `PX_CLI_VERSION`, with preference towards `PX_CLI_VERSION` if both are set.

src/pixie_cli/install.sh
```diff
@@ -219,26 +226,13 @@ execute curl -fsSL "$(artifact_url)" -o "${INSTALL_PATH}"/px_new
 execute chmod +x "${INSTALL_PATH}"/px_new
 execute mv "${INSTALL_PATH}"/px_new "${INSTALL_PATH}"/px
```
vihangm, line 227: I know this is not necessarily within the scope of this change but do you mind adding the sha256 check? It's also in the github releases. (Checking the .asc signature if the user has GPG installed is better but probably significantly more complex to implement)

vihangm (approved):

JamesMBartlett (approved):

vihangm (approved):

PR #1558, opened 2023-06-22
Fix dropdown selection in admin cluster detail view

NickLanam:
Summary: A wrong variable reference was causing "unknown cluster" to be selected instead of what the user clicked.

Relevant Issues: N/A

Type of change: /kind bugfix

Test Plan: Visit `/admin/clusters`. Click one. In the dialog that comes up, use the dropdown in the top left to choose another.
Now, it should reload the dialog with info for the new cluster correctly.


vihangm (approved):

vihangm:
Can you add a test for this?

NickLanam:
> Can you add a test for this?

Sure, will make a separate PR as this component needs to be adjusted to make that type of test possible.

PR #1561, opened 2023-06-22
Trace statically linked OpenSSL compatible TLS libraries

ddelnano:
Summary: Trace statically linked OpenSSL compatible TLS libraries

This adds a feature toggle for enabling TLS tracing applications statically linked with an OpenSSL compatible library. The plan is to enable this for internal clusters for 1 week before enabling it for all users (and making it the default).

We originally intended to add these probes for BoringSSL applications, however, detecting BoringSSL is more difficult than anticipated. One of the reliable indicators, checking for BoringSSL's [magic tag](https://github.com/google/boringssl/commit/89386ac89bef26888b8d866ebb98b696d82dfde8), is only available if an application is using a BoringSSL from Oct 2021 or later and from checking the following dependencies (different envoy distributions, Clickhouse, the Mono runtime, [cloudflare/boring](https://github.com/cloudflare/boring) applications and Go binaries with `boringcrypto` enabled) that wasn't the case. In addition, checking for the magic tag could cause performance issues.

Since our latest TLS tracing implementation provides a broad set of coverage, we opted to trace any application that contains one of the OpenSSL compatible symbols necessary for tracing (`SSL_write`). For BIO native applications, we will successfully trace the traffic. For non BIO native cases (envoy), the uprobes will trigger but won't capture any data. This was deemed an acceptable trade off since detecting BoringSSL was challenging and any indicator would likely involve a long tail of upstream adoption.

Relevant Issues: #692

Type of change: /kind feature

Test Plan: Verified this change through the following:
- [x] `boringssl_trace_bpf_test` verifies tracing is successful
- [x] Verified that new tls source (`kStaticallyLinkedSource`) is identified during `boringssl_trace_bpf_test` (when prometheus metrics are logged out)
```
$ ./scripts/sudo_bazel_run.sh src/stirling/source_connectors/socket_tracer:boringssl_trace_bpf_test
data_loss_bytes{protocol="kProtocolHTTP",tls_source="kStaticallyLinkedSource"} 1244
data_loss_bytes{protocol="kProtocolDNS",tls_source="kSSLNone"} 90
data_loss_bytes{protocol="kProtocolUnknown",tls_source="kSSLNone"} 0
# HELP conn_stats_bytes Total bytes of data tracked by conn stats for this protocol.
# TYPE conn_stats_bytes counter
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kStaticallyLinkedSource"} 1638
conn_stats_bytes{protocol="kProtocolDNS",tls_source="kSSLNone"} 0
conn_stats_bytes{protocol="kProtocolUnknown",tls_source="kSSLNone"} 76521
```

Changelog Message:
```release-note
Add support for tracing encrypted traffic for statically linked OpenSSL/BoringSSL applications. This functionality is currently disabled but will be enabled by default in an upcoming release.
```

### DIFF
--- a/bazel/external/boringssl.patch
+++ b/bazel/external/boringssl.patch
@@ -3,7 +3,7 @@
 @@ -206,7 +206,7 @@
    BIO_puts(bio.get(), "HTTP/1.0 200 OK\r\nContent-Type: text/plain\r\n\r\n");
    PrintConnectionInfo(bio.get(), ssl);
- 
+
 -  char request[4];
 +  char request[74];
    size_t request_len = 0;
--- a/bazel/external/boringssl.patch
+++ b/bazel/external/boringssl.patch
@@ -1,0 +1,11 @@
+--- src/tool/server.cc	2023-01-20 13:31:06.619119943 -0800
++++ src/tool/server.cc	2023-01-20 13:30:59.707052958 -0800
+@@ -206,7 +206,7 @@
+   BIO_puts(bio.get(), "HTTP/1.0 200 OK\r\nContent-Type: text/plain\r\n\r\n");
+   PrintConnectionInfo(bio.get(), ssl);
+ 
+-  char request[4];
++  char request[74];
+   size_t request_len = 0;
+   while (request_len < sizeof(request)) {
+     int ssl_ret =
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -124,7 +124,7 @@ def _cc_deps():
     _bazel_repo("com_google_protobuf", patches = ["//bazel/external:protobuf_gogo_hack.patch", "//bazel/external:protobuf_text_format.patch", "//bazel/external:protobuf_warning.patch"], patch_args = ["-p1"])
     _bazel_repo("com_github_grpc_grpc", patches = ["//bazel/external:grpc.patch", "//bazel/external:grpc_go_toolchain.patch", "//bazel/external:grpc_test_visibility.patch"], patch_args = ["-p1"])
 
-    _bazel_repo("boringssl")
+    _bazel_repo("boringssl", patches = ["//bazel/external:boringssl.patch"], patch_args = ["-p0"])
     _bazel_repo("com_google_benchmark")
     _bazel_repo("com_google_googletest")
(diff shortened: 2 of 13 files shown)


bazel/external/boringssl.patch
```diff
@@ -0,0 +1,11 @@
...
++++ src/tool/server.cc	2023-01-20 13:30:59.707052958 -0800
+@@ -206,7 +206,7 @@
+   BIO_puts(bio.get(), "HTTP/1.0 200 OK\r\nContent-Type: text/plain\r\n\r\n");
+   PrintConnectionInfo(bio.get(), ssl);
+ 
+-  char request[4];
++  char request[74];
+   size_t request_len = 0;
+   while (request_len < sizeof(request)) {
+     int ssl_ret =
```
ddelnano, line 11: Unfortunately `bssl` is not fully featured enough as a http server to work for our test case. Since it only reads 4 bytes of data, stirling never sees enough data to identify its an HTTP request.

I attempted to use nginx linked with BoringSSL as a test case, but ran into complications with getting it to build. Other potential options include using a c++ natsc client (since we link [BoringSSL](https://github.com/pixie-io/pixie/blob/68fc3cf8734a4a3359b2aa20cd07147659dddee8/third_party/BUILD.bazel#L70)), writing another BoringSSL C++ application or using something upstream. Unfortunately our upstream options are limited since we don't support tracing non BIO native applications (otherwise Envoy would work).

This patch is minor enough that I believe this is an acceptable tradeoff. The other options mentioned above require maintaining another test client or server application, which I wanted to avoid. I would rather invest that time in building new functionality that overlaps with the needs of this test case (a Clickhouse protocol parser would allow us to use a clickhouse server).

etep (approved):
Discussed 1:1, but, in a future change we should clean up the counting of uprobe attaches. There are some places where the count of uprobes attached can be wrong i.e. because we return the attach count in a status-or object which cannot capture the correct count, from inside a loop, if the first uprobe is attached but the second one fails. Related, but in `DeployOpenSSLUProbes` we could refactor the individual attach logic code blocks into separate functions for dynamic, NodeJS, and static.

vihangm (approved):

PR #1570, opened 2023-06-24
Add tests for the Org Setup page

vihangm:
Summary: This takes coverage for `pages/setup.tsx` from 0% to 100% (though there are some uncovered branches).
Also adds marginal coverage to the sidebar, footer, topbar as a sideeffect from the snapshot test.

Type of change: /kind test

Test Plan: Ran the test with yarn.

### DIFF
--- a/src/ui/src/pages/setup/setup-test.tsx
+++ b/src/ui/src/pages/setup/setup-test.tsx
@@ -110,23 +110,17 @@ describe('setup page', () => {
     const el = await within(container).getByRole('textbox');
 
     fireEvent.change(el, { target: { value: 'a' } });
-    await waitFor(() => {});
-    let errorMsg = screen.queryByText('Name is too short');
-    expect(errorMsg).not.toBeNull();
+    await screen.findByText('Name is too short');
 
     fireEvent.change(el, { target: { value: 'a'.repeat(51) } });
-    await waitFor(() => {});
-    errorMsg = screen.queryByText('Name is too long');
-    expect(errorMsg).not.toBeNull();
+    await screen.findByText('Name is too long');
 
     fireEvent.change(el, { target: { value: 'testingorg.com' } });
-    await waitFor(() => {});
-    errorMsg = screen.queryByText('Name must not contain special characters (ex. ./\\$@)');
-    expect(errorMsg).not.toBeNull();
+    await screen.findByText('Name must not contain special characters (ex. ./\\$@)');
 
     fireEvent.change(el, { target: { value: 'validorg' } });
     await waitFor(() => {});
-    errorMsg = screen.queryByText('Name is too short');
+    let errorMsg = screen.queryByText('Name is too short');
     expect(errorMsg).toBeNull();
     errorMsg = screen.queryByText('Name is too long');
     expect(errorMsg).toBeNull();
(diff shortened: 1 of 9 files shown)


src/ui/src/pages/setup/setup-test.tsx
```diff
@@ -0,0 +1,196 @@
...
+      ({ container } = render(
+        <Tester />,
+      ));
+    });
+    await waitFor(() => {});
+    const el = await within(container).getByRole('textbox');
+
+    fireEvent.change(el, { target: { value: 'a' } });
+    await waitFor(() => {});
+    let errorMsg = screen.queryByText('Name is too short');
```
NickLanam, line 114: All of these `queryByText` calls can instead be `await findByText` to avoid the same flakiness issue we saw last week with the cluster details test.

As a bonus, that method throws an error when it fails to find the element already, so the `not.toBeNull()` becomes redundant if you make that change.

src/ui/src/utils/env.tsx
```diff
@@ -28,8 +28,7 @@ const BUILD_TIMESTAMP = process.env.BUILD_TIMESTAMP;
 /* eslint-enable prefer-destructuring */
 
 const timestampSec = Number.parseInt(BUILD_TIMESTAMP, 10);
-const date = Number.isNaN(timestampSec) ? new Date() : new Date(timestampSec * 1000);
-const dateStr = format(date, 'yyyy.MM.dd.hh.mm');
+const dateStr = Number.isNaN(timestampSec) ? 'time unknown' : format(timestampSec * 1000, 'yyyy.MM.dd.hh.mm');
```
NickLanam, line 31: Does this materially change any other snapshots? If not, oof we're still missing even more tests than I thought.

src/ui/src/utils/env.tsx
```diff
@@ -28,8 +28,7 @@ const BUILD_TIMESTAMP = process.env.BUILD_TIMESTAMP;
 /* eslint-enable prefer-destructuring */
 
 const timestampSec = Number.parseInt(BUILD_TIMESTAMP, 10);
```
NickLanam, line 30: This is called `timestampSec` but it's being multiplied as if it's milliseconds below. What is it really?

NickLanam (approved):

src/ui/src/utils/env.tsx
vihangm: no as you see in this PR, no other changes and all tests pass.

src/ui/src/utils/env.tsx
vihangm: the TIMESTAMP passed in is in seconds.
the date constructor and the date-fns formatters expect millis. hence we multiply by 1000 below.

src/ui/src/pages/setup/setup-test.tsx
```diff
@@ -0,0 +1,196 @@
...
+      ({ container } = render(
+        <Tester />,
+      ));
+    });
+
+    jest.mock('axios');
+    Axios.post = jest.fn().mockResolvedValue({});
+
+    const hrefSpyFn = jest.fn();
+    Object.defineProperty(window.location, 'href', {
```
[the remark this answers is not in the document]
vihangm: done

src/ui/src/pages/setup/setup-test.tsx
vihangm: done

JamesMBartlett (approved):

PR #1604, opened 2023-07-11
Skip coverage of generated UI code

NickLanam:
Summary: `src/ui/src/types/generated` is, well, generated code. Coverage reporting should not be counting the 6,000+ lines of code in there.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: `yarn coverage`. The generated types should no longer count against the total.


JamesMBartlett (approved):
This is already ignored at the codecov level, but this is fine.

PR #1607, opened 2023-07-11
Show the same sidebar everywhere

NickLanam:
Summary: Instead of showing and hiding parts depending on where you are.
We don't have a huge number of items in this bar, so it makes sense to always have all of them.
In the admin and data retention views, it will pick the default cluster for the script links.

Relevant Issues: N/A

Type of change: /kind bugfix

Test Plan: Load the live views, the admin pages, and the data retention view.
All of them should keep the same sidebar, and whichever view you're on should be highlighted.


aimichelle:
I'm not sure that i like always showing the admin pages on the sidebar when you're not in the admin view. I feel like many of those pages are less important and shouldn't be front-and-center all of the time. It wasn't perfect, but can we go back to the behavior we had before these refactors?

NickLanam:
> I'm not sure that i like always showing the admin pages on the sidebar when you're not in the admin view. I feel like many of those pages are less important and shouldn't be front-and-center all of the time. It wasn't perfect, but can we go back to the behavior we had before these refactors?

Done. They no longer show in the live views but do still show in the admin views and in the data retention views.

src/ui/src/pages/auth/utils.ts
```diff
@@ -38,6 +39,12 @@ export const GetOAuthProvider = (): OAuthProviderClient => {
       return HydraClient;
     case 'oidc':
       return OIDCClient;
+    case undefined:
+      if (process.env.NODE_ENV === 'test' || process.env.JEST_WORKER_ID !== undefined) {
```
aimichelle, line 43: Any way we can avoid test fixtures leaking to the main source code?

NickLanam: Done.

aimichelle (approved):

Issue #1594, opened 2023-06-30
Pixie's OTEL exports tries to send non UTF-8 data as string protos


vihangm:
**Describe the bug**
The protobuf spec clearly states that `string` types must contain UTF-8 strings.
Pixie uses `bytes` for `StringColumn` since traced data might not always be valid UTF-8.
The OTel [spec](https://opentelemetry.io/docs/specs/otel/common/attribute-type-mapping/#string-values) also suggests using `string` or `bytes` depending on whether the data is valid UTF-8.

However the `Pixie` OTEL exporter uses `string` type when exporting columns that are `types::STRING`
See [here](https://github.com/pixie-io/pixie/blob/c092058a5172396ea9bead1e209c2ed4e2336943/src/carnot/exec/otel_export_sink_node.cc#L106-L109).

This causes errors like the following:
```
[libprotobuf ERROR external/com_google_protobuf/src/google/protobuf/wire_format_lite.cc:618] String field 'opentelemetry.proto.common.v1.AnyValue.string_value' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes. 
```
This causes Pixie to drop such data during the export.

**Expected behavior**
Pixie should export data to OTel using the bytes type when there are non UTF-8 strings involved.

rosenhouse:
Is this issue fixed by #1617?

aimichelle:
Hey @rosenhouse !  Yup, it should be fixed by #1617 . 
Although there is one thing we were thinking about. Currently we'll send the data in the bytes field if it contains non-UTF-8 characters. However, it does seem annoying from the consumer-side to have to switch between bytes/string depending on the contents.
Would it actually be better for you all if we just send all data in bytes, assuming any could contain non-UTF-8.

rosenhouse:
Skimming the OTLP specs, it sounds like you're doing the right thing now: https://opentelemetry.io/docs/specs/otel/common/attribute-type-mapping/#string-values

I guess that depends on how its treated by common OTLP servers.  We're using [opentelemetry-collector-contrib](https://github.com/open-telemetry/) and I know others who are using Telegraf.

We should probably add some tests to the later stages of our processing to cover this non-UTF-8 case.

PR #1617, opened 2023-07-12
OTel export should send non-utf8 data as bytes

aimichelle:
Summary: When doing an otel export, Pixie sends all string data as a string value. This does not work for data with non-utf8 characters and results in many errors. Instead we should detect if non-utf8 characters are present and send bytes data instead.

Relevant Issues: #1594

Type of change: /kind bug

Test Plan: Added unit test, skaffolded Vizier

Changelog Message:
```release-note
Fix bug where we were sending data with non-UTF-8 characters as strings during OTel export.
```

### DIFF
--- a/src/carnot/exec/otel_export_sink_node.cc
+++ b/src/carnot/exec/otel_export_sink_node.cc
@@ -19,6 +19,7 @@
 #include "src/carnot/exec/otel_export_sink_node.h"
 
 #include <rapidjson/document.h>
+#include <simdutf.h>
 #include <chrono>
 #include <memory>
 #include <queue>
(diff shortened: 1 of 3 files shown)


src/carnot/exec/otel_export_sink_node.cc
```diff
@@ -90,6 +91,14 @@ Status OTelExportSinkNode::CloseImpl(ExecState* exec_state) {
   return Status::OK();
 }
 
+void SetStringOrBytes(std::string str, ::opentelemetry::proto::common::v1::KeyValue* otel_attr) {
+  if (!simdutf::validate_utf8(str.data(), str.length())) {
```
JamesMBartlett, line 95: I'm a little worried about the performance overhead of this check for every single byte of data that passes through, but at the same time I don't really see an alternative.

JamesMBartlett (approved):

src/carnot/exec/otel_export_sink_node_test.cc
```diff
@@ -146,6 +146,69 @@ metrics {
...
+          [&i, &actual_protos, &auth_context](const auto& ctx, const auto& proto, const auto&) {
+            auth_context = ctx->auth_context();
+            actual_protos[i] = proto;
+            ++i;
+            return grpc::Status::OK;
+          }));
+
+  auto tester = exec::ExecNodeTester<OTelExportSinkNode, plan::OTelExportSinkOperator>(
+      *plan_node, output_rd, {input_rd}, exec_state_.get());
+  char non_utf_8_bytes[] = {static_cast<char>(0xC0)};
```
JamesMBartlett, line 194: I think if you use an `std::string` here instead of `char[]`, it will avoid the asan error

aimichelle:
It's not perfect, but I did a quick benchmark on a cluster with a bunch of demos apps running. I timed the `ConsumeSpans` function (minus the export part) with/without these changes. 
Looks like with the changes `ConsumeSpans` takes an average of 82ms, and without the changes it takes an average of 67ms.

JamesMBartlett (approved):

Issue #1620, opened 2023-07-14
kelvin crash when handling certain types of expired metadata


caphrim007:
**Describe the bug**

Under certain conditionsm `kelvin` will start up and crash after attempting to remove old metadata. This can result in a CrashLoopBackoff which will entirely prevent kelvin from running and, by extension, the rest of the pixie stack from starting.

**To Reproduce**

The behavior appears to manifest when the invocation of the `CleanupExpiredMetadata` runs. This function is missing handling of ReplicaSets and Deployments. Therefore, when one of these is found in the cleanup process, kelvin will raise an "Unexpected object tyoe" error and log a DFATAL message.

**Expected behavior**

kelvin should be able to handle ReplicaSet and Deployment expired metadata removal without crashing.

**Screenshots**
If applicable, add screenshots to help explain your problem. Please make sure the screenshot does not contain any sensitive information such as API keys or access tokens.

**Logs**
Please attach the logs by running the following command:
```
I20230713 23:39:56.575673     1 metadata_state.cc:323] Didn't find pod UID cf273f86-b618-47dd-8f5d-f933c7ad4137 for service pl/kelvin-service
I20230713 23:39:56.577438     1 metadata_state.cc:323] Didn't find pod UID cf273f86-b618-47dd-8f5d-f933c7ad4137 for service pl/kelvin-service
I20230713 23:39:56.682813     1 metadata_state.cc:323] Didn't find pod UID ec212f83-0ac3-474c-a558-5b2914fbc46c for service mas-be/cloud-migration-service
I20230713 23:39:56.682951     1 metadata_state.cc:323] Didn't find pod UID ec212f83-0ac3-474c-a558-5b2914fbc46c for service mas-be/cloud-migration-service
I20230713 23:39:56.707820     1 metadata_state.cc:323] Didn't find pod UID ec212f83-0ac3-474c-a558-5b2914fbc46c for service mas-be/cloud-migration-service
I20230713 23:39:57.333297     1 metadata_state.h:111] Service CIDR updated to 10.100.0.0/16
F20230713 23:39:57.333928     1 metadata_state.cc:492] Unexpected object type: 4
*** Check failure stack trace: ***
E20230713 23:39:57.333977     1 signal_action.cc:63] Caught Aborted, suspect faulting address 0x277400000001. Trace:
**************************
PC: @     0x7f2db0a3587c  (unknown)  abort
    @     0x55ac86eef8dd  (unknown)  google::LogMessage::Fail()
[...]

PR #1621, opened 2023-07-14
fix: add cleanup of expirred metadata for ReplicaSets and Deployments

caphrim007:
Summary: The existing cleanup of resources does not include these kinds. When kelvin encounters them in a cluster, a fatal error is raised and a traceback printed as kelvin crashes (in debug builds). In non-debug builds, the metadata service will just continue to collect these expired resources; never removing them.

This patch adds cleanup of these resources to prevent crashing (in debug builds) and to reclaim resources and do housekeeping in non-debug builds.

Relevant Issues: #1620

Type of change: /kind bugfix

Test Plan: Tests in the metadata_state_test's CleanupExpiredMetadata were enhanced to include tests for these two kinds

### DIFF
--- a/src/shared/metadata/metadata_state.cc
+++ b/src/shared/metadata/metadata_state.cc
@@ -532,6 +532,18 @@ Status K8sMetadataState::CleanupExpiredMetadata(int64_t now, int64_t retention_t
           services_by_name_.erase({k8s_object->ns(), k8s_object->name()});
         }
         break;
+      case K8sObjectType::kReplicaSet:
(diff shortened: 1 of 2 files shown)


aimichelle (approved):

aimichelle:
@caphrim007 we require our commits to be GPG signed. Would you mind following these instructions? https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

caphrim007:
@aimichelle I just forgot to upload it to my GH account. check now?

Issue #1163, opened 2023-04-05
Add record & replay capability to Stirling data connector.


etep:
**Is your feature request related to a problem? Please describe.**
We would like to create tests that can reproduce issues such as #946.

**Describe the solution you'd like**
Capability to record events from eBPF, in particular, a recording of events read from the eBPF perf buffers. The recorded events could then be replayed into the Stirling source connectors.

**Describe alternatives you've considered**
Crafted test programs are useful, but we don't always know how to craft certain cases of interest.

PR #1623, opened 2023-07-14
`stirling/bpf_tools/bcc_wrapper.h`: Add wrapped BPF maps & arrays.

etep:
Summary: We add wrappers for BPF maps and arrays to our BCC wrapper class. These wrappers anticipate our upcoming "record & replay" feature. In specific, they will be used to shim in recording logic and return replayed values, i.e. when either record or replay ("RR") is in use. When "RR" is not in use, they pass through to the underlying BPF map or array.

Relevant Issues: https://github.com/pixie-io/pixie/issues/1163

Type of change: /kind feature

Test Plan: Existing test coverage exercises all of these.

### DIFF
--- a/src/stirling/bpf_tools/bcc_wrapper.h
+++ b/src/stirling/bpf_tools/bcc_wrapper.h
@@ -479,10 +479,6 @@ class WrappedBCCStackTable {
   std::unique_ptr<U> underlying_;
 };
 
-// ebpf::BPFStackTable GetStackTable(const std::string& table_name) {
-//   return bpf_.get_stack_table(table_name);
-// }
-
 }  // namespace bpf_tools
 }  // namespace stirling
 }  // namespace px
--- a/src/stirling/bpf_tools/bcc_wrapper.h
+++ b/src/stirling/bpf_tools/bcc_wrapper.h
@@ -303,10 +303,6 @@ class WrappedBCCArrayTable {
  public:
   using U = ebpf::BPFArrayTable<T>;
 
-  WrappedBCCArrayTable(bpf_tools::BCCWrapper* bcc, const std::string& name) : name_(name) {
-    underlying_ = std::make_unique<U>(bcc->BPF().get_array_table<T>(name_));
-  }
-
   static std::unique_ptr<WrappedBCCArrayTable> Create(bpf_tools::BCCWrapper* bcc,
                                                       const std::string& name) {
     return std::unique_ptr<WrappedBCCArrayTable>(new WrappedBCCArrayTable(bcc, name));
@@ -332,6 +328,10 @@ class WrappedBCCArrayTable {
   }
 
  private:
+  WrappedBCCArrayTable(bpf_tools::BCCWrapper* bcc, const std::string& name) : name_(name) {
+    underlying_ = std::make_unique<U>(bcc->BPF().get_array_table<T>(name_));
+  }
+
   const std::string name_;
   std::unique_ptr<U> underlying_;
 };
@@ -341,13 +341,9 @@ class WrappedBCCMap {
  public:
   using U = ebpf::BPFHashTable<K, V>;
 
-  WrappedBCCMap(bpf_tools::BCCWrapper* bcc, const std::string& name) : name_(name) {
-    underlying_ = std::make_unique<U>(bcc->BPF().get_hash_table<K, V>(name_));
-  }
-
-  static std::unique_ptr<WrappedBCCMap> Create(bpf_tools::BCCWrapper* bcc,
-                                               const std::string& name) {
-    return std::unique_ptr<WrappedBCCMap>(new WrappedBCCMap(bcc, name));
+  static std::unique_ptr<WrappedBCCMap> Create(bpf_tools::BCCWrapper* bcc, const std::string& name,
+                                               const bool user_space_managed = false) {
+    return std::unique_ptr<WrappedBCCMap>(new WrappedBCCMap(bcc, name, user_space_managed));
   }
 
   size_t capacity() const { return underlying_->capacity(); }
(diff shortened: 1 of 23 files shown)


src/stirling/bpf_tools/bcc_wrapper.h
```diff
@@ -480,6 +455,176 @@ class BCCWrapper {
...
+      r[k] = v;
+      if (clear_table) {
+        PX_UNUSED(underlying_->remove_value(k));
+      }
+    }
+    if (clear_table) {
+      shadow_keys_.clear();
+    }
+    return r;
+  }
```
ddelnano, line 557: Is there a reason this needs to take on reimplementing a function available on the `underlying_` instance? I think it would be best to always defer to the underlying BCC function rather than including BCC implementation details in this shim.

etep: We updated the wrapped BCC map to use the shadow keys only if template parameter `kUserSpaceManaged` is set to `true`. It should be the case that iterating over the keys in user space is faster than through iterated syscalls.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -651,15 +652,14 @@ void SocketTraceConnector::CheckTracerState() {
     return;
   }
 
-  int error_code;
-  openssl_trace_state_->get_value(kOpenSSLTraceStatusIdx, error_code);
+  const int error_code = (openssl_trace_state_->GetValue(kOpenSSLTraceStatusIdx)).ConsumeValueOr(0);
```
ddelnano, line 655:
```suggestion
  const int error_code = (openssl_trace_state_->GetValue(kOpenSSLTraceStatusIdx)).ConsumeValueOr(kOpenSSLTraceOk);
```

src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
```diff
@@ -415,7 +409,7 @@ StatusOr<int> UProbeManager::AttachOpenSSLUProbesOnDynamicLib(uint32_t pid) {
     // before the BPF map is updated. This value is cleaned up when the upid is
     // terminated, so if attachment fails it will be deleted prior to the pid being
     // reused.
-    openssl_source_map_->UpdateValue(pid, ssl_source);
+    PX_RETURN_IF_ERROR(openssl_source_map_->SetValue(pid, ssl_source));
```
ddelnano, line 412: I don't think this should cause a uprobe attachment to fail. Even though it's ideal to have each traced application attributed correctly, the TLS tracing functions properly if this fails (falling back to `kSSLUnspecified`). Given that my preference is to not check this return value for an error.

src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
```diff
@@ -499,7 +493,7 @@ StatusOr<int> UProbeManager::AttachNodeJsOpenSSLUprobes(const uint32_t pid) {
   // before the BPF map is updated. This value is cleaned up when the upid is
   // terminated, so if attachment fails it will be deleted prior to the pid being
   // reused.
-  openssl_source_map_->UpdateValue(pid, kNodeJSSource);
+  PX_RETURN_IF_ERROR(openssl_source_map_->SetValue(pid, kNodeJSSource));
```
ddelnano, line 496: Same consideration here.

src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
```diff
@@ -654,7 +648,7 @@ int UProbeManager::DeployOpenSSLUProbes(const absl::flat_hash_set<md::UPID>& pid
       // before the BPF map is updated. This value is cleaned up when the upid is
       // terminated, so if attachment fails it will be deleted prior to the pid being
       // reused.
-      openssl_source_map_->UpdateValue(pid.pid(), kStaticallyLinkedSource);
+      PX_UNUSED(openssl_source_map_->SetValue(pid.pid(), kStaticallyLinkedSource));
```
ddelnano, line 651: This is what I would prefer to do with the other `openssl_source_map_->SetValue` calls. Regardless of what we pick, I think it's best to treat all of them the same.

src/stirling/bpf_tools/bcc_wrapper.h
```diff
@@ -480,6 +455,192 @@ class BCCWrapper {
...
+      r.push_back({k, v});
+      if (clear_table) {
+        PX_UNUSED(underlying_->remove_value(k));
+      }
+    }
+    if (clear_table) {
+      shadow_keys_.clear();
+    }
+    return r;
+  }
```
ddelnano, line 568: Did the previous user space map implementation even provide this implementation? It appears to me this is new API surface area and didn't exist before. Assuming that understanding is correct, my preference would be to only add what is used.

ddelnano (commented):
Overall looks good. Added some questions and minor comments.

src/stirling/bpf_tools/bcc_wrapper.h
etep: I'd like to agree with the insight here, but it is tricky. I removed the impl. and added `static_assert(kUserSpaceManaged)`. However, this static assertion failed (and I'd like to disagree with the compiler insomuch as the offending instantiation was not user space managed). Here are some other options:
1. Return a `StatusOr` from this method.
2. `ECHECK(false)` for user space managed.
3. More template magic to ensure that we don't compile this wrongly.

src/stirling/source_connectors/perf_profiler/bcc_bpf_intf/stack_event.h
```diff
@@ -48,6 +52,23 @@ struct stack_trace_key_t {
...
+  friend bool operator==(const stack_trace_key_t& lhs, const stack_trace_key_t& rhs) {
+    if (lhs.upid != rhs.upid) {
+      return false;
+    }
+    if (lhs.user_stack_id != rhs.user_stack_id) {
+      return false;
+    }
+    return lhs.kernel_stack_id == rhs.kernel_stack_id;
+  }
+#endif
```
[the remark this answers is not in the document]
etep: The shadow keys in the wrapped bcc map force us to supply the hash impl.

src/stirling/source_connectors/perf_profiler/perf_profile_connector.cc
```diff
@@ -343,34 +339,32 @@ void PerfProfileConnector::ProcessBPFStackTraces(ConnectorContext* ctx, DataTabl
...
   auto& stack_traces = using_map_set_a ? stack_traces_a_ : stack_traces_b_;
-  auto& histo_perf_buf = using_map_set_a ? histogram_a_perf_buffer_ : histogram_b_perf_buffer_;
+  auto& perfbuf_name = using_map_set_a ? kHistogramAName : kHistogramBName;
   const uint32_t sample_count_idx = using_map_set_a ? kSampleCountAIdx : kSampleCountBIdx;
 
   // Read out the perf buffer that contains the histogram for this iteration.
   // TODO(jps): change PollPerfBuffer() to use std::chrono.
   constexpr int kPollTimeoutMS = 0;
-  histo_perf_buf->poll(kPollTimeoutMS);
+  PollPerfBuffer(perfbuf_name, kPollTimeoutMS);
```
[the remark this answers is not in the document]
etep: The main reason we find it this way is legacy. I did not introduce a wrapped perf buffer simply to follow our current setup. Wrapped perf buffer has crossed my mind, and so has moving all the wrapped bcc maps & arrays into bcc wrapper. There are some degrees of freedom here.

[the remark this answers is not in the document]
ddelnano: Sounds good. I'm fine with the current implementation, but I wanted to make sure I understood the rationale 👍

src/stirling/source_connectors/perf_profiler/bcc_bpf_intf/stack_event.h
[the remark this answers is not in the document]
ddelnano: Thanks for clarifying 👍. Just wanted to make sure I didn't miss something important since I wasn't able to make that connection on my first pass.

src/stirling/bpf_tools/bcc_wrapper.h
ddelnano: ah, I see. In that case 5 is good with me.

ddelnano (approved):

src/stirling/bpf_tools/bcc_wrapper.h
```diff
@@ -480,6 +455,191 @@ class BCCWrapper {
...
+ private:
+  WrappedBCCStackTable(bpf_tools::BCCWrapper* bcc, const std::string& name) : name_(name) {
+    underlying_ = std::make_unique<U>(bcc->BPF().get_stack_table(name_));
+  }
+
+  const std::string name_;
+  std::unique_ptr<U> underlying_;
+};
+
+// ebpf::BPFStackTable GetStackTable(const std::string& table_name) {
```
aimichelle, line 639: Any reason this is commented out? Should it be removed?

etep: Yikes, that should be removed.

aimichelle (approved):

PR #1633, opened 2023-07-21
Fix default metadata namespaces

aimichelle:
Summary: #1368 added the ability to limit the metadata service to specific namespaces. By default, if no namespaces are mentioned, the flag is supposed to choose all namespaces.
However, if the environment variable is not specified, it is actually parsed as an empty list and the default is not properly chosen. Instead, we can assume no users actually want to watch 0 namespaces, and set it to all namespaces when that occurs.

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Skaffold deploy metadata service

### DIFF
--- a/src/vizier/services/metadata/metadata_server.go
+++ b/src/vizier/services/metadata/metadata_server.go
@@ -238,6 +238,9 @@ func main() {
 	mdh := k8smeta.NewHandler(updateCh, k8sMds, k8sMds, nc)
 
 	namespaces := viper.GetStringSlice("metadata_namespaces")
+	if len(namespaces) == 0 {
+		namespaces = []string{v1.NamespaceAll}
+	}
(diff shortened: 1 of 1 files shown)


ddelnano (approved):

ddelnano:
What is injecting the environment variable? Is that the `px` or `helm` deploy? I scanned #1368 and didn't see where what env var is bound to that flag.

aimichelle:
> What is injecting the environment variable? Is that the `px` or `helm` deploy? I scanned #1368 and didn't see where what env var is bound to that flag.

We should probably add it to both `px` and `helm` to inject the environment variable. Currently you can only add it by either manually editing the metadata deployment, or adding it as a "patch" in `px`/`helm`.

vihangm (approved):

Issue #598, opened 2022-09-22
Ability to add tolerations to PEMs


aimichelle:
**Is your feature request related to a problem? Please describe.**
Users would like to easily add tolerations in order to deploy PEMs to nodes that have taints. 
This can be done by manually updating the resources to include the tolerations, however these changes aren't sticky after auto-updates. 
You can currently use patches to add the toleration or nodeSelector to various resources, but this is prone to syntax errors. For example, setting a nodeSelector to the PEMs using helm:
```
patches.vizier-pem='\\{\"spec\"\: {\"template\"\: {\"spec\"\: { \"nodeSelector\"\: {\"pixie\"\: \"allow\" }}}}}' \
```

**Describe the solution you'd like**
Users should be able to easily specify tolerations for their PEMs through the CLI and Helm Chart. This should look similarly to how tolerations are specified on K8s resources.

yousaf-aftab:
Hi @aimichelle I'm newbie and I'm interested in this issue. Can you tell a little bit about how you are passing toleration parameters in the following  command 
```
patches.vizier-pem='\\{\"spec\"\: {\"template\"\: {\"spec\"\: { \"nodeSelector\"\: {\"pixie\"\: \"allow\" }}}}}' \
```
because setting nodeSelector will not tolerate the taint, it will only match the labels of node and scheduler will try to schedule the pod on that node if the labels matches but it will not be scheduled if
 you don't pass the tolerations as well.

aimichelle:
Hey @yousaf-aftab ! The patching mechanism is essentially performing the equivalent of a [kubectl patch](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/).

I think something similar to what was done for nodeSelector (example changes here https://github.com/pixie-io/pixie/pull/362) would work really well for the tolerations! Let me know if you have any questions after going through that PR.

aimichelle:
Closed thanks to @benkilimnik 's work on #1649 and #1648!

PR #1648, opened 2023-08-03
Add support for tolerations in operator

benkilimnik:
Summary: Adds support for tolerations to enable users to deploy PEMs to nodes that have taints through the CLI.

Relevant Issues: #598

Type of change: /kind feature

Test Plan: Used skaffold to deploy an unreleased version of the operator. Verified that tolerations parameter actually gets passed through to the pod spec.


aimichelle:
Looks like it is failing `pr-genfile-checker`, so you may need to rerun `go generate`.

aimichelle (approved):

Issue #1632, opened 2023-07-21
Vizier feature flags are set only at deploy time


aimichelle:
**Describe the bug**
We have a system for setting feature flags on Viziers. The flow is as follows:
1. When the operator is deploying/updating Vizier to a cluster, it makes a request to the control plane's config-manager-service: https://github.com/pixie-io/pixie/blob/ea9a357864dcde4df7d8476b40a18caf58a3e569/src/api/proto/cloudpb/cloudapi.proto#L945 The `deployKey` is included as part of the vzspec.
2. The config manager service uses the deployKey to lookup which org is attempting to deploy/update the Vizier. 
3. The org ID is then used to make a request to LaunchDarkly for the Vizier feature flags. Feature flags are set on a per-org basis, for example org A may have featureFlag1 set to true.
4. The config manager updates the Vizier YAMLs with the correct feature flags flipped.
5. Operator receives the Vizier YAMLs as a response from the config manager service, and applies the YAMLs.

However, this flow only works when the deployKey is valid (so basically, only at deploy time). Most times, a deployKey is generated when deploying a Vizier, and automatically deleted once the Vizier is up-and-running. When the Vizier updates, since the deployKey is no longer valid, the config manager service is unable to fetch the associated orgID. If it is unable to get an orgID, it simply does not add any of the feature flags to the Vizier YAMLs.

**Expected behavior**
Regardless of whether a user is deploying or updating a Vizier, the feature flags should be set properly based on which org is deploying the Vizier. Since the deploy key is not always valid, we can determine the org associated with a Vizier through its Vizier ID. 
1. Add VizierID to `ConfigForVizierRequest` protos: `src/api/proto/cloudpb/cloudapi.proto`, `src/cloud/config_manager/configmanagerpb/service.proto`
2. Update the operator to include VizierID in `ConfigForVizierRequest`: https://github.com/pixie-io/pixie/blob/ea9a357864dcde4df7d8476b40a18caf58a3e569/src/operator/controllers/vizier_controller.go#L809  Note: Vizier ID will not be available if the user is deploying Vizier for the first time, but it will be available if the Vizier is going through an update. We can use https://github.com/pixie-io/pixie/blob/ea9a357864dcde4df7d8476b40a18caf58a3e569/src/operator/controllers/vizier_controller.go#L1018-L1024 to get the Vizier ID.
[...]

PR #1650, opened 2023-08-03
 Allow VizierID to be used to find the Vizier's org

kpattaswamy:
Summary: This PR allows the VizierID to be used when getting the org as opposed to only relying on the DeployKey when adding feature flags.

Relevant Issues: #1632

Type of change: /kind bug

Test Plan: Another PR needs to come in on the operator side to retrieve/send the VizierID to the cloud. Getting this PR in on the cloud side will simplify the process of testing the integration between the two later.


src/api/proto/cloudpb/cloudapi.proto
```diff
@@ -956,6 +956,8 @@ message ConfigForVizierRequest {
   vizierconfigpb.VizierSpec vz_spec = 2;
   // Kubernetes version of the cluster Vizier is running on.
   string k8s_version = 3 [ (gogoproto.customname) = "K8sVersion" ];
+  // ID of the Vizier (Cluster ID).
+  string vizier_id = 4 [ (gogoproto.customname) = "VizierID" ];
```
ddelnano, line 960: I'm not knowledgable on cloud so my comment should be vetted, but I think we will want to update [this](https://github.com/pixie-io/pixie/blob/26a31df536614d5e75133dc3f39e5c8573382834/src/cloud/api/controllers/config_grpc.go#L43) ConfigForVizierRequest from cloud to the config manager service in this review.

Without that change, I believe the next step of our plan (to modify the operator to provide a vizier id on update) won't be able to have its VizierID parameter propagated fully through to the config manager service.

cc @vihangm

src/cloud/config_manager/controllers/server.go
```diff
@@ -218,7 +219,19 @@ func (s *Server) GetConfigForVizier(ctx context.Context,
 	AddDefaultTableStoreSize(tmplValues.PEMMemoryRequest, tmplValues.CustomPEMFlags)
 
 	// Next we inject any feature flags that we want to set for this org.
-	orgID, err := s.getOrgIDForDeployKey(tmplValues.DeployKey)
+	var orgID uuid.UUID
+	// Attempt to get the org from Vizier, otherwise from the DeployKey.
+	if in.VizierID != "" {
```
kpattaswamy, line 224: In this part of the logic, I wanted to get some feedback on a different approach where the org information is attempted initially by `GetOrgFromVizier` and if that doesn't work then tried by `getOrgIDForDeployKey`.  
Currently it tries one method and if that doesn't work it wont try the other.

src/api/proto/cloudpb/cloudapi.proto
aimichelle: That is correct, we need to add VizierID to this proto as well: https://github.com/pixie-io/pixie/blob/26a31df536614d5e75133dc3f39e5c8573382834/src/cloud/config_manager/configmanagerpb/service.proto#L38

aimichelle: My bad, I was just going through the order for comments. I see this was done already in the PR!

aimichelle: Sorry I think I actually misunderstood your comment for a sec (thought you were talking about the protos). We do need to update that file you referenced above to pass the vizierID along in the ConfigForVizierRequest.

src/cloud/config_manager/controllers/server.go
```diff
@@ -52,6 +52,7 @@ type Server struct {
 	atClient            atpb.ArtifactTrackerClient
 	deployKeyClient     vzmgrpb.VZDeploymentKeyServiceClient
 	vzFeatureFlagClient VizierFeatureFlagClient
+	vzmgrClient         vzmgrpb.VZMgrServiceClient
```
aimichelle, line 55: We'll need to instantiate vzmgrClient (otherwise it will be nil here).
To do that, we'll need to update the constructor to take the vzmgrClient: https://github.com/pixie-io/pixie/blob/main/src/cloud/config_manager/controllers/server.go#L58 similar to how we do `deployKeyClient`.
And then we need to update https://github.com/pixie-io/pixie/blob/main/src/cloud/config_manager/config_manager_server.go#L96 to initiate vzMgrClient and pass it to the constructor.

src/cloud/config_manager/controllers/server.go
aimichelle: I think it might be more robust if we do something like:
- Check if getting org from deployKey works (I would start with this case first, since this is the status quo)
- If deployKey method doesn't work, fallback to using the VizierID.

src/cloud/config_manager/controllers/server.go
```diff
@@ -217,12 +219,29 @@ func (s *Server) GetConfigForVizier(ctx context.Context,
...
+	getOrgID := true
+	var orgID uuid.UUID
+	orgID, err = s.getOrgIDForDeployKey(tmplValues.DeployKey)
 	if err != nil || orgID == uuid.Nil {
-		log.WithError(err).Error("Error getting org ID from deploy key, skipping feature flag logic")
-	} else {
+		log.WithError(err).Error("Error getting org ID from deploy key")
+		getOrgID = false
+	}
+	if !getOrgID && in.VizierID != "" {
```
aimichelle, line 230: Instead of tracking a separate `getOrgID` var you can just check if `orgID == uuid.Nil`.

vihangm (approved):

PR #1652, opened 2023-08-04
Add TLS tracing source debugging mode and associated feature flag

ddelnano:
Summary: Add TLS tracing source debugging mode and associated feature flag

This debugging feature flag is intended for adhoc investigations for determining why a particular TLS source (libpython, statically linked open/boringssl, OpenSSL v1.x, etc) was traced for a given protocol. BoringSSL tracing was recently rolled out and it appeared this functionality was covering every protocol that Pixie supports. This was unexpected based on the open source / popular projects we believe should be covered by this tracing.

As I tested this change, I found a bug in how a TLS source is attributed. The libpython TLS source was identified as statically linked (see Test Plan below for details). I believe this is likely the discrepancy causing the statically linked source to show up  for every supported protocol. That bug will be addressed in a follow up change.

When reviewing the Test Plan, please note that the changes in 5a23004f566195d5c57e27ca47a4be2c358cc55b were needed to produce the prometheus output and to verify that the TLS source misattribution is present.

Relevant Issues: #692

Type of change: /kind feature

Test Plan: Verified the following
- [x] Simulating the mismatched fd case still causes the `DCHECK` to occur
<details><summary>openssl_trace_bpf_test with mismatched fd error</summary>

```
# Introduce change to simulate mismatched fd case.
The openssl_trace_bpf_test cases are known to require calling send/read multiple times for the same fd. This change will cause the DCHECK to occur on the second read/write call
--
$ git diff
diff --git a/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c b/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
index ba5df955dd..e1c64a1a7d 100644
--- a/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
+++ b/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
@@ -185,7 +185,7 @@ static __inline void propagate_fd_to_user_space_call(uint64_t pid_tgid, int fd)
int current_fd = nested_syscall_fd_ptr->fd;
if (current_fd == kInvalidFD) {
nested_syscall_fd_ptr->fd = fd;
-    } else if (current_fd != fd) {
+    } else /*if (current_fd != fd)*/ {
// Found two different fds during a single SSL_write/SSL_read call. This invalidates
// our tls tracing assumptions and must be recorded.
nested_syscall_fd_ptr->mismatched_fds = true;
 
[...]

PR #1655, opened 2023-08-07
Ensure `kStaticallyLinkedSource` is attributed when static TLS tracing probes are added

ddelnano:
Summary: Ensure `kStaticallyLinkedSource` is attributed when static TLS tracing probes are added

When developing #1652, we noticed that there are situations where our prometheus metrics are annotated with `kStaticallyLinkedSource` instead of the correct tls source. This is due to the following [conditional logic](https://github.com/pixie-io/pixie/blob/44a8338a60e74564d94d474a7ed723717d2ebaaf/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc#L646) using a mutable variable which was returning true even when uprobes were already attached. When tracing a dynamically linked process, the uprobe count will be recorded to a non zero value when the dynamic probes are [attached](https://github.com/pixie-io/pixie/blob/44a8338a60e74564d94d474a7ed723717d2ebaaf/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc#L615). The NodeJS uprobes in [AttachNodeJsOpenSSLUprobes](https://github.com/pixie-io/pixie/blob/44a8338a60e74564d94d474a7ed723717d2ebaaf/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc#L629C16-L629C42) will then reset the `count_or` integer value back to 0 (since dynamically linked applications cannot be NodeJS -- it's statically linked). This causes us to erroneously attempt (and fail) to attach the statically linked probes in addition to setting the tls source to `kStaticallyLinkedSource`.

I attempted to refactor the uprobe counting logic and this bug in a single change, but I believe it will increase the scope of this. The OpenSSL probe specs are shared amongst the dynamic, statically linked and NodeJS cases, which causes complications with counting the number of attached (rather than returning the potential count). While we are confident this bug is likely the source of the confusing BoringSSL tracing results, it would be beneficial to have that validation sooner.

I will follow up this change with refactoring this code, but would prefer to proceed with this fix to unblock the BoringSSL tracing validation.

Relevant Issues: #692

Type of change: /kind bug

Test Plan: Printed out the promtheus `conn_stats_bytes` metrics during `openssl_trace_bpf_test` and verified they match the given test case
<details><summary> openssl_trace_bpf_test metric output</summary>

```
$ ./scripts/sudo_bazel_run.sh -c dbg  src/stirling/source_connectors/socket_tracer:openssl_trace_bpf_test   --test_arg='--gtest_filter=OpenSSLTraceTest/*.ssl_capture_curl_client' 2>&1 \| grep '^conn_stats_bytes\\|TypeParam'
--
[----------] 1 test from OpenSSLTraceTest/0, where TypeParam = px::stirling::NginxOpenSSL_1_1_0_ContainerWrapper
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_1_1_Source"} 3191
conn_stats_bytes{protocol="kProtocolUnknown",tls_source="kSSLNone"} 3591
conn_stats_bytes{protocol="kProtocolDNS",tls_source="kSSLNone"} 0
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kSSLNone"} 416
[----------] 1 test from OpenSSLTraceTest/1, where TypeParam = px::stirling::NginxOpenSSL_1_1_1_ContainerWrapper
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_1_1_Source"} 6375
conn_stats_bytes{protocol="kProtocolUnknown",tls_source="kSSLNone"} 6989
conn_stats_bytes{protocol="kProtocolDNS",tls_source="kSSLNone"} 0
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kSSLNone"} 832
[----------] 1 test from OpenSSLTraceTest/2, where TypeParam = px::stirling::NginxOpenSSL_3_0_8_ContainerWrapper
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_3_Source"} 3184
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_1_1_Source"} 6375
conn_stats_bytes{protocol="kProtocolUnknown",tls_source="kSSLNone"} 10484
conn_stats_bytes{protocol="kProtocolDNS",tls_source="kSSLNone"} 307
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kSSLNone"} 1248
[----------] 1 test from OpenSSLTraceTest/3, where TypeParam = px::stirling::Python310ContainerWrapper
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibPythonSource"} 3650
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_3_Source"} 3184
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_1_1_Source"} 6375
conn_stats_bytes{protocol="kProtocolUnknown",tls_source="kSSLNone"} 14350
conn_stats_bytes{protocol="kProtocolDNS",tls_source="kSSLNone"} 307
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kSSLNone"} 1664
[----------] 1 test from OpenSSLTraceTest/4, where TypeParam = px::stirling::Node12_3_1ContainerWrapper
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kNodeJSSource"} 1360
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibPythonSource"} 3650
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_3_Source"} 3184
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_1_1_Source"} 6375
conn_stats_bytes{protocol="kProtocolUnknown",tls_source="kSSLNone"} 18367
conn_stats_bytes{protocol="kProtocolDNS",tls_source="kSSLNone"} 307
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kSSLNone"} 133152
[----------] 1 test from OpenSSLTraceTest/5, where TypeParam = px::stirling::Node14_18_1AlpineContainerWrapper
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kNodeJSSource"} 2743
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibPythonSource"} 3650
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_3_Source"} 3184
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kLibSSL_1_1_Source"} 6375
conn_stats_bytes{protocol="kProtocolUnknown",tls_source="kSSLNone"} 22319
conn_stats_bytes{protocol="kProtocolDNS",tls_source="kSSLNone"} 307
conn_stats_bytes{protocol="kProtocolHTTP",tls_source="kSSLNone"} 155820
```
</details>

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
+++ b/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
@@ -644,10 +644,6 @@ int UProbeManager::DeployOpenSSLUProbes(const absl::flat_hash_set<md::UPID>& pid
 
     // Attach uprobes to statically linked applications only if no other probes have been attached.
     if (FLAGS_stirling_trace_static_tls_binaries && count_or.ok() && count_or.ValueOrDie() == 0) {
-      // Optimisitcally update the SSL lib source since the probes can trigger
-      // before the BPF map is updated. This value is cleaned up when the upid is
-      // terminated, so if attachment fails it will be deleted prior to the pid being
-      // reused.
       count_or = AttachOpenSSLUProbesOnStaticBinary(pid.pid());
 
       if (count_or.ok() && count_or.ValueOrDie() > 0) {
--- a/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
+++ b/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
@@ -648,11 +648,11 @@ int UProbeManager::DeployOpenSSLUProbes(const absl::flat_hash_set<md::UPID>& pid
       // before the BPF map is updated. This value is cleaned up when the upid is
       // terminated, so if attachment fails it will be deleted prior to the pid being
       // reused.
-      PX_UNUSED(openssl_source_map_->SetValue(pid.pid(), kStaticallyLinkedSource));
       count_or = AttachOpenSSLUProbesOnStaticBinary(pid.pid());
 
-      if (count_or.ok()) {
+      if (count_or.ok() && count_or.ValueOrDie() > 0) {
         uprobe_count += count_or.ValueOrDie();
+        PX_UNUSED(openssl_source_map_->SetValue(pid.pid(), kStaticallyLinkedSource));
 
         VLOG(1) << absl::Substitute(
             "Attaching OpenSSL uprobes on executable statically linked OpenSSL library"


src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
```diff
@@ -648,11 +648,11 @@ int UProbeManager::DeployOpenSSLUProbes(const absl::flat_hash_set<md::UPID>& pid
       // before the BPF map is updated. This value is cleaned up when the upid is
```
etep, line 648: This comment will be out of date I think.

Is it possible to break this up (granted this might just be a future PR though)... i.e. like so:
1. Check if static ssl uprobes are appropriate.
2. Attach.
This is different from just "TryAttach" which is what we have now. The above two step approach would enable us to update the map before attach. (But I'm not sure how important that update before attach is, i.e. the out-of-date comment here makes me think it is important to update then attach).

etep (approved):

src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
ddelnano: ah, good call. I went ahead and removed that comment. I believe we don't need to do it optimistically and that's something I'll clean up with the following refactor.

vihangm (approved):

Issue #1056, opened 2023-03-15
Only partial data is being captured by the pixie agent - No interpod traffic


benmathews:
**Describe the bug**
Only partial data is being captured by the pixie agent

**To Reproduce**
1. Install pixie with `px deploy`
2. Run the `px/cluster` script at https://work.withpixie.ai/live
3. Observe that pixie knows about namespaces and pods, but none of the traffic between them.
![image](https://user-images.githubusercontent.com/446380/225333831-62503acb-d68b-4f29-8368-f431d61f14df.png)
4 Run the `px/cluster` script from the CLI
```
➜ px run px/cluster
<snip>
Table ID: Services
  SERVICE                                                                                                                                                                                            POD COUNT  HTTP LATENCY IN  HTTP REQ THROUGHPUT IN  HTTP ERROR RATE IN  INBOUND CONNS  OUTBOUND CONNS  
  ["nebo/history-headless","nebo/history"]                                                                                                                                                           7                           0.000000 /sec           0.00 %              0 B/sec        0 B/sec         
  nebo/propm-properties-api                                                                                                                                                                          3                           0.000000 /sec           0.00 %              0 B/sec        0 B/sec         
<snip>
```

**Expected behavior**
Pixie should know about the HTTP, kafka, etc. traffic.

**Logs**
[pixie_logs_20230315080754.zip](https://github.com/pixie-io/pixie/files/10980870/pixie_logs_20230315080754.zip)

**App information (please complete the following information):**
```
➜ px version
Pixie CLI
0.7.17+Distribution.1f8cdfb.20220815151559.3.jenkins
☸ sekube (pl) in ~ via 🐹 v1.20.1 
➜ k version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.17", GitCommit:"a7736eaf34d823d7652415337ac0ad06db9167fc", GitTreeState:"clean", BuildDate:"2023-01-04T19:18:50Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.15", GitCommit:"b84cb8ab29366daa1bba65bc67f54de2f6c34848", GitTreeState:"clean", BuildDate:"2022-12-08T10:42:57Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"linux/amd64"}
$ uname -a
[...]

PR #1663, opened 2023-08-09
Fix inconsistent state after failed prepackaged header installation

### LINKED ISSUE #1092: Prepackaged linux header installation can result in partially applied state
Pixie recommends installing linux headers on a cluster's nodes, however, we have a prepackaged linux header fallback that can be used. This installation process requires downloading a linux header bundle and applies some patches to ensure the headers are in a proper state.

In #1056 we discovered that it's possible for the prepackaged header install to end up in an inconsistent state. The installation process must perform a number of steps and does not clean up after itself if one of those steps fail. This results in the following error:
```
Failed to install packaged Linux headers to /lib/modules/5.15.0-58-generic, error: Internal : Refusing to clobber existing directory: /usr/src/linux-headers-5.15.90-pl.
```

1. Remove the kernel config files necessary for header install ([source](https://github.com/pixie-io/pixie/blob/2190b2afa52f638a3114ad677e11b8d00d6c69e0/src/stirling/utils/linux_headers.cc#L309-L321))
2. Install pixie on the cluster the instance is a part of
3. Verify that the first error indicates the headers can't be installed
4. Verify that the subsequent errors indicate there is a leftover header directory present (error message shown above)

**Expected behavior**
The prepackaged header install should fail and clean up after itself or successfully apply.

benkilimnik:
Summary: Pixie's installation process requires downloading a Linux header bundle and applying patches. Issue #1056 identified that a failed installation process could lead to an inconsistent state, resulting in the error:

```bash
Failed to install packaged Linux headers to /lib/modules/5.15.0-58-generic, error: Internal: Refusing to clobber existing directory: /usr/src/linux-headers-5.15.90-pl.
```

The problem was that the installation process did not clean up after a failed attempt.

This PR modifies the extraction process for prepackaged Linux headers by introducing a staging directory. Extraction is now performed in this staging directory first, which helps ensure that the check for existing directories does not raise an error from a previous failed run. After successfully extracting and applying patches, the contents are moved from the staging directory to the expected directory.

Additional context: The function for extraction `Minitar::Extract` was modified to accept an additional parameter, `prefix_to_strip`. This helps remove any specified prefix from the pathnames of the untarred files to prevent unnecessary nesting in the staging directory.

Relevant Issues: Fixes #1092

Type of change: /kind bug

Test Plan: Removed the kernel config files on a node, then installed Pixie and verified that the initial error which previously led to inconsistent state is raised:
```
Failed to install packaged Linux headers to /lib/modules/5.15.0-58-generic, error: Not Found : No kernel config found. Searched: /proc/config,/proc/config.gz,/host/boot/config-5.15.0-58-generic,/host/lib/modules/5.15.0-58-generic/config,/proc/1/root/media/sr0/boot/config-virt,/proc/1/root/media/vda/boot/config-virt.
```
After this, I added the config files back and redeployed, verifying that things work as intended i.e. that the config files are found and that there are no leftover files from a failed installation in the expected directory.

### DIFF
--- a/src/stirling/utils/linux_headers.cc
+++ b/src/stirling/utils/linux_headers.cc
@@ -499,17 +499,14 @@ Status LinkHostLinuxHeaders(const std::filesystem::path& lib_modules_dir) {
 }
 
 Status ExtractPackagedHeaders(PackagedLinuxHeadersSpec* headers_package,
-                              std::string staging_directory, std::string expected_directory) {
-  // This is a loose check that we don't clobber what we *think* should the output directory.
-  // If someone built a tar.gz with an incorrect directory structure, this check wouldn't save us.
-  if (fs::Exists(expected_directory)) {
-    return error::Internal("Refusing to clobber existing directory: $0.", expected_directory);
-  }
+                              const std::string staging_directory,
+                              const std::string expected_directory) {
   std::filesystem::create_directories(staging_directory);
   // Extract the files from the tarball, stripping the leading prefix
   // "usr/src/linux-headers-$0.$1.$2-pl" to avoid unnecessary nesting in the staging directory.
   ::px::tools::Minitar minitar(headers_package->path.string());
-  PX_RETURN_IF_ERROR(minitar.Extract(staging_directory, expected_directory.substr(1)));
+  const std::string prefix_to_strip = expected_directory.substr(1);
+  PX_RETURN_IF_ERROR(minitar.Extract(staging_directory, prefix_to_strip));
   // Check that the staging path was created.
   if (!fs::Exists(staging_directory)) {
     return error::Internal(
@@ -623,16 +620,28 @@ Status InstallPackagedLinuxHeaders(const std::filesystem::path& lib_modules_dir)
   PX_ASSIGN_OR_RETURN(PackagedLinuxHeadersSpec packaged_headers,
                       FindClosestPackagedLinuxHeaders(kPackagedHeadersRoot, kernel_version));
   LOG(INFO) << absl::Substitute("Using packaged header: $0", packaged_headers.path.string());
-  std::string version =
+  const std::string version =
       absl::Substitute("$0.$1.$2-pl", packaged_headers.version.version,
                        packaged_headers.version.major_rev, packaged_headers.version.minor_rev);
-  std::string staging_directory = absl::Substitute("/usr/src/staging/linux-headers-" + version);
-  std::string expected_directory = absl::Substitute("/usr/src/linux-headers-" + version);
+  const std::string staging_directory = absl::StrCat("/usr/src/staging/linux-headers-", version);
+  const std::string expected_directory = absl::StrCat("/usr/src/linux-headers-" + version);
+  // Verify that the target directory doesn't already exist.
+  // If someone built a tar.gz with an incorrect directory structure, this check wouldn't save us.
+  if (fs::Exists(expected_directory)) {
+    return error::Internal(
+        "Not attempting to install packaged headers because the target directory already exists: "
+        "$0",
+        expected_directory);
+  }
+  // Extract the packaged headers to a staging directory, stripping the expected target directory
+  // prefix.
   PX_RETURN_IF_ERROR(
       ExtractPackagedHeaders(&packaged_headers, staging_directory, expected_directory));
+  // Modify version.h to the specific kernel version in the staged headers.
   PX_RETURN_IF_ERROR(ModifyKernelVersion(packaged_headers.path, kernel_version.code()));
+  // Find valid kernel config and patch the staged headers to match.
   PX_RETURN_IF_ERROR(ApplyConfigPatches(packaged_headers.path));
-  // move staging dir to final location
+  // Move the staged headers to the expected, target directory.
   std::filesystem::rename(packaged_headers.path, expected_directory);
   packaged_headers.path = expected_directory;
   PX_RETURN_IF_ERROR(fs::CreateSymlinkIfNotExists(packaged_headers.path, lib_modules_build_dir));
--- a/src/stirling/utils/linux_headers.cc
+++ b/src/stirling/utils/linux_headers.cc
@@ -498,12 +498,12 @@ Status LinkHostLinuxHeaders(const std::filesystem::path& lib_modules_dir) {
   return Status::OK();
 }
 
-Status ExtractPackagedHeaders(const PackagedLinuxHeadersSpec* headers_package,
+Status ExtractPackagedHeaders(const PackagedLinuxHeadersSpec& headers_package,
                               const std::string& staging_directory,
                               const std::string& expected_directory) {
   std::filesystem::create_directories(staging_directory);
   // Instantiate a minitar object with the path to the tarball.
-  ::px::tools::Minitar minitar(headers_package->path.string());
+  ::px::tools::Minitar minitar(headers_package.path.string());
   // Extract the files from the tarball, stripping the leading prefix
   // "usr/src/linux-headers-$0.$1.$2-pl" to avoid unnecessary nesting in the staging directory.
   std::string_view expected_directory_view = expected_directory;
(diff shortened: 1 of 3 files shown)


src/stirling/utils/linux_headers.cc
```diff
@@ -627,9 +623,18 @@ Status InstallPackagedLinuxHeaders(const std::filesystem::path& lib_modules_dir)
   PX_ASSIGN_OR_RETURN(PackagedLinuxHeadersSpec packaged_headers,
                       FindClosestPackagedLinuxHeaders(kPackagedHeadersRoot, kernel_version));
   LOG(INFO) << absl::Substitute("Using packaged header: $0", packaged_headers.path.string());
-  PX_RETURN_IF_ERROR(ExtractPackagedHeaders(&packaged_headers));
+  std::string version =
```
etep, line 626: `const std::string version` and a few others that can be made const.

src/stirling/utils/linux_headers.cc
```diff
@@ -627,9 +623,18 @@ Status InstallPackagedLinuxHeaders(const std::filesystem::path& lib_modules_dir)
...
+      absl::Substitute("$0.$1.$2-pl", packaged_headers.version.version,
+                       packaged_headers.version.major_rev, packaged_headers.version.minor_rev);
+  std::string staging_directory = absl::Substitute("/usr/src/staging/linux-headers-" + version);
+  std::string expected_directory = absl::Substitute("/usr/src/linux-headers-" + version);
```
etep, line 630: It looks like use of `absl::Substitute` is not needed here. We could take this in a few directions:
1. Don't use `absl::Substitute`.
2. `absl::Substitute("/usr/src/linux-headers-$0", version);`
3. Use `absl::StrCat`.

src/stirling/utils/linux_headers.cc
```diff
@@ -498,30 +498,26 @@ Status LinkHostLinuxHeaders(const std::filesystem::path& lib_modules_dir) {
...
-  // Extract the files.
+  std::filesystem::create_directories(staging_directory);
+  // Extract the files from the tarball, stripping the leading prefix
+  // "usr/src/linux-headers-$0.$1.$2-pl" to avoid unnecessary nesting in the staging directory.
   ::px::tools::Minitar minitar(headers_package->path.string());
-  PX_RETURN_IF_ERROR(minitar.Extract("/"));
-
-  // Check that the expected path was created.
-  if (!fs::Exists(expected_directory)) {
+  PX_RETURN_IF_ERROR(minitar.Extract(staging_directory, expected_directory.substr(1)));
```
etep, line 512: For clarity, would suggest a local like this:
`const std::string prefix_to_strip = expected_directory.substr(1);`

src/stirling/utils/linux_headers.cc
```diff
@@ -498,30 +498,26 @@ Status LinkHostLinuxHeaders(const std::filesystem::path& lib_modules_dir) {
...
-      absl::Substitute("/usr/src/linux-headers-$0.$1.$2-pl", headers_package->version.version,
-                       headers_package->version.major_rev, headers_package->version.minor_rev);
-
+Status ExtractPackagedHeaders(PackagedLinuxHeadersSpec* headers_package,
+                              std::string staging_directory, std::string expected_directory) {
   // This is a loose check that we don't clobber what we *think* should the output directory.
   // If someone built a tar.gz with an incorrect directory structure, this check wouldn't save us.
   if (fs::Exists(expected_directory)) {
     return error::Internal("Refusing to clobber existing directory: $0.", expected_directory);
   }
```
etep, line 507: Does it make sense to move this check into the enclosing context? Maybe it should say, "not attempting to install linux headers because target directory already exists." It seems strange to invoke our "extract" method and then hit this, i.e. especially now that the extractor is targeting a staging dir.

src/stirling/utils/linux_headers.cc
```diff
@@ -498,30 +498,26 @@ Status LinkHostLinuxHeaders(const std::filesystem::path& lib_modules_dir) {
   return Status::OK();
 }
 
-Status ExtractPackagedHeaders(PackagedLinuxHeadersSpec* headers_package) {
-  std::string expected_directory =
-      absl::Substitute("/usr/src/linux-headers-$0.$1.$2-pl", headers_package->version.version,
-                       headers_package->version.major_rev, headers_package->version.minor_rev);
-
+Status ExtractPackagedHeaders(PackagedLinuxHeadersSpec* headers_package,
```
etep, line 501: `const` the string args.

src/stirling/utils/linux_headers.cc
```diff
@@ -627,9 +623,18 @@ Status InstallPackagedLinuxHeaders(const std::filesystem::path& lib_modules_dir)
   PX_ASSIGN_OR_RETURN(PackagedLinuxHeadersSpec packaged_headers,
                       FindClosestPackagedLinuxHeaders(kPackagedHeadersRoot, kernel_version));
   LOG(INFO) << absl::Substitute("Using packaged header: $0", packaged_headers.path.string());
```
etep, line 625: Please add a comment here (or nearby) explaining the flow of events, I think it is something like this:
1. verify the expected target dir. does not exist (see comment about moving this check into its enclosing context).
2. extract to a staging dir., but with the target dir. prefix stripped.
3. attempt to do the rest of the flow (patches, whatever).
4. move the staging dir. to the target dir.

src/stirling/utils/linux_headers.cc
benkilimnik: Switched to absl::StrCat!

src/stirling/utils/linux_headers.cc
benkilimnik: I agree. I've moved it into the enclosing context i.e. into `InstallPackagedLinuxHeaders`

src/stirling/utils/linux_headers.cc
benkilimnik: Added some comments!

src/stirling/utils/linux_headers.cc
```diff
@@ -627,9 +620,30 @@ Status InstallPackagedLinuxHeaders(const std::filesystem::path& lib_modules_dir)
...
+  if (fs::Exists(expected_directory)) {
+    return error::Internal(
+        "Not attempting to install packaged headers because the target directory already exists: "
+        "$0",
+        expected_directory);
+  }
+  // Extract the packaged headers to a staging directory, stripping the expected target directory
+  // prefix.
+  PX_RETURN_IF_ERROR(
+      ExtractPackagedHeaders(&packaged_headers, staging_directory, expected_directory));
```
etep, line 639: It seems like there is a side effect in `ExtractPackagedHeaders`. I wonder if we can pass in the "headers package path" and the "staging dir" as separate and immutable args., and then below here just use `staging_directory` on lines 640, 642, and 645.

benkilimnik: Great point! I have passed  `staging_directory` around instead of mutating `packaged_headers.path`. Does that address your comment?

src/stirling/utils/linux_headers.cc
```diff
@@ -498,31 +498,21 @@ Status LinkHostLinuxHeaders(const std::filesystem::path& lib_modules_dir) {
...
-                       headers_package->version.major_rev, headers_package->version.minor_rev);
-
-  // This is a loose check that we don't clobber what we *think* should the output directory.
-  // If someone built a tar.gz with an incorrect directory structure, this check wouldn't save us.
-  if (fs::Exists(expected_directory)) {
-    return error::Internal("Refusing to clobber existing directory: $0.", expected_directory);
-  }
-
-  // Extract the files.
+Status ExtractPackagedHeaders(const PackagedLinuxHeadersSpec* headers_package,
```
etep, line 501: I think it is possible to remove the arg. `const PackagedLinuxHeadersSpec* headers_package` and just use the two other args. Should make these args. (though) either `const std::string&` or `std::string_view`.

benkilimnik: I think the minitar constructor needs to take the path to the tarball at `headers_package->path`, so we have to pass it into `ExtractPackagedHeaders` for this line to work properly.   
```
::px::tools::Minitar minitar(headers_package->path.string());
```
Does that make sense?

etep: Ah. Makes sense.

etep (approved):

src/stirling/utils/linux_headers.cc
etep: We should make that arg a const ref though.

zasgar: should be:

`const PackagedLinuxHeadersSpec& headers_package`

zasgar (changes requested):

zasgar (approved):

benkilimnik:
Accidentally pushed a commit relating to a different issue a few minutes ago - reverted now. No changes since last approval.

PR #1668, opened 2023-08-11
Use VizierID to find the Vizier's Org 

kpattaswamy:
Summary: This PR allows the VizierID to be used when getting the Vizier's org as opposed to only relying on the DeployKey when adding feature flags. This PR also contains changes to prevent a segmentation fault that occurred previously.

Relevant Issues: https://github.com/pixie-io/pixie/issues/1632

Type of change: /kind bug


src/cloud/config_manager/controllers/server.go
```diff
@@ -217,12 +219,37 @@ func (s *Server) GetConfigForVizier(ctx context.Context,
...
-	} else {
+		log.WithError(err).Error("Error getting org ID from deploy key")
+	}
+	if orgID == uuid.Nil && in.VizierID != "" {
+		// Generate token to use to get the org information from Vizier.
+		claims := srvutils.GenerateJWTForService("vzmgr Service", viper.GetString("domain_name"))
+		token, err := srvutils.SignJWTClaims(claims, viper.GetString("jwt_signing_key"))
+		if err != nil {
+			return nil, err
+		}
```
vihangm, line 234: use `getServiceCredentials` here, see `getOrgIDForDeployKey` for an example.

src/cloud/config_manager/controllers/server.go
```diff
@@ -217,12 +219,37 @@ func (s *Server) GetConfigForVizier(ctx context.Context,
...
+
+		ctx = metadata.AppendToOutgoingContext(context.Background(), "authorization",
+			fmt.Sprintf("bearer %s", token))
+
+		resp, err := s.vzmgrClient.GetOrgFromVizier(ctx, utils.ProtoFromUUIDStrOrNil(in.VizierID))
+		if err != nil {
+			log.WithError(err).Error("Error getting the org ID from Vizier")
+		} else {
+			orgID = utils.UUIDFromProtoOrNil(resp.OrgID)
+		}
```
vihangm, line 244: IMO it's worth pulling this out into `getOrgIDForVizier` similar to how `getOrgIDForDeployKey` is a helper func.

vihangm (approved):

PR #1669, opened 2023-08-11
Restore ability to disable tracing for protocols via PEM flags

### LINKED ISSUE #1614: Inconsistencies with documented protocol tracing PEM flags and potential backwards compatibility issues
## Background

The `px` cli and `helm` install support a concept called PEM flags. This provides a mechanism for specifying configuration settings to override default PEM behavior. Despite being labeled a "flag", it actually is a mechanism for injecting environment variables on the vizier-pem DaemonSet. These environment variables are consumed on the PEM side by specifying a gflag that reads its value from an environment variable ([source](https://github.com/pixie-io/pixie/blob/95807eed1029a5a6a239e5e3ffde3a4917be2c1c/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc#L148-L150)). One thing to note about the flag defined below is that the `gflags` library requires the environment variable to match the specified type.

```
DEFINE_uint32(datastream_buffer_retention_size,
              gflags::Uint32FromEnv("PL_DATASTREAM_BUFFER_SIZE", 1024 * 1024),
              "The maximum size of a data stream buffer retained between cycles.");
```

Our [documentation's](https://docs.px.dev/reference/admin/deploy-options#set-a-pem-flag) example use case is for enabling AMQP protocol tracing via the `STIRLING_ENABLE_AMQP_TRACING` flag. This can be accomplished with the following `px deploy` or `helm` command:
```
$ px deploy --pem_flags="STIRLING_ENABLE_AMQP_TRACING=true"
$ helm install pixie pixie-operator/pixie-operator-chart --set dataCollectorParams.customPEMFlags.STIRLING_ENABLE_AMQP_TRACING=true
```

Since that documentation was written, the `STIRLING_ENABLE_AMQP_TRACING` flag was removed on the stirling side (meaning `gflags` does not look for an env variable anymore). In addition to that, the [stirling_enable_amqp_tracing flag](https://github.com/pixie-io/pixie/blob/95807eed1029a5a6a239e5e3ffde3a4917be2c1c/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc#L101-L102) is no longer a boolean. The type was changed to specify what protocols should be enabled only for newer kernels (43c3bc2d110f7fdc3cde7e55569ddf2aaa38e6ee). This creates a problem where a previously defined PEM flag could cause a type error if we were to reintroduce support for opting out of protocol tracing on a per protocol basis.

Aside from the backwards compatibility concern, disabling specific protocols from tracing is requested by end users and often useful when debugging (or side stepping) PEM bugs. End users that want to optimize their PEM's memory usage often want to disable protocols they don't use.

## Solutions

At the bare minimum, we should ensure that our documentation is consistent with what is currently supported (removing examples that are incorrect). In addition to that, I believe we should restore the ability to disable tracing of specific protocols.

benkilimnik:
Summary: Restores ability to disable tracing of specific protocols via PEM flags (i.e. environment variable injection). Also changes the flag name listed in the [docs](https://docs.px.dev/reference/admin/deploy-options#set-a-pem-flag) for backwards compatibility, as a previously defined PEM flag could cause a type error.

Disabling specific protocols from tracing is requested by end users (e.g. for optimizing memory usage) and often useful when debugging (or side stepping) PEM bugs.

Related issues: Fixes #1614

Type of change: /kind feature

Related documentation change: https://github.com/pixie-io/docs.px.dev/pull/274

Test Plan: Verified that tracing is still on by default for each protocol and that tracing still runs as expected via the `socket_tracer` tests.

A further test is required to ensure that the enum `TraceMode::OnForNewerKernel` is set correctly for `mux` tracing in older kernels. Due to permission issues, I can't currently run this `QEMU` test:

```
bazel test -c opt --config=qemu-bpf --test_output=all  --cache_test_results=no --remote_download_outputs=all //src/stirling/source_connectors/socket_tracer:mux_trace_bpf_test --test_output=streamed --//bazel/test_runners/qemu_with_kernel:kernel_version=oldest
```

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
+++ b/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
@@ -78,27 +78,39 @@ DEFINE_string(socket_trace_data_events_output_path, "",
 // Due to BPF instruction limits (< 4096 instructions) on kernels older than
 // 5.2, we can't simultaneously enable all protocols. Thus, some protocols
 // are only enabled on newer kernels.
-DEFINE_int32(stirling_enable_http_tracing, px::stirling::TraceMode::On,
+DEFINE_int32(stirling_enable_http_tracing,
+             gflags::Int32FromEnv("PX_STIRLING_ENABLE_HTTP_TRACING", px::stirling::TraceMode::On),
              "If true, stirling will trace and process HTTP messages");
-DEFINE_int32(stirling_enable_http2_tracing, px::stirling::TraceMode::On,
+DEFINE_int32(stirling_enable_http2_tracing,
+             gflags::Int32FromEnv("PX_STIRLING_ENABLE_HTTP2_TRACING", px::stirling::TraceMode::On),
              "If true, stirling will trace and process gRPC RPCs.");
-DEFINE_int32(stirling_enable_mysql_tracing, px::stirling::TraceMode::On,
+DEFINE_int32(stirling_enable_mysql_tracing,
+             gflags::Int32FromEnv("PX_STIRLING_ENABLE_MYSQL_TRACING", px::stirling::TraceMode::On),
              "If true, stirling will trace and process MySQL messages.");
-DEFINE_int32(stirling_enable_pgsql_tracing, px::stirling::TraceMode::On,
+DEFINE_int32(stirling_enable_pgsql_tracing,
+             gflags::Int32FromEnv("PX_STIRLING_ENABLE_PGSQL_TRACING", px::stirling::TraceMode::On),
              "If true, stirling will trace and process PostgreSQL messages.");
-DEFINE_int32(stirling_enable_cass_tracing, px::stirling::TraceMode::On,
+DEFINE_int32(stirling_enable_cass_tracing,
+             gflags::Int32FromEnv("PX_STIRLING_ENABLE_CASS_TRACING", px::stirling::TraceMode::On),
              "If true, stirling will trace and process Cassandra messages.");
-DEFINE_int32(stirling_enable_dns_tracing, px::stirling::TraceMode::On,
+DEFINE_int32(stirling_enable_dns_tracing,
+             gflags::Int32FromEnv("PX_STIRLING_ENABLE_DNS_TRACING", px::stirling::TraceMode::On),
              "If true, stirling will trace and process DNS messages.");
-DEFINE_int32(stirling_enable_redis_tracing, px::stirling::TraceMode::On,
+DEFINE_int32(stirling_enable_redis_tracing,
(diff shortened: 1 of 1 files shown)


ddelnano:
When we discussed the `mux_trace_bpf_test` verification steps, I forgot that the test is managing the flags internally to allow it to work on older kernels. I made the following change and ran the qemu test to verify that mux tracing should not be enabled on older kernels:
```
diff --git a/src/stirling/source_connectors/socket_tracer/mux_trace_bpf_test.cc b/src/stirling/source_connectors/socket_tracer/mux_trace_bpf_test.cc
index 9ef2b5455d..28146823fe 100644
--- a/src/stirling/source_connectors/socket_tracer/mux_trace_bpf_test.cc
+++ b/src/stirling/source_connectors/socket_tracer/mux_trace_bpf_test.cc
@@ -61,12 +61,12 @@ using ::testing::MatchesRegex;
 // (SocketTraceBPFTest will already have been constructed).
 bool Init() {
   // Make sure Mux tracing is enabled.
-  FLAGS_stirling_enable_mux_tracing = true;
+  /* FLAGS_stirling_enable_mux_tracing = true; */

   // We turn off CQL and NATS tracing to give some BPF instructions back for Mux.
   // This is required for older kernels with only 4096 BPF instructions.
-  FLAGS_stirling_enable_cass_tracing = false;
-  FLAGS_stirling_enable_nats_tracing = false;
+  /* FLAGS_stirling_enable_cass_tracing = true; */
+  /* FLAGS_stirling_enable_nats_tracing = true; */

ddelnano@vigenere:~/code/pixie-worktree (pem-flags) $ bazel test -c opt --config=qemu-bpf --test_output=all  --cache_test_results=no --remote_download_outputs=all //src/stirling/source_connectors/socket_tracer:mux_trace_bpf_test --test_output=streamed --//bazel/test_runners/qemu_with_kernel:kernel_version=oldest
[ .. ]
src/stirling/source_connectors/socket_tracer/mux_trace_bpf_test.cc:151: Failure
  tablets.size()
    Which is: 0
  1
I20230811 22:35:20.994000   227 container_runner.cc:53] podman rm -f thriftmux_server_62972361690 &>/dev/null
[  FAILED  ] MuxTraceTest.Capture (88106 ms)
[----------] 1 test from MuxTraceTest (88108 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (88109 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] MuxTraceTest.Capture

```

ddelnano (approved):

vihangm (approved):

PR #1670, opened 2023-08-11
Let the Operator get the VizierID and send to Config Manager Service

### LINKED ISSUE #1632: Vizier feature flags are set only at deploy time

kpattaswamy:
Summary: This PR lets the operator get the VizierID from the cluster's secrets and then sends it to the config manager service for it to ultimately set feature flags.

Relevant Issues: Fixes #1632

Type of change: /kind bug

### DIFF
--- a/src/operator/controllers/vizier_controller.go
+++ b/src/operator/controllers/vizier_controller.go
@@ -47,11 +47,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"px.dev/pixie/src/api/proto/cloudpb"
+	"px.dev/pixie/src/api/proto/uuidpb"
 	"px.dev/pixie/src/api/proto/vizierconfigpb"
 	"px.dev/pixie/src/operator/apis/px.dev/v1alpha1"
 	version "px.dev/pixie/src/shared/goversion"
 	"px.dev/pixie/src/shared/services"
 	"px.dev/pixie/src/shared/status"
+	"px.dev/pixie/src/utils"
 	"px.dev/pixie/src/utils/shared/certs"
 	"px.dev/pixie/src/utils/shared/k8s"
 )
@@ -809,7 +811,7 @@ func convertResourceType(originalLst v1.ResourceList) *vizierconfigpb.ResourceLi
 
 // generateVizierYAMLsConfig is responsible retrieving a yaml map of configurations from
 // Pixie Cloud.
-func generateVizierYAMLsConfig(ctx context.Context, ns string, k8sVersion string, vizierID string, vz *v1alpha1.Vizier, conn *grpc.ClientConn) (*cloudpb.ConfigForVizierResponse,
+func generateVizierYAMLsConfig(ctx context.Context, ns string, k8sVersion string, vizierID *uuidpb.UUID, vz *v1alpha1.Vizier, conn *grpc.ClientConn) (*cloudpb.ConfigForVizierResponse,
 	error) {
 	client := cloudpb.NewConfigServiceClient(conn)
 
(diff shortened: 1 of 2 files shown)


kpattaswamy:
Changing this to a draft until the other PR with the cloud changes gets in.

src/operator/controllers/vizier_controller.go
```diff
@@ -1123,6 +1130,32 @@ func getClusterUID(clientset *kubernetes.Clientset) (string, error) {
 	return string(ksNS.UID), nil
 }
 
+// getVizierID gets the ID of the cluster the Vizier is in.
+func getVizierID(clientset *kubernetes.Clientset, namespace string) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
```
JamesMBartlett, line 1135: Instead of implementing the retry logic yourself, I would use the `backoff` library. We use it in a number of places. For example: https://github.com/pixie-io/pixie/blob/main/src/vizier/services/cloud_connector/bridge/server.go\#L330

kpattaswamy: Restructured it to include that

JamesMBartlett (approved):

PR #1675, opened 2023-08-15
[stirling] Remove hardcoded path in and cleanup logic in `linux_headers.cc`.

etep:
Summary: In `linux_headers.cc`, we remove use of a hard-coded path to `/host` which may not be available in every Pixie deployment. In addition, we clean up the logic inside of `linux_headers.cc` to make the algorithm more robust and readable.

Type of change: /kind cleanup

Test Plan: Existing tests. Also deployed to ubuntu & cos images on GKE, observed headers found/installed for both.


src/common/base/status.h
```diff
@@ -147,6 +147,15 @@ inline Status StatusAdapter<px::statuspb::Status>(const px::statuspb::Status& s)
 
 }  // namespace px
 
+// Other possible variations:
+// PX_RETURN_VALUE_IF_OK
+// PX_RETURN_STATUS_OR_IF_OK
```
ddelnano, line 152: Are these suggestions for other possible names for this macro? They seem to articulate slightly different behavior and seems like something that can be removed.

ddelnano (approved):
One minor comment but otherwise this lgtm.

src/common/base/status.h
etep: Those are suggestions for different macros with different behavior, but agreed about removing. Doing that now.

src/stirling/utils/linux_headers.cc
```diff
@@ -420,81 +420,76 @@ Status ApplyConfigPatches(const std::filesystem::path& linux_headers_base) {
...
+
+  if (build_path_exists && source_path_exists) {
+    LOG(INFO) << absl::Substitute("Using Linux headers from: $0 and $1.",
+                                  lib_modules_build_dir.string(), lib_modules_source_dir.string());
+    return Status::OK();
+  } else if (build_path_exists) {
+    LOG(INFO) << absl::Substitute("Using Linux headers from: $0.", lib_modules_build_dir.string());
+    return Status::OK();
+  }
+  return error::NotFound("Could not found 'source' or 'build' under $0.", lib_modules_dir.string());
```
JamesMBartlett, line 438: nit: find

src/common/base/status.h
```diff
@@ -147,6 +147,12 @@ inline Status StatusAdapter<px::statuspb::Status>(const px::statuspb::Status& s)
 
 }  // namespace px
 
+#define PX_RETURN_STATUS_OK_IF_OK(__materialized_status, __status_gen) \
```
JamesMBartlett, line 150: imo this macro is more confusing than it's worth to save a couple lines, but if you really want to use it than you should wrap it in a do while false like the others

JamesMBartlett (approved):
Some minor feedback

src/stirling/utils/linux_headers.cc
etep: This is my favorite nit ever. On it.

src/common/base/status.h
etep: There was a long standing TODO in the previous version of `linux_headers.cc` and I agreed fully with the insight behind that, however, given that this macro isn't used elsewhere, would be a good middle ground to just re-locate the macro into `linux_headers.cc`? We can pull it into `base` later if we want to.

etep: Also, realized we don't want the braces for this macro.

PR #1676, opened 2023-08-15
[planner] Enable partial aggregates

### LINKED ISSUE #1440: [carnot] Implement execution of partial aggregates.
**Describe the Task**
We already have support for planning queries with partial aggregates (Partial aggregates means some of the aggregation is done on the agent nodes before sending the partially aggregated results to Kelvin for final aggregation). However, the query execution engine doesn't yet support executing partial aggregate nodes. This issue tracks the implementation on the execution side.

**Subtasks**
- [x] Add UDADefinition accessors for Serialize/Deserialize methods so that the execution side can access them.
- [x] Fix the compiler's output schema for partial aggregates. Currently, the compiler outputs a single column for all partial aggregates, but its much more convenient from the execution's point of view for there to be one column for each UDA executed.
- [x] Implement the deserialization, merging, and serialization in the agg node.
- [x] Finally, enable partial aggregates on the planner side.

JamesMBartlett:
Summary: Flips the switch to enable partial aggregates. This should improve query performance.

Relevant Issues: Fixes #1440

Type of change: /kind feature

Test Plan: Tested by skaffolding to a cluster. Saw that `px/cluster` and other scripts return reasonable results. Compared the node CPU % from `px/cluster` (calculated through an aggregate) to the actual usage on the node. Also, used `px:set explain=true` `px:set analyze=true` to check that the query plan was using partial aggregates.

Changelog Message:
```release-note
Queries now support partial aggregates which should improve query performance significantly in some cases.
```


src/carnot/planner/distributed/coordinator/coordinator.cc
```diff
@@ -180,7 +180,7 @@ StatusOr<SchemaToAgentsMap> LoadSchemaMap(
 StatusOr<std::unique_ptr<DistributedPlan>> CoordinatorImpl::CoordinateImpl(const IR* logical_plan) {
   // TODO(zasgar) set support_partial_agg to true to enable partial aggs.
```
vihangm, line 181: rm? :)

vihangm (approved):

PR #1677, opened 2023-08-16
Change VizierID type from string to uuidpb.UUID

kpattaswamy:
Summary: This PR adjusts the cloud side to accept a type uuidpb.UUID for the VizierID instead of a string.

Relevant Issues: #1632

Type of change: /kind bug


JamesMBartlett (approved):
The VizierID field isn't set by any callers of the `GetConfigForVizier` yet, correct?

kpattaswamy:
@JamesMBartlett That's right, the operator would fill the VizierID field before sending it but that code isn't in yet

Issue #1638, opened 2023-07-26
Capturing Short Running Processes


AdityaAtulTewari:
**Describe the bug**
In many cases, Pixie fails to report HTTP events for short-lived processes, including for short-lived Kubernetes Pods and for short-lived child-processes within longer-living Pods. This is because Pixie's data collector Stirling fails to collect the necessary process metadata. This metadata is used to reconcile communication traces with their source/destination pod(s).

Currently, running processes have their process id (pid) communicated to Stirling by each data capture ebpf probe. In order to properly associate this pid with a kubernetes pod, a lookup procedure in the Stirling's userspace periodically probes Linux's sysfs in order to collect all pids in each cgroup, along with that cgroup's metadata. The procedure to associate a pid with a kubernetes pod does a lookup using the pid to find the cgroup containing the process. The cgroup path name is related to the pod unique id (pod uid) and they are able to match them using regex (code [here](https://github.com/pixie-io/pixie/blob/main/src/shared/metadata/cgroup_path_resolver.cc)). However pids of short-lived processes will leave the cgroup once the process terminates. Thus if the process terminates before a lookup occurs, then the data for that pid cannot be exported from stirling.

**To Reproduce**
1. Deploy pixie
2. Start a container and ssh into it
3. Run `curl example.com`
4. Check Pixie Dashboard and you'll see that the behavior was not captured in the http_data script.

**Expected behavior**
The Behavior of Short Running processes should be captured. There are several ways to achieve this but this is the desired high level behavior.

**App information (please complete the following information):**
This is an issue discussed with the Pixie team, and impacts pixie in all its current versions.

**Additional Information**
This is a known issue to the Pixie Team, and I am working with them to fix it.
[Here](https://docs.google.com/document/d/1bz5nsmM5yJFNhmr6oohAJnkjnFbHSzakKA_df4zxj7g/edit) is a link to the evolving design document.

AdityaAtulTewari:
Please view [this](https://github.com/AdityaAtulTewari/pixie/tree/trigger_slp_error/trigger_slp_error) pixie branch that has a mechanism to deploy a bug triggering test.

PR #1681, opened 2023-08-18
Added cgid capture to stirling with testing

AdityaAtulTewari:
Summary: Added Capturing of Cgroup ids to the ebpf probes in the socket tracer. This includes modifications to data structures for communicating that information to userspace and other data tables. It also includes some minimal testing for ensuring that cgroup ids are captured correctly. The reason for this change is given in the design document for the solution to this [issue](https://github.com/pixie-io/pixie/issues/1638): in short it is the first step in using cgids rather than pids to associate a particular piece of communication data with the correct pod.

Test Plan: Added Cgroup id capture checks to existing `socket_tracer` tests.

Type of change: /kind Feature expansion

Signed-off-as: AdityaAtulTewari [adityaatewari@gmail.com](mailto:adityaatewari@gmail.com)

### DIFF
--- a/src/shared/metadata/cgroup_path_resolver.cc
+++ b/src/shared/metadata/cgroup_path_resolver.cc
@@ -26,6 +26,8 @@
 
 #include <fstream>
 #include "src/common/fs/fs_wrapper.h"
+#include "src/common/system/config.h"
+#include "src/common/system/proc_pid_path.h"
 #include "src/shared/metadata/cgroup_path_resolver.h"
 
 DEFINE_bool(force_cgroup2_mode, true, "Flag to force assume cgroup2 fs for testing purposes");
@@ -78,7 +80,7 @@ StatusOr<std::string> FindSelfCGroupProcs(std::string_view base_path) {
 }
 
 StatusOr<uint64_t> FindCgroupIDFromPID(pid_t pid) {
-  std::string cgroup_name = absl::StrCat("/proc/", pid, "/cgroup");
+  auto cgroup_name = px::system::ProcPidPath(pid, "cgroup");
   std::ifstream proc_cgroup(cgroup_name);
 
   // get rid of the id at the front
@@ -94,7 +96,9 @@ StatusOr<uint64_t> FindCgroupIDFromPID(pid_t pid) {
     return error::NotFound(failure_string);
   }
 
-  std::string cgroup_path_name = absl::StrCat("/sys/fs/cgroup/", line.substr(2, line.size()));
+  std::string cgroup_path_name =
+      absl::StrCat(px::system::Config::GetInstance().sysfs_path().string(), "/cgroup/",
+                   line.substr(2, line.size()));
 
   // lstat what we are pointing to
   struct stat stat_buf;
--- a/src/shared/metadata/cgroup_path_resolver.cc
+++ b/src/shared/metadata/cgroup_path_resolver.cc
@@ -24,6 +24,8 @@
 
 #include <absl/strings/str_replace.h>
 
+#include <stdio.h>
+#include <fstream>
 #include "src/common/fs/fs_wrapper.h"
 #include "src/shared/metadata/cgroup_path_resolver.h"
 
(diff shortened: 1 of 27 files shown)


ddelnano:
@pixie-io/maintainers in the meantime would we be able to let the CI jobs run? @AdityaAtulTewari is interested in seeing if there are any failures to address.

src/shared/metadata/cgroup_path_resolver.cc
```diff
@@ -76,6 +77,35 @@ StatusOr<std::string> FindSelfCGroupProcs(std::string_view base_path) {
...
+  std::getline(proc_cgroup, line);
+  if (proc_cgroup.fail() || proc_cgroup.bad()) {
+    std::ostringstream failure_string;
+    failure_string << "procfs cgroup file getline failed with"
+                   << "fail: " << proc_cgroup.fail() << "bad : " << proc_cgroup.bad()
+                   << "eof: " << proc_cgroup.eof();
+    return error::NotFound(failure_string.str());
+  }
+  std::ostringstream oss;
+  oss << "/sys/fs/cgroup/" << line.substr(2, line.size());
```
vihangm, line 98: ditto on absl::Substitute usage here.

src/shared/metadata/cgroup_path_resolver.cc
```diff
@@ -76,6 +78,35 @@ StatusOr<std::string> FindSelfCGroupProcs(std::string_view base_path) {
   return error::NotFound("Could not find self as a template.");
 }
 
+StatusOr<uint64_t> FindCgroupIDFromPID(pid_t pid) {
+  std::ostringstream cgroup_name;
+  cgroup_name << "/proc/" << pid << "/cgroup";
```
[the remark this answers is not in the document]
AdityaAtulTewari: I used StrCat it seemed more natural I hope that is appropriate.

src/shared/metadata/cgroup_path_resolver.cc
AdityaAtulTewari: ditto for StrCat

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -1686,6 +1698,7 @@ void SocketTraceConnector::TransferConnStats(ConnectorContext* ctx, DataTable* d
 
       r.Append<idx::kTime>(time);
       r.Append<idx::kUPID>(upid.value());
+      r.Append<idx::kCGID>(UINT64_MAX);
```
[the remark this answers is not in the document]
AdityaAtulTewari: Not populating the table causes a template compilation error iirc. Happy to discuss other failure mechanisms.

AdityaAtulTewari:
> We had previously discussed writing a trace bpf test that would exhibit this short lived process behavior. I think it's important to develop that before we make the change to store the cgroup ids (this PR). Essentially my thoughts are that we would take your containerized application from [this branch](https://github.com/AdityaAtulTewari/pixie/tree/be0bcdd827116122da4f8f8ef6d6dd5fe2fac328/trigger_slp_error), create a new trace bpf test with it and then follow it up with this change to verify it results in the records we expect.

This change does not fix the race condition, it just exposes the cgroup id so that a future pr can. The changes are broken up this way as it allows me to break apart the solution to this problem into easier to design pieces. I'm happy to add that test to the change in progress that fixes the issue. However I don't think it is appropriate here because it will still fail.

src/stirling/source_connectors/socket_tracer/bcc_bpf_intf/common.h
```diff
@@ -79,6 +79,8 @@ struct conn_id_t {
   int32_t fd;
   // Unique id of the conn_id (timestamp).
   uint64_t tsid;
+  // TODO(AdityaAtulTewari) This belongs in the Upid struct but that creates a rewrite.
```
ddelnano, line 82: I think we should remove this TODO. There are definitely valid reasons for making the cgid more of a useful construct for the rest of Pixie, but I think that specific solution is going to require a more detailed proposal and discussion.

src/stirling/source_connectors/socket_tracer/conn_stats.h
```diff
@@ -75,6 +75,7 @@ class ConnStats {
     uint64_t conn_close = 0;
     uint64_t bytes_sent = 0;
     uint64_t bytes_recv = 0;
+    uint64_t cgid_curr = 0;
```
ddelnano, line 78: Minor comment: My understanding is that this should never change. So it seems strange to label it as "current". I would prefer this to be just `cgid`.

ddelnano (commented):
Two more additional comments on my latest round of review.

src/stirling/source_connectors/socket_tracer/conn_stats.h
AdityaAtulTewari: This can change if the process migrates cgroups. While this is not possible explicitly in kubernetes nothing stops someone from making this behavior materialize in an actual linux program, within kubernetes.

ddelnano: Thanks for calling that out. I was vaguely aware of that, but didn't make the connection at first.

ddelnano (approved):
The test suite still needs to pass, but the changes look good to me pending that 👍

We will need to regenerate the data table docs with the following 
 tool: `//src/carnot/docstring/tables:datatables`. Happy to help with that after this is merged.

ddelnano:
cc @etep since I believe you are interested in reviewing this as well.

etep (approved):

vihangm (approved):

Issue #1659, opened 2023-08-08
Scope bpftrace/PxL scripts based on kernel version


benkilimnik:
**Is your feature request related to a problem? Please describe.**
As the kernel changes, certain PxL scripts can break over time, particularly those relying on kprobes. For example,`kprobe:tcp_drop` was removed in v5.19, breaking the `tcp_drops` script (see #1582). A user may be running a heterogenous cluster with some nodes running older kernels, so simply updating the script for newer kernels would not provide full coverage. 

**Describe the solution you'd like**
It would be convenient if Pixie could scope scripts to nodes running certain versions of the kernel. This might require that:

- The metadata service tracks the kernel version for each PEM
- When the script is executed, the querybroker retrieves kernel versions from the metadata service
- When the querybroker sends the kprobe/tracepoint to each PEM, it uses the appropriate function depending on the node’s kernel version.

**Describe alternatives you've considered**
Keep around legacy versions of scripts that work on nodes running older kernel versions.

Issue #1582, opened 2023-06-27
`px/tcp_drops` pxl script broken due to missing tcp_drop kprobe


ddelnano:
Running the `px/tcp_drops` pxl script results in the following errors:

```
$ kubectl -n pl logs vizier-pem-qrn9l

[ .. ]

stdin:26-28: WARNING: tcp_drop is not traceable (either non-existing, inlined, or marked as "notrace"); attaching to it will likely fail
I20230627 17:32:19.979256 1894345 stirling.cc:549] DynamicTraceConnector [DT_e2a6ba86-df73-4975-a291-fd8728296097] created in 5570.15 ms.
I20230627 17:32:19.979339 1894345 source_connector.cc:35] Initializing source connector: DT_e2a6ba86-df73-4975-a291-fd8728296097
cannot attach kprobe, probe entry may not exist
ERROR: Error attaching probe: 'kprobe:tcp_drop'
I20230627 17:32:20.119061 1894345 stirling.cc:566] Internal : Failed to run BPF code.
I20230627 17:32:22.009603 3168351 tracepoint_manager.cc:180] Tracepoint[e2a6ba86-df73-4975-a291-fd8728296097]::tcp_drop_tracer has transitioned PENDING_STATE -> FAILED_STATE
```

It appears that the `tcp_drop` function was removed in v5.19 kernel and that removal may have been backported to other v5.x releases. For these newer kernels there is a tracepoint available that should serve a similar purpose (`tracepoint:skb:kfree_skb`). BCC's `tcpdrop` tool uses this as a [fallback](https://github.com/iovisor/bcc/blob/6ef0f0dd2c8d315e2b6dda26e63d014a49ca9ec1/tools/tcpdrop.py#L228-L232) if the kprobe doesn't exist, so we should implement a similar fallback mechanism.

ddelnano:
This isn't quite complete yet so I'm reopening. We need to merge the remaining PRs (#1686, #1687 and #1688) in order to call this done.

PR #1684, opened 2023-08-28
[Kernel scoping 1/5] Move kernel version utils to src/common/system

benkilimnik:
Summary: Moves utils related to finding the kernel version out of `src/stirling/utils` to `common/system`, so we can add the kernel version as a property of `HostInfo` during agent registration. This is the first step towards fixing #1582.

Related issues: #1582 #1659

Type of change: /kind cleanup

Test Plan: Tested all existing targets - `bazel test ...`


src/stirling/utils/linux_headers_test.cc
```diff
@@ -28,116 +29,13 @@ namespace px {
...
-  KernelVersion kernel_version = kernel_version_status.ValueOrDie();
-
-  // We don't know on what host this test will run, so we don't know what the version code will be.
-  // But we can put some bounds, to check for obvious screw-ups.
-  // We assume test will run on a Linux machine with kernel 3.x.x or higher,
-  // and version 9.x.x or lower.
-  // Yes, we're being very generous here, but we want this test to pass the test of time.
-  EXPECT_GE(kernel_version.code(), 0x030000);
-  EXPECT_LE(kernel_version.code(), 0x090000);
-}
```
ddelnano, line 89: [src/stirling/utils/BUILD.bazel](https://github.com/pixie-io/pixie/blob/1eb7873ed8263f0223bef1b87ff20c0b35432d7b/src/stirling/utils/BUILD.bazel#L78-L80) is still enabled to use full qemu emulation, but it seems like we moved the VDSO specific test to `src/common/system`. Did you investigate if we could remove that tag from here?

src/common/system/testdata/sample_host_ubuntu/proc/version_signature
```diff
@@ -0,0 +1 @@
+Ubuntu 5.4.0-52.57-generic 5.4.65
```
ddelnano, line 1: I believe the `src/stirling/utils/testdata/sample_host_{ubuntu,debian}` directories can be removed.

ddelnano (commented):
Two minor comments but otherwise looks good.

src/stirling/utils/linux_headers_test.cc
benkilimnik: I believe full qemu emulation is unneeded here. I've removed the tag in the most recent commit.

ddelnano (approved):

pixie-io-buildbot (dismissed):

JamesMBartlett (approved):

PR #1685, opened 2023-08-28
[Kernel scoping 2/5] Add kernel version to HostInfo

### LINKED ISSUE #1582: `px/tcp_drops` pxl script broken due to missing tcp_drop kprobe

benkilimnik:
Summary: Adds kernel version to `HostInfo` during PEM registration. This is used to fix #1582 by scoping tracepoint deployments to nodes running certain versions of the kernel.

Related issues: #1582 #1659

Type of change: /kind feature

Test Plan: Tested all existing targets and added tests to `src/vizier/services/agent/shared/manager/registration_test.cc` and `src/vizier/services/agent/pem/pem_manager_test.cc`


src/vizier/services/agent/kelvin/kelvin_manager.h
```diff
@@ -43,9 +44,11 @@ class KelvinManager : public Manager {
   KelvinManager() = delete;
   KelvinManager(sole::uuid agent_id, std::string_view pod_name, std::string_view host_ip,
                 std::string_view addr, int grpc_server_port, std::string_view nats_url,
-                std::string_view mds_url)
+                std::string_view mds_url, px::system::KernelVersion kernel_version)
+      // TODO(@benkilimnik): Kernel version may not be needed in kelvin, only in PEMManager for
+      // script selection based on HostInfo. Could use dummy value or refactor Manager class.
```
etep, line 49: Agreed that Kelvin may not need kernel version, at least for the purposes we have right now. Is it possible to remove from Kelvin (or is there some common base class where we add kernel version as an arg. making it required as an arg. here)?

benkilimnik: There is a common base class `Manager` in `vizier/services/agent/shared/manager/manager.{h,cc}` to which I've added the kernel version.

I think we could handle this in a number of ways:

1. Set a dummy value for the kernel version in `KelvinManager`. 

2. Set a flag in `Manager`
```cpp
protected:
    Manager(..., bool use_kernel_version = true);
```
```cpp
Manager::Manager(..., bool use_kernel_version) {
    if (use_kernel_version) {
        // Find and set the kernel version
    }
    // other initialization
}
```
```cpp
KelvinManager::KelvinManager(...) 
    : Manager(..., false) { // Pass false to indicate that we don't need to go find the kernel version
    // KelvinManager specific initialization
}
```

3. Refactor `Manager` such that it does not require the kernel version in its constructor, but provides a separate method to set it. Derived classes like `PEMManager` set the kernel version after construction if needed.

4. Make the kernel version optional in the base Manager class e.g. `std::optional<px::system::KernelVersion>`

ddelnano: I have another proposal for addressing this. What if we mutated the `info_` member's kernel_version field within the PEMManager's constructor body? This would allow us to remove any references of kernel version from the Manager base class and the KelvinManager.

This commit (https://github.com/pixie-io/pixie/commit/179598858ff9ee551bafcc1d36122e5b380bca2b) shows an example of what I mean. If we like that direction, I think we should also add tests to ensure that a KelvinManager and PEMManager have an `info_` member in the correct state after construction: former having kernel_version uninitialized and the latter having it set to a kernel version.

I see this as an improvement to option 3 because it ensures that a newly minted PEMManager is properly configured and reduces risk of using the API incorrectly (forgetting to call the kernel version setter).

ddelnano (commented):
A few minor comments in addition to the existing discussion about how to handle the kernel version for KelvinManagers.

src/vizier/services/agent/shared/manager/registration.cc
```diff
@@ -20,8 +20,10 @@
 #include <utility>
 #include <vector>
 
+#include "src/vizier/services/agent/shared/base/info.h"
 #include "src/vizier/services/agent/shared/manager/manager.h"
 #include "src/vizier/services/agent/shared/manager/registration.h"
+#include "src/vizier/services/metadata/metadatapb/service.grpc.pb.h"
```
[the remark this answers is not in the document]
benkilimnik: This is a leftover from an old change - I've removed it in the most recent commit. Thanks for catching this.

src/vizier/services/agent/kelvin/kelvin_manager.h
benkilimnik: That seems like a more robust solution to me.

Regarding testing, are you imagining something like this, where we check that the kernel version is initialized after construction? (I think the member function info() is protected, so we'd need a getter or friend test function). 

```cpp
class PEMManagerTest : public ::testing::Test {
    protected:
        PEMManagerTest() {
            agent_info_ = agent::Info{};
            agent_info_.agent_id = sole::uuid4();
            agent_info_.hostname = "hostname";
            agent_info_.address = "address";
            agent_info_.pod_name = "pod_name";
            agent_info_.host_ip = "host_ip";
            agent_info_.kernel_version =
                system::ParseKernelVersionString("5.15.0-106-generic").ValueOrDie();
        }

    agent::Info agent_info_;
};

TEST_F(PEMManagerTest, ConstructorTest) {
    auto manager =
      PEMManager::Create(agent_info_.agent_id, agent_info_.pod_name, agent_info_.host_ip,
                         "nats_url", agent_info_.kernel_version)
          .ConsumeValueOrDie();  // Remove the semicolon before .ConsumeValueOrDie()
    EXPECT_EQ(manager->info()->agent_id, agent_info_.agent_id);
    EXPECT_EQ(manager->info()->pod_name, agent_info_.pod_name);
    EXPECT_EQ(manager->info()->host_ip, agent_info_.host_ip);
    EXPECT_EQ(manager->info()->kernel_version, agent_info_.kernel_version);
}
```

ddelnano: @benkilimnik that's what I was thinking and I think making it a friend test function sounds good.

benkilimnik: Because of the way `PEMManager` inherits from `Manager` I have actually found it rather difficult to write a test case for this. When I instantiate the `PEMManager` I get this error

```cpp
F20230908 07:05:26.144451    12 statusor.h:148] Check failed: _s.ok() Bad Status: Internal : Failed to read file ../../services/certs/ca.crt (No such file or directory)
```

which I believe comes from `SSL::DefaultGRPCClientCreds()` in `manager.cc:103`

```cpp
Manager::Manager(sole::uuid agent_id, std::string_view pod_name, std::string_view host_ip,
                 int grpc_server_port, services::shared::agent::AgentCapabilities capabilities,
                 services::shared::agent::AgentParameters parameters, std::string_view nats_url,
                 std::string_view mds_url)
    : grpc_channel_creds_(SSL::DefaultGRPCClientCreds()),
...
```

I think a test that mocks out parts of the `Manager` constructor will require some design changes. As I see it, we could:
1. Move initialization logic out of the `Manager` constructor, so that we can avoid calling it in tests and only instantiate the object
2. Instead of directly creating instances / reading files inside the constructor, we could pass them as parameters, so that we can pass mock versions in tests.
3. Instead of inheriting from Manager, we could make `PEMManager` and `Manager` implement the same interface. Then, PEMManager can have a Manager member that it delegates to for the shared functionality. This way, we could mock out the entire Manager when testing PEMManager.

@ddelnano @etep eager to hear your thoughts on this.

benkilimnik: After some discussion with @ddelnano, I ended up adding a private constructor to `PEMManager` that contains a flag to skip the `Init()` of the base `Manager`. This flag is set to `false` in the public constructor and `true` in friend tests which call it.
The cert error was fixed by setting `FLAGS_disable_SSL = true;`

src/vizier/services/agent/kelvin/kelvin_main.cc
```diff
@@ -27,6 +27,7 @@
 #include "src/common/base/base.h"
 #include "src/common/event/nats.h"
 #include "src/common/signal/signal.h"
+#include "src/common/system/kernel_version.h"
```
ddelnano, line 30: This appears to be unused.

src/vizier/services/agent/shared/manager/manager.cc
```diff
@@ -33,6 +33,7 @@
 #include "src/common/base/base.h"
 #include "src/common/metrics/metrics.h"
 #include "src/common/perf/perf.h"
+#include "src/common/system/kernel_version.h"
```
ddelnano, line 36: This also appears unused (compiling locally works).

ddelnano (commented):
To more minor comments. Assuming @etep doesn't have any issues with the direction of how we tested this, this lgtm.

src/vizier/services/agent/pem/pem_manager_test.cc
```diff
@@ -0,0 +1,60 @@
...
+
+DECLARE_bool(disable_SSL);
+
+namespace px {
+namespace vizier {
+namespace agent {
+
+class PEMManagerTest : public ::testing::Test {
+ protected:
+  // void TearDown() override { dispatcher_->Exit(); }
```
ddelnano, line 32: @benkilimnik did you intend to have this code uncommented or can this be removed?

benkilimnik: Just removed

src/vizier/services/agent/kelvin/kelvin_manager.h
etep: I'd be ok with option (5) keep the version in Kelvin if it nets out to "simpler." It appears we already have a "way" to avoid going here, but is the pain worth it? I am ok with either route btw.

ddelnano: Despite advocating for the current direction, I do think that having the kernel version for both is likely simpler.

Sorry for causing additional change @benkilimnik, but after reconsidering (5) is probably best. If the Manager class was an area where future test coverage would helpful that might tip the scale, but as we've seen it's very difficult to exercise its code.

src/vizier/services/agent/pem/pem_manager.h
```diff
@@ -73,7 +77,9 @@ class PEMManager : public Manager {
                                .Help("Total amount of memory on the node (includes inuse and free "
                                      "memory). Corresponds to /proc/meminfo MemTotal.")
                                .Register(GetMetricsRegistry())
-                               .Add({})) {}
+                               .Add({})) {
+    info()->kernel_version = kernel_version;
+  }
```
ddelnano, line 82: We can remove this since we went back to setting it within `Manager::Manager`.

ddelnano (approved):
One more minor change, but approving since we have consensus on the direction.

etep (approved):

JamesMBartlett (approved):

PR #1686, opened 2023-08-28
[Kernel scoping 3/5] Add selectors to tracepoint deployment

benkilimnik:
Summary: Adds selectors to `TracepointDeployment` and modifies query broker to deploy tracepoints only to agents with matching selectors. Currently intended to fix #1582, but can capture other restrictions (OS version etc.) in the future.

Related issues: #1582 #1659

Type of change: /kind feature

Test Plan: Tested all existing targets and tested selection based on kernel version in `src/vizier/services/metadata/controllers/tracepoint/tracepoint_test.go`

PR #1687, opened 2023-08-28
[Kernel scoping 4/5] Add TraceProgram object

### LINKED ISSUE #1659: Scope bpftrace/PxL scripts based on kernel version

benkilimnik:
Summary: Adds new `TraceProgram` object to `pxtrace` which accepts a BPFtrace program and selectors to specify deployment restrictions as key value arguments. Also modifies the `probe_fn` argument to `pxtrace.UpsertTracepoint` to accept a single `TraceProgram` object or a list of them, extracting selectors and populating `TracepointDeployment` as required, so that the query broker can register tracepoints with agents that match those selectors.

Example usage

```
import pxtrace
import px

before_518_trace_program = pxtrace.TraceProgram(
  program="""$0""",
  max_kernel='5.18',
)
after_519_trace_program = pxtrace.TraceProgram(
  program="""$1""",
  min_kernel='5.19',
)

table_name = 'tcp_drop_table'
pxtrace.UpsertTracepoint('tcp_drop_tracer',
                          table_name,
                          [before_518_trace_program, after_519_trace_program],
                          pxtrace.kprobe(),
                          '10m')
)pxl";
```

Related issues: #1582. Closes #1659.

Type of change: /kind feature

Test Plan: Tested all existing targets, added compiler tests to `src/carnot/planner/logical_planner_test.cc` and `src/carnot/planner/probes/probes_test.cc`, and skaffolded changes to a cluster, verifying expected behavior and interoperability with selectors introduced in PR #1686.

Known issues
- Two or more programs on one PEM: If selectors are specified such that two or more bpftrace programs are deployed to a PEM as part of a single `TracepointDeployment`, stirling throws an error `stirling.cc:511` `error::Internal("Only one Tracepoint is currently supported.");`. This is an existing limitation, which is a little more exposed as a result of the selector feature introduced in this PR. To mitigate this, we could add documentation, improve the error message, and/or have just one of the bpftrace programs in the `TracepointDeployment` run.
- Schemas differ: If the schemas of two bpftrace programs passed to one `pxtrace.UpsertTracepoint` call differ, only one of the schemas will be used to the create the output table (currently the first one that is processed). When the program whose schema differs tries to push a record to the table, we get the error `source_connector.cc:64] Failed to push data. Message = Table_id ## doesn't exist.` The user has to manually create another table/make another call to `UpsertTracepoint` if they want to deploy bpftrace programs with different schemas.

### DIFF
--- a/src/carnot/planner/probes/probes_test.cc
+++ b/src/carnot/planner/probes/probes_test.cc
@@ -706,10 +706,7 @@ TEST_F(ProbeCompilerTest, parse_single_bpftrace_program_object) {
   ASSERT_EQ(pb.mutations_size(), 1);
 
   std::string literal_bpf_trace_min = kBPFTraceProgramMinKernel;
-  literal_bpf_trace_min =
-      std::regex_replace(literal_bpf_trace_min, std::regex(R"(\\\n)"), R"(\\\\n)");
   literal_bpf_trace_min = std::regex_replace(literal_bpf_trace_min, std::regex("\n"), "\\n");
-  literal_bpf_trace_min = std::regex_replace(literal_bpf_trace_min, std::regex("\""), "\\\"");
 
   EXPECT_THAT(pb.mutations()[0].trace(),
               testing::proto::EqualsProto(
@@ -775,16 +772,10 @@ TEST_F(ProbeCompilerTest, parse_multiple_bpftrace_program_objects) {
   ASSERT_EQ(pb.mutations_size(), 1);
 
   std::string literal_bpf_trace_min = kBPFTraceProgramMinKernel;
-  literal_bpf_trace_min =
-      std::regex_replace(literal_bpf_trace_min, std::regex(R"(\\\n)"), R"(\\\\n)");
   literal_bpf_trace_min = std::regex_replace(literal_bpf_trace_min, std::regex("\n"), "\\n");
-  literal_bpf_trace_min = std::regex_replace(literal_bpf_trace_min, std::regex("\""), "\\\"");
 
   std::string literal_bpf_trace_max = kBPFTraceProgramMaxKernel;
-  literal_bpf_trace_max =
-      std::regex_replace(literal_bpf_trace_max, std::regex(R"(\\\n)"), R"(\\\\n)");
   literal_bpf_trace_max = std::regex_replace(literal_bpf_trace_max, std::regex("\n"), "\\n");
-  literal_bpf_trace_max = std::regex_replace(literal_bpf_trace_max, std::regex("\""), "\\\"");
 
   EXPECT_THAT(pb.mutations()[0].trace(),
               testing::proto::EqualsProto(absl::Substitute(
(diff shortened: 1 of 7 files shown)


src/carnot/planner/probes/probes_test.cc
```diff
@@ -648,6 +648,228 @@ TEST_F(ProbeCompilerTest, parse_bpftrace) {
...
+                                                           kBPFTraceProgramMinKernel)));
+  plannerpb::CompileMutationsResponse pb;
+  EXPECT_OK(probe_ir->ToProto(&pb));
+  ASSERT_EQ(pb.mutations_size(), 1);
+
+  std::string literal_bpf_trace_min = kBPFTraceProgramMinKernel;
+  literal_bpf_trace_min =
+      std::regex_replace(literal_bpf_trace_min, std::regex(R"(\\\n)"), R"(\\\\n)");
+  literal_bpf_trace_min = std::regex_replace(literal_bpf_trace_min, std::regex("\n"), "\\n");
+  literal_bpf_trace_min = std::regex_replace(literal_bpf_trace_min, std::regex("\""), "\\\"");
```
ddelnano, line 712: Do we have to use regular expressions? It seems the other tests use `absl::Substitute` and imo it's much more readable.

ddelnano (commented):
Just one comment about the use of regular expressions, but otherwise lgtm!

src/carnot/planner/probes/probes_test.cc
benkilimnik: Thanks for pointing that out - I've removed a few calls to `regex_replace` which I believe are unneeded.

I think `absl::Substitute` is used for parameterized string substitution rather than pattern-based substitution which we're doing here. The existing test case for bpftrace strings on lines 635-649 also makes use of `regex_replace`, so I simply copied those over. They were [added in this commit.](https://github.com/benkilimnik/pixie-privy/commit/1593bcb2d31eaff8eb5d8b207d7ad574f0696b17)

Another option would be to use [absl::StrReplaceAll](https://github.com/abseil/abseil-cpp/blob/master/absl/strings/str_replace.h#L194) but I ran into the `error: no member named 'StrReplaceAll' in namespace 'absl'`

ddelnano (approved):

JamesMBartlett (approved):

PR #1689, opened 2023-08-28
[Fix CQL Stitcher 1/4] Change StitchFrames API for protocols with streams only

benkilimnik:
Summary: This is the first of four PRs aimed at fixing an issue with the CQL stitcher discussed in #1375. It introduces an additional interface for `StitchFrames` for protocols with streams:
```cpp
// interface for protocols with streams
StitchFrames(absl::flat_hash_map<stream_id_t, std::deque<Frame>>* req_messages, absl::flat_hash_map<stream_id_t, std::deque<Frame>>* res_messages)
```
The old interface is retained for other protocols by default. To use the new map interface, protocols must have the `StreamSupport` enum set to `UseStream` in their `ProtocolTraits`.

```cpp
// interface for protocols without a notion of streams i.e. which don't use streams for frame stitching
StitchFrames(std::deque<Frame>* req_messages, std::deque<Frame>* res_messages)
```
This change should be a no op, as the default interface is used across the board. The next PR has the CQL stitcher and tests actually use it.

Note that in PRs 1/4 and 2/4 the map is populated after parsing frames into request/response deques. PRs 3/4 and 4/4 populate the map earlier during parsing and has protocols without streams use the first key i.e. a single stream. This is a larger change and therefore separated.

Related issues: #1375

Type of change: /kind bug

Test Plan: Tested all existing targets - `bazel test ...`

Performance: I've benchmarked the performance impact of this API change using a [new demo](https://github.com/benkilimnik/pixie-privy/tree/cql-perf-test) that cycles through streams. The demo uses a [fork of the datastax driver for cassandra](https://github.com/benkilimnik/python-driver) which I tinkered with to prevent stream ID reuse. This demo goes all the way up to `2^15-1`, the absolute worst case in terms of stream ID reuse. Note that this is probably a very unrealistic scenario (by default, this particular driver only goes up to 300 streams max), but it helps us test the limits of the map interface. As a sanity check, I also ran an experiment for the existing k8ssandra demo.

Note that I only ran one experiment for each of k8ssandra and the bad-stream-reuse demos. More perf tests may be required to reliably assess performance impact, but this at least rules out a major degradation. Also, since this change is currently scoped to only the cassandra stitcher, it shouldn't impact any of the other protocols.

### Python demo poor stream ID reuse

Baseline
baseline_poor_stream_reuse

API change (including changes to CQL Stitcher to make use of new interface PR 2/3 #1715): streams up to `2^15-1`
map_interface_poor_stream_reuse

### K8ssandra demo
Baseline
baseline_k8ssandra

API change (including changes to CQL Stitcher to make use of new interface in PR 2/3 #1715)
map_interface_k8ssandra

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/conn_tracker.h
+++ b/src/stirling/source_connectors/socket_tracer/conn_tracker.h
@@ -274,16 +274,26 @@ class ConnTracker : NotCopyMoveable {
     protocols::RecordsWithErrorCount<TRecordType> result;
 
     // If this protocol doesn't support streams, we call StitchFrames with just the deque.
-    // If it does, we populate a map of stream ID to deque.
+    // If it does, we use a map of stream ID to deque.
     if constexpr (TProtocolTraits::stream_support ==
                   protocols::BaseProtocolTraits<TRecordType>::UseStream) {
       using TKey = typename TProtocolTraits::key_type;
       std::map<TKey, std::deque<TFrameType>> requests;
       std::map<TKey, std::deque<TFrameType>> responses;
+      // TODO(@benkilimnik): Hard code the stream for now. Populate the map in a future PR.
       requests[0] = std::move(req_frames);
       responses[0] = std::move(resp_frames);
       result = protocols::StitchFrames<TRecordType, TKey, TFrameType, TStateType>(
           &requests, &responses, state_ptr);
+      // TODO(@benkilimnik): Update req and resp frame deques to match maps for now. Populate maps
+      // during parsing in a future PR.
+      req_frames.clear();
+      resp_frames.clear();
+      for (auto& [_, frames] : requests) {
+        for (auto& frame : frames) {
+          req_frames.push_back(frame);
+        }
+      }
     } else {
       result = protocols::StitchFrames<TRecordType, TFrameType, TStateType>(
           &req_frames, &resp_frames, state_ptr);
--- a/src/stirling/source_connectors/socket_tracer/conn_tracker.h
+++ b/src/stirling/source_connectors/socket_tracer/conn_tracker.h
@@ -276,6 +276,7 @@ class ConnTracker : NotCopyMoveable {
 
     // If this protocol doesn't support streams, we call StitchFrames with just the deque.
     // If it does, we use a map of stream ID to deque.
+    // TODO(@benkilimnik): Eventually, we should migrate all of the protocols to use the map.
     if constexpr (TProtocolTraits::stream_support ==
                   protocols::BaseProtocolTraits<TRecordType>::UseStream) {
       using TKey = typename TProtocolTraits::key_type;
--- a/src/stirling/source_connectors/socket_tracer/conn_tracker.h
+++ b/src/stirling/source_connectors/socket_tracer/conn_tracker.h
@@ -294,6 +294,11 @@ class ConnTracker : NotCopyMoveable {
           req_frames.push_back(frame);
         }
       }
+      for (auto& [_, frames] : responses) {
+        for (auto& frame : frames) {
+          resp_frames.push_back(frame);
+        }
+      }
     } else {
       result = protocols::StitchFrames<TRecordType, TFrameType, TStateType>(
           &req_frames, &resp_frames, state_ptr);
--- a/src/stirling/source_connectors/socket_tracer/conn_tracker.h
+++ b/src/stirling/source_connectors/socket_tracer/conn_tracker.h
@@ -291,12 +291,12 @@ class ConnTracker : NotCopyMoveable {
       resp_frames.clear();
       for (auto& [_, frames] : requests) {
         for (auto& frame : frames) {
-          req_frames.push_back(frame);
+          req_frames.push_back(std::move(frame));
         }
       }
       for (auto& [_, frames] : responses) {
         for (auto& frame : frames) {
-          resp_frames.push_back(frame);
+          resp_frames.push_back(std::move(frame));
         }
       }
     } else {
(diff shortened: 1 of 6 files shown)


src/stirling/source_connectors/socket_tracer/conn_tracker.h
```diff
@@ -271,9 +271,33 @@ class ConnTracker : NotCopyMoveable {
...
+      std::map<TKey, std::deque<TFrameType>> responses;
+      // TODO(@benkilimnik): Hard code the stream for now. Populate the map in a future PR.
+      requests[0] = std::move(req_frames);
+      responses[0] = std::move(resp_frames);
+      result = protocols::StitchFrames<TRecordType, TKey, TFrameType, TStateType>(
+          &requests, &responses, state_ptr);
+      // TODO(@benkilimnik): Update req and resp frame deques to match maps for now. Populate maps
+      // during parsing in a future PR.
+      req_frames.clear();
+      resp_frames.clear();
```
ddelnano, line 291: Can you explain why these should be cleared? I believe this will cause us to purge frames that should persist between stitching rounds. I don't know that CQL has a test for that, but I know that was something we saw in the pgsql tests ([background](https://github.com/pixie-io/pixie/issues/697#issuecomment-1411120312)).

src/stirling/source_connectors/socket_tracer/conn_tracker.h
```diff
@@ -271,9 +271,33 @@ class ConnTracker : NotCopyMoveable {
...
+      for (auto& [_, frames] : requests) {
+        for (auto& frame : frames) {
+          req_frames.push_back(frame);
+        }
+      }
```
ddelnano, line 296: Was it intentional to only copy the frames from the request map?

src/stirling/source_connectors/socket_tracer/conn_tracker.h
benkilimnik: In lines 284-285 we're transferring ownership of the resources managed by `req_frames` to `requests[0]` and `resp_frames` to `responses[0]`. This leaves `req_frames` in a 'moved-from' state. While it's still a valid object, its state is now unspecified. So, when `protocols::StitchFrames` is called and modifies the deque referenced by `requests[0]`, those changes won't be seen in `req_frames` because `req_frames` has transferred its resources away. I think it is generally not safe to make assumptions about the state of moved-from objects, so I've added these two calls to clear the deques before they are re-populated using the map that was passed in to `StitchFrames`.

Please note that this code is intended to be a stopgap solution until PR 3/3 gets in, which populates the map earlier during parsing.

src/stirling/source_connectors/socket_tracer/conn_tracker.h
benkilimnik: For the CQL stitcher and a few other protocols, the response deque is cleared at the end of `StitchFrames`, so I don't think we need to copy the frames from the responses map. In this PR, only CQL has been set to use the map interface.

You're right that this might be fail if other protocols used the map, though this should be resolved once 3/3 gets in and the maps are populated earlier during parsing.

src/stirling/source_connectors/socket_tracer/conn_tracker.h
ddelnano: ah, I see. I wasn't sure how the move would work in this case. After looking at the [move operator docs](https://en.cppreference.com/w/cpp/container/deque/operator%3D) and that [.clear()](https://stackoverflow.com/questions/9168823/reusing-a-moved-container) is safe to call on a container in an unspecified state, I understand now.

Screenshot 2023-10-03 at 1 48 40 PM

src/stirling/source_connectors/socket_tracer/conn_tracker.h
ddelnano: While this code should be short lived, I think it would be best to avoid relying on any implicit behavior within the stitcher. So my preference would be to iterate over both of them.

benkilimnik: Added in most recent commit

src/stirling/source_connectors/socket_tracer/conn_tracker.h
```diff
@@ -271,9 +271,38 @@ class ConnTracker : NotCopyMoveable {
...
+      responses[0] = std::move(resp_frames);
+      result = protocols::StitchFrames<TRecordType, TKey, TFrameType, TStateType>(
+          &requests, &responses, state_ptr);
+      // TODO(@benkilimnik): Update req and resp frame deques to match maps for now. Populate maps
+      // during parsing in a future PR.
+      req_frames.clear();
+      resp_frames.clear();
+      for (auto& [_, frames] : requests) {
+        for (auto& frame : frames) {
+          req_frames.push_back(frame);
```
ddelnano, line 294: Do you know if this will trigger a copy? I'm wondering if this (and the response loop) should provide `std::move(frame)` as the function argument.

benkilimnik: You're right, this will trigger a copy. I've changed it to `std::move(frame)`.

ddelnano (approved):
Changes lgtm now. Please make sure that @etep has a chance to review before the pixie-io/maintainers review.

src/stirling/source_connectors/socket_tracer/conn_tracker.h
```diff
@@ -271,9 +271,38 @@ class ConnTracker : NotCopyMoveable {
...
                                       resp_frames.size());
 
-    protocols::RecordsWithErrorCount<TRecordType> result =
-        protocols::StitchFrames<TRecordType, TFrameType, TStateType>(&req_frames, &resp_frames,
-                                                                     state_ptr);
+    protocols::RecordsWithErrorCount<TRecordType> result;
+
+    // If this protocol doesn't support streams, we call StitchFrames with just the deque.
+    // If it does, we use a map of stream ID to deque.
+    if constexpr (TProtocolTraits::stream_support ==
```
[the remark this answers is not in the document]
benkilimnik: Agreed. I will add a `TODO` to remove the constexpr branch. Migrating all of the protocols to use the map will require further changes which we can address in a future PR.

Regarding the perf bug, that repopulating `TODO` should be closed out in PR 3/3.

etep (approved):

aimichelle (approved):

PR #1693, opened 2023-08-31
[record & replay] Add rr sub-tree and impl.

etep:
Summary: WIP / draft.

Type of change: /kind feature

Test Plan: Adding a test case `rr_bpf_test`.

### DIFF
--- a/src/stirling/bpf_tools/bcc_wrapper.h
+++ b/src/stirling/bpf_tools/bcc_wrapper.h
@@ -499,6 +499,44 @@ class WrappedBCCArrayTableImpl : public WrappedBCCArrayTable<T> {
   std::unique_ptr<U> underlying_;
 };
 
+template <typename T>
+class RecordingWrappedBCCArrayTableImpl : public WrappedBCCArrayTableImpl<T> {
+ public:
+  using Super = WrappedBCCArrayTableImpl<T>;
+
+  StatusOr<T> GetValue(const uint32_t idx) override {
+    PX_ASSIGN_OR_RETURN(const T value, Super::GetValue(idx));
+    recorder_.RecordBPFArrayTableGetValueEvent(this->name_, idx, sizeof(value), &value);
+    return value;
+  }
+
+  RecordingWrappedBCCArrayTableImpl(bpf_tools::BCCWrapper* bcc, const std::string& name)
+      : WrappedBCCArrayTableImpl<T>(bcc, name),
+        recorder_(*bcc->GetBPFRecorder().ConsumeValueOrDie()) {}
+
+ private:
+  BPFRecorder& recorder_;
+};
+
+template <typename T>
+class ReplayingWrappedBCCArrayTableImpl : public WrappedBCCArrayTable<T> {
+ public:
+  StatusOr<T> GetValue(const uint32_t idx) override {
+    T value;
+    PX_RETURN_IF_ERROR(replayer_.ReplayArrayGetValue(this->name_, idx, sizeof(T), &value));
+    return value;
+  }
+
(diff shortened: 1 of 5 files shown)


src/stirling/bpf_tools/rr/rr.cc
```diff
@@ -0,0 +1,372 @@
...
+
+  const auto event_wrapper = events_proto_.event(playback_event_idx_);
+  if (!event_wrapper.has_map_get_table_offline_event()) {
+    return error::Internal("Map event not available.");
+  }
+
+  const auto event = event_wrapper.map_get_table_offline_event();
+
+  if (name != event.name()) {
+    return error::Internal("Mismatched name.");
```
ddelnano, line 292: Having the expected vs actual name in the error string would be helpful for when this does fail.

src/stirling/bpf_tools/bcc_wrapper.h
```diff
@@ -283,13 +298,11 @@ class BCCWrapperImpl : public BCCWrapper {
...
   }
-  void PollPerfBuffers(int timeout_ms = 0) override;
-  void PollPerfBuffer(std::string_view perf_buffer_name, int timeout_ms) override;
+  void PollPerfBuffers(const int timeout_ms = 0) override;
+  Status PollPerfBuffer(const std::string& name, const int timeout_ms = 0) override;
   void Close() override;
 
- protected:
   Status ClosePerfBuffer(const PerfBufferSpec& perf_buffer) override;
-  std::vector<PerfBufferSpec> perf_buffers_;
```
ddelnano, line 292: It seems like these changes aren't specifically related to the record and replay code:
* perf_buffers_ -> perf_buffers_spec rename
* Function signature changes to use `const` and `std::string_view`

Could we separate these out into another PR since the `bcc_wrapper.*` changes are some of the largest? Maybe we can introduce the `CommonPerfBufferSetup` construct without using it as well to further reduce the lines of code?

ddelnano (commented):
Adding a few comments regarding my initial review and our conversation. We also discussed splitting this up into a few PRs:
1. Splitting out the `bcc_wrapper.*` changes that are improvements but not related to record and replay (renaming members, changing function prototypes, etc)
2. The recorder or replayer changes (whichever one is easiest first)
3. Followed by the other BCCWrapper impl (whichever one is deemed easier to do second)

src/stirling/bpf_tools/rr/rr.proto
```diff
@@ -0,0 +1,78 @@
...
+
+message BPFMapGetValueEvent {
+  string name = 1;
+  bytes key = 2;
+  bytes value = 3;
+}
+
+message BPFMapGetTableOfflineEvent {
+  string name = 1;
+  int32 size = 2;
```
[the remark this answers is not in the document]
etep: Added a comment to explain this.

src/stirling/bpf_tools/bcc_wrapper.h
etep: Done in #1697. Thank you!

etep:
> Adding a few comments regarding my initial review and our conversation. We also discussed splitting this up into a few PRs:
> 
>     1. Splitting out the `bcc_wrapper.*` changes that are improvements but not related to record and replay (renaming members, changing function prototypes, etc)
> 
>     2. The recorder or replayer changes (whichever one is easiest first)
> 
>     3. Followed by the other BCCWrapper impl (whichever one is deemed easier to do second)

After landing #1697, I will start splitting this up and pushing pieces out. We can go with "record" then "replay." Alternately, we could do perf buffer, than bpf array, then bpf map. We can discuss and decide that later.

PR #1697, opened 2023-09-07
[record & replay] Tidy up `BCCWrapper` & perf profiler before adding the `rr` sub-tree.

etep:
Summary: We implement a few cosmetic changes to BCC wrapper that support the record & replay effort:
1. We add method `CommonPerfBufferSetup` which will be shared between the recording & normal BCC wrapper impls.
2. We rename the vector of perf buffer specs to `perf_buffer_specs_` i.e. because its inherent name is confusing.
3. We return a status from `PollPerfBuffer` which will be useful for the replaying BCC wrapper.
4. We return BCC as a pointer from BCC wrapper to enable the eventual replaying BCC wrapper.
5. In the perf profiler, we change the data type for the perf buffer names to `std::string` to avoid creating new strings every time they are used.

In response to code review comments, we will rebase #1693 onto this PR, and continue to split that PR into pieces and merge it gradually.

Type of change: /kind feature

Test Plan: Existing test coverage.

### DIFF
--- a/src/stirling/bpf_tools/bcc_wrapper.h
+++ b/src/stirling/bpf_tools/bcc_wrapper.h
@@ -555,10 +555,7 @@ class WrappedBCCStackTableImpl : public WrappedBCCStackTable {
   std::unique_ptr<U> underlying_;
 };
 
-////////////////////////////////////////////////////////////////////////////////////////////////////
-////////////////////////////////////////////////////////////////////////////////////////////////////
 // Creators fns for wrapped maps & arrays:
-
 template <typename BaseT, typename ImplT>
 std::unique_ptr<BaseT> CreateBCCWrappedMapOrArray(BCCWrapper* bcc, const std::string& name) {
   return std::make_unique<ImplT>(bcc, name);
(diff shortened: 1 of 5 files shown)


src/stirling/bpf_tools/bcc_wrapper.h
```diff
@@ -89,7 +89,7 @@ class BCCWrapper {
     return task_struct_offsets_opt_;
   }
 
-  virtual ebpf::BPF& BPF() = 0;
+  virtual ebpf::BPF* BPF() = 0;
```
ddelnano, line 92: What's the rationale for why this change is needed?

ddelnano (approved):
I have one question regarding the `BPF()` function signature change, but otherwise the changes lgtm.

src/stirling/bpf_tools/bcc_wrapper.h
etep: It will be easier to override this way. The replaying BCC wrapper will not have a BPF instance to return. I tried a status-or containing a reference and that is unsupported (if I've got that right...).

zasgar (approved):

Issue #640, opened 2022-11-03
Support for MongoDB query profiling.


rmoyan:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

mayanksingh2298:
I think this would be a great addition to Pixie

PR #1698, opened 2023-09-07
Add support to extract LE Ints in BinaryDecoder & rename the existing ExtractInt function to ExtractBEInt

kpattaswamy:
Summary: This PR adds a function to the BinaryDecoder class to extract little endian encoded integers from the buffer and renames the existing `ExtractInt` function to `ExtractBEInt` to make naming reflect the type of integer. There are several protocol parsing implementations, MongoDB included, which are little endian encoded but have been directly using its relevant functions within utils namespace. Adding this to BinaryDecoder would make it consistent with the existing extract integer functions that deals with big endian encoded integers.

Ran `git grep ExtractInt | cut -d : -f1 | uniq | xargs -I{} sed -i 's/ExtractInt(/ExtractBEInt(/g' {}` and `git grep ExtractInt | cut -d : -f1 | uniq | xargs -I{} sed -i 's/ExtractInt</ExtractBEInt</g' {}` to make the naming changes and reran AMQP's code generation to verify changes were done right.

Related issues: #640

Type of change: /kind feature

Test Plan: `bazel build //...` and `bazel test //...`

### DIFF
--- a/src/stirling/obj_tools/go_syms.cc
+++ b/src/stirling/obj_tools/go_syms.cc
@@ -83,8 +83,8 @@ StatusOr<std::string> ReadGoBuildVersion(ElfReader* elf_reader) {
   BinaryDecoder binary_decoder(buildInfoByteCode);
 
   PX_CHECK_OK(binary_decoder.ExtractStringUntil(kGoBuildInfoMagic));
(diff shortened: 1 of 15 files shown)


ddelnano (approved):

ddelnano:
@kpattaswamy can you update the PR title to reflect that this also renames the existing `ExtractInt` function? I know the name is already long, but I think it's best to include both if possible.

vihangm (approved):

PR #1699, opened 2023-09-08
[socket tracer] Add context to E|D CHECKs in files `parser.cc` and `sticher.cc`.

etep:
Summary: We implement `CTX_ECHECK` and `CTX_DCHECK` inside of the protocols sub-tree of the socket tracer. Previously, if any E or D check fired, the error message would not specify which protocol triggered the problem. This new macro guarantees that the full file path is printed with any failure.

Type of change: /kind feature

Test Plan: Existing test coverage.

### DIFF
--- a/src/stirling/source_connectors/perf_profiler/symbolizers/symbolizer_test.cc
+++ b/src/stirling/source_connectors/perf_profiler/symbolizers/symbolizer_test.cc
@@ -71,7 +71,6 @@ using SymbolizerTypes = ::testing::Types<ElfSymbolizer, BCCSymbolizer>;
 TYPED_TEST_SUITE(SymbolizerTest, SymbolizerTypes);
 
 TYPED_TEST(SymbolizerTest, UserSymbols) {
-  CTX_ECHECK_EQ(0, 1);
   // We will use our self pid for symbolizing symbols from within this process.
   struct upid_t this_upid;
   this_upid.pid = static_cast<uint32_t>(getpid());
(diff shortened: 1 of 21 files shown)


ddelnano (approved):
This approach sounds good to me. In addition I read through the glog docs to see if there was any feature for dealing with this situation and didn't find any remedy.

vihangm (changes requested):
It seems like the OP versions already handle this in glog, but our ECHECK macros do not.
https://github.com/google/glog/blame/bc1fada1cf63ad12aee26847ab9ed4c62cffdcf9/src/glog/logging.h.in#L889
https://github.com/pixie-io/pixie/blob/97d13beb734c8a5be8eacd38757b1b7a1e421caa/src/common/base/logging.h#L62-L68

We should 1) consider upgrading log and 2) update our ECHECK macro instead.
Additionally if we end up needing to add this additional macro, maybe add the line number with `__LINE__` too just like glog does it.

etep:
@vihangm Can you please be more specific about the direction you are advocating?

vihangm:
> @vihangm Can you please be more specific about the direction you are advocating?

Upgrade glog to see if it includes context for DCHECKs. Unless I'm mistaken the macros seem to suggest it should include context.
Update our ECHECK macro to match the CHECK/DCHECH impls so that it also includes context (assuming that step 1 works).

vihangm (approved):

vihangm:
Gotcha, option 2 makes sense.

PR #1702, opened 2023-09-11
Fix and re-enable partial aggregates.

etep:
Summary: The implementation of the "any" UDA was broken for partial aggregates. In this PR, we fix that and re-enable partial aggregates.

Type of change: /kind bug fix

Test Plan: We add two new test cases: `MergeAnyUDA` and `SerDesAnyUDA`.
- [x] Both new test cases fail with the original UDA impl.
- [x] Both new test cases pass with the fixed UDA impl. (this PR).
- [x] Skaffolded up the new impl. and verified that flamegraphs render correctly.

### DIFF
--- a/src/carnot/funcs/builtins/collections.h
+++ b/src/carnot/funcs/builtins/collections.h
@@ -88,7 +88,7 @@ class AnyUDA : public udf::UDA {
   void Merge(FunctionContext*, const AnyUDA& other) { SetValue(other.val_); }
 
   TArg Finalize(FunctionContext*) {
-    DCHECK(picked);
+    DCHECK(picked) << "AnyUDA uninitialized.";
     return val_;
   }
 
@@ -97,7 +97,7 @@ class AnyUDA : public udf::UDA {
   }
 
   StringValue Serialize(FunctionContext*) {
-    DCHECK(picked);
+    DCHECK(picked) << "AnyUDA uninitialized.";
     return SerializeScalar(&val_);
   }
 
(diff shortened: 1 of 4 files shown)


src/carnot/funcs/builtins/collections_test.cc
```diff
@@ -103,7 +153,6 @@ TEST(CollectionsTest, CanSerializeDeserialize_Float) {
   }
 
   auto any_uda = AnyUDA<types::Float64Value>();
-  EXPECT_NE(uda_tester.Result(), any_uda.Finalize(nullptr));
```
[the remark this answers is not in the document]
etep: In response to this comment, we updated this PR to `EXPECT_DEATH` in lieu of the prior expectations. In the original impl., the `AnyUDA` would yield a value, even if it was uninitialized -- essentially, it would provide whatever value was default constructed for the UDA data type. In this PR, we added a `DCHECK` that requires the UDA be initialized before providing a value. This `DCHECK` will fail if we use the original test expectations, hence we change those original test expectations to `EXPECT_DEATH`.

src/carnot/funcs/builtins/collections.h
```diff
@@ -82,26 +82,27 @@ class AnyUDA : public udf::UDA {
...
-      val_ = val;
-      picked = true;
-    }
+    SetValue(val);
   }
 
-  void Merge(FunctionContext*, const AnyUDA&) {
-    // Does nothing.
-  }
+  void Merge(FunctionContext*, const AnyUDA& other) { SetValue(other.val_); }
```
JamesMBartlett, line 88: I think you want `if (other.picked) SetValue...` to avoid setting `picked = true` in the case that both `this` and `other` haven't picked a value yet.

JamesMBartlett (approved):

src/carnot/funcs/builtins/collections.h
etep: Fixed in `5fb0b1e`.

ddelnano (approved):

PR #1704, opened 2023-09-12
[MongoDB Parsing 3/3] Add support to parse MongoDB frames

kpattaswamy:
Summary: This is the final PR for MongoDB protocol parsing. It adds the logic to parse frames and includes corresponding tests. The initial parsing functionality currently supports frames of type `OP_MSG`.

Type of change: /kind feature

Test Plan: Included tests


src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse.cc
```diff
@@ -0,0 +1,95 @@
...
+
+  // Parser will ignore Op Codes that have been deprecated/removed from version 5.0 onwards.
+  if (!(frame_type == Type::kOPMsg || frame_type == Type::kOPCompressed ||
+        frame_type == Type::kReserved)) {
+    return ParseState::kIgnored;
+  }
+
+  ParseState parse_state = mongodb::ProcessPayload(&decoder, frame);
+  if (parse_state == ParseState::kSuccess) {
+    buf->remove_prefix(frame->length);
```
ddelnano, line 74: The binary decoder extract functions will call `remove_prefix` as data is parsed ([source](https://github.com/pixie-io/pixie/blob/2645963f4efb39629d19ec58e06a312e1214d053/src/stirling/utils/binary_decoder.h#L71)). I noticed that the mux implementation is doing the same thing, and I think we moving the buffer forward 2x as much as we should.

Do you know if this was required to get the tests to pass? According to the C++ docs, the behavior is undefined if n is larger than the remaining size of the string view.

Screenshot 2023-09-22 at 2 09 46 PM

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,175 @@
...
+      }
+      remaining_section_length = section.length - kSectionLengthSize - seq_identifier.length() - 1;
+
+    } else {
+      return ParseState::kInvalid;
+    }
+
+    // Extract the document(s) from the section and convert it from type BSON to a JSON string.
+    auto parse_until = decoder->BufSize() - remaining_section_length;
+    while (decoder->BufSize() > parse_until) {
```
ddelnano, line 91: Is this loop necessary? I know the spec says that zero or more BSON objects will be provided. Reading the `bson_new_from_data` function docs, it appears they refer to this as a document and not as an object. It makes me think that the library _might_ be handling the object looping for us.

Screenshot 2023-09-22 at 2 35 44 PM

ddelnano (commented):
This is looking really good. Sorry the review has taken longer than anticipated. I wanted to make sure I was up to speed on the mongo specifics as I reviewed this.

src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse.cc
```diff
@@ -0,0 +1,95 @@
...
+
+  BinaryDecoder decoder(*buf);
+  if (decoder.BufSize() < mongodb::kHeaderLength) {
+    return ParseState::kNeedsMoreData;
+  }
+
+  // Get the length of the packet. This length contains the size of the field containing the
+  // message's length itself.
+  PX_ASSIGN_OR(frame->length, decoder.ExtractLEInt<int32_t>(), return ParseState::kInvalid);
+  if (static_cast<int32_t>(decoder.BufSize()) < (frame->length - mongodb::kMessageLengthSize)) {
```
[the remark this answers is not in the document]
ddelnano:
```suggestion
  if (decoder.BufSize() < size_t(frame->length - mongodb::kMessageLengthSize)) {
```

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,175 @@
...
+
+  // Get the section(s) data from the buffer.
+  while (decoder->BufSize() > checksum_bytes) {
+    mongodb::Section section;
+    PX_ASSIGN_OR(section.kind, decoder->ExtractLEInt<uint8_t>(), return ParseState::kInvalid);
+    // Length of the section still remaining in the buffer.
+    int32_t remaining_section_length = 0;
+
+    if (section.kind == 0) {
+      section.length = utils::LEndianBytesToInt<int32_t, 4>(decoder->Buf());
```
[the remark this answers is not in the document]
kpattaswamy: The reason I didn't use the binary decoder here was because I didn't want to extract the length bytes out of the buffer since the `bson_new_from_data()` function expects the length bytes to be included in the data as well as the expected length of the data

Kind 1 sections have 3 fields (kind, section length, identifier) before the series of documents, so I was able to extract the section length bytes using the binary decoder but still had to use  `utils::LEndianBytesToInt<int32_t, 4>(decoder->Buf())`to find the document length at the time of BSON to JSON conversion

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,175 @@
...
+      if (section.length < kSectionLengthSize) {
+        return ParseState::kInvalid;
+      }
+
+      // Get the sequence identifier (command argument).
+      PX_ASSIGN_OR(std::string_view seq_identifier, decoder->ExtractStringUntil('\0'),
+                   return ParseState::kInvalid);
+      // Make sure the sequence identifier is a valid OP_MSG kind 1 command argument.
+      if (seq_identifier != "documents" && seq_identifier != "updates" &&
+          seq_identifier != "deletes") {
```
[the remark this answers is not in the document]
kpattaswamy: The docs in github link I referenced in the `types.h` file lists out these identifiers, should I add a comment saying to reference that link for more information?

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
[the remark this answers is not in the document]
ddelnano: Cool, that sounds like a valid reason for opting out of consistency. Can we add a comment here to explain that?

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
kpattaswamy: I don't think it is necessary to have it. 

However in the case that there are multiple documents (for kind 1 sections only) the no while loop implementation would successfully parse the n number of documents and store them in one document in the `Section` data structure. Having the loop though would let us parse each document individually and store them separately in Section's document vector. I can see why the code might be a unpleasant when handling this scenario. 

We would still need to potentially truncate the data at the time of making the table so maybe it's not that necessary to keep all the docs separated?

ddelnano: I'm not questioning if we should parse the entire payload (I believe we should). It's just that there seems to be a discrepancy regarding the difference between a BSON object and a BSON document. I'm asking if we are confident that they are one in the same. Because my reading of the docs seems like a single BSON document could contain multiple BSON objects.

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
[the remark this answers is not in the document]
ddelnano: I thought I read the docs carefully enough and I didn't see this. So I think another reference would help. I would specifically use a permalink and make sure it scrolls to the right section (with the `#`) like so:
 https://github.com/mongodb/specifications/blob/e09b41df206f9efaa36ba4c332c47d04ddb7d6d1/source/message/OP_MSG.rst#command-arguments-as-payload

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
kpattaswamy: I believe they are referring to the same data. The github docs and the wire protocol docs refer to the same portion of the section body as documents and objects. 

https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#sections:~:text=0%20or%20more%20BSON%20documents.%20Each%20BSON%20document%20cannot%20be%20larger%20than%20maxBSONObjectSize.

 https://www.mongodb.com/docs/v5.2/reference/mongodb-wire-protocol/#:~:text=Zero%20or%20more%20BSON%20objects

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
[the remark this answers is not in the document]
kpattaswamy: Added the comment in the frame diagram in `types.h`

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,180 @@
...
+  }
+  if ((flag_bits >> 2) & required_unset_bitmask) {
+    return ParseState::kInvalid;
+  }
+
+  // Determine the number of checksum bytes in the buffer.
+  uint8_t checksum_bytes = (frame->checksum_present) ? 4 : 0;
+
+  // Get the section(s) data from the buffer.
+  while (decoder->BufSize() > checksum_bytes) {
```
ddelnano, line 57: I believe this will fail if the buffer has more than a full frame of data. We should continue parse at most `frame->length - checksum_bytes` bytes.

In order to test this out, I added additional data to the payload within `MongoDBParserTest.ParseFrameValidRequest` and the test fails rather than parsing the first complete frame and leaving the buffer with the excess data.

```
$ git diff
diff --git a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test.cc b/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test.cc
index e9ffa072da..14f1ddd733 100644
--- a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test.cc
@@ -249,6 +249,11 @@ constexpr uint8_t mongoDBValidRequest[] = {
     0x39, 0x8b, 0x75, 0x41, 0x97, 0xc4, 0x97, 0x06, 0xd1, 0x00, 0x02,
     0x24, 0x64, 0x62, 0x00, 0x06, 0x00, 0x00, 0x00, 0x6d, 0x79, 0x64,
     0x62, 0x31, 0x00, 0x00,
+    // Bogus data
+    0x64, 0x65, 0x72, 0x65, 0x64, 0x00, 0x01, 0x03, 0x6c, 0x73, 0x69,
+    0x64, 0x00, 0x1e, 0x00, 0x00, 0x00, 0x05, 0x69, 0x64, 0x00, 0x10,
+    0x00, 0x00, 0x00, 0x04, 0xe7, 0xd7, 0x16, 0xb3, 0x75, 0xb7, 0x4c,
+    0x39, 0x8b, 0x75, 0x41, 0x97, 0xc4, 0x97, 0x06, 0xd1, 0x00, 0x02,
 };

 constexpr uint8_t mongoDBValidResponse[] = {
$ bazel test --test_output=all --spawn_strategy=standalone src/stirling/source_connectors/socket_tracer/protocols/mongodb:all --test_arg='--gtest_filter=MongoDBParserTest.ParseFrameValidRequest'
INFO: Invocation ID: 63d705ed-b34c-4853-b17f-372237ba2e7c
INFO: Streaming build results to: https://bb.corp.pixielabs.ai/invocation/63d705ed-b34c-4853-b17f-372237ba2e7c
INFO: Analyzed 3 targets (0 packages loaded, 552 targets configured).
INFO: Found 2 targets and 1 test target...
FAIL: //src/stirling/source_connectors/socket_tracer/protocols/mongodb:parse_test (see /home/ddelnano/.cache/bazel/_bazel_ddelnano/fa8604ca46879b9b1ffd44d337c531f3/execroot/px/bazel-out/k8-fastbuild/testlogs/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test/test.log)
INFO: From Testing //src/stirling/source_connectors/socket_tracer/protocols/mongodb:parse_test:
==================== Test output for //src/stirling/source_connectors/socket_tracer/protocols/mongodb:parse_test:
I20230926 22:28:49.784341 1475921 env.cc:47] Started: /home/ddelnano/.cache/bazel/_bazel_ddelnano/fa8604ca46879b9b1ffd44d337c531f3/execroot/px/bazel-out/k8-fastbuild/bin/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test.runfiles/px/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test
Note: Google Test filter = MongoDBParserTest.ParseFrameValidRequest
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from MongoDBParserTest
[ RUN      ] MongoDBParserTest.ParseFrameValidRequest
src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test.cc:439: Failure
  state
    Which is: 4-byte object <01-00 00-00>
  ParseState::kSuccess
    Which is: 4-byte object <05-00 00-00>
[  FAILED  ] MongoDBParserTest.ParseFrameValidRequest (0 ms)
[----------] 1 test from MongoDBParserTest (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] MongoDBParserTest.ParseFrameValidRequest

 1 FAILED TEST
I20230926 22:28:49.784876 1475921 env.cc:51] Shutting down
================================================================================
INFO: Elapsed time: 0.376s, Critical Path: 0.03s
INFO: 2 processes: 1 internal, 1 local.
//src/stirling/source_connectors/socket_tracer/protocols/mongodb:parse_test FAILED in 0.0s
  /home/ddelnano/.cache/bazel/_bazel_ddelnano/fa8604ca46879b9b1ffd44d337c531f3/execroot/px/bazel-out/k8-fastbuild/testlogs/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse_test/test.log

Executed 1 out of 1 test: 1 fails locally.
INFO: Streaming build results to: https://bb.corp.pixielabs.ai/invocation/63d705ed-b34c-4853-b17f-372237ba2e7c
INFO: Build completed, 1 test FAILED, 2 total actions
```

With the bogus test data appended, the parser should return a valid frame followed by `ParseState::kInvalid`

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,180 @@
...
+        return ParseState::kInvalid;
+      }
+      remaining_section_length = section.length - kSectionLengthSize - seq_identifier.length() - 1;
+
+    } else {
+      return ParseState::kInvalid;
+    }
+
+    // Extract the document(s) from the section and convert it from type BSON to a JSON string.
+    auto parse_until = decoder->BufSize() - remaining_section_length;
```
ddelnano, line 93: I find both of these `while` loops difficult to follow due to the variable naming. The first loop is for processing zero or more sections, while the second is for processing zero or more BSON objects/documents. Would we be able to try to align those high level concepts with the variable naming?

Happy to pair on this to discuss this.

ddelnano (commented):
Thanks for addressing the first round of comments and I've added more feedback on this one. I believe this should be the last round of major feedback.

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,180 @@
...
+  // Bits 0-15 are required and bits 16-31 are optional.
+  if (flag_bits & checksum_bitmask) {
+    frame->checksum_present = true;
+  }
+  if (flag_bits & more_to_come_bitmask) {
+    frame->more_to_come = true;
+  }
+  if (flag_bits & exhaust_allowed_bitmask) {
+    frame->exhaust_allowed = true;
+  }
```
etep, line 48: I believe that the frame is processed in a single shot, i.e. we would not re-enter this code with a pointer to the same frame. It seems that these members: checksum present, more to come, and exhaust allowed, can be set unambiguously here, i.e. we do not need to rely on default values. Expressing a bias here against default values, could you please consider updating to the following, e.g.:
```
frame->checksum_present = (flag_bits & checksum_bitmask) == checksum_bitmask;
```

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,180 @@
...
+  if ((flag_bits >> 2) & required_unset_bitmask) {
+    return ParseState::kInvalid;
+  }
+
+  // Determine the number of checksum bytes in the buffer.
+  uint8_t checksum_bytes = (frame->checksum_present) ? 4 : 0;
```
etep, line 54: Please consider `const`. Also, why is this `uint8_t`? Perhaps its size should match whatever `BufSize()` returns?

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,180 @@
...
+    // Length of the section still remaining in the buffer.
+    int32_t remaining_section_length = 0;
+
+    if (section.kind == 0) {
+      // Check the length but don't extract it since the later logic requires the buffer to retain
+      // it.
+      section.length = utils::LEndianBytesToInt<int32_t, 4>(decoder->Buf());
+      if (section.length < kSectionLengthSize) {
+        return ParseState::kInvalid;
+      }
```
etep, line 69: Is there a maximum section length size that would indicate an invalid parse state?

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
```diff
@@ -0,0 +1,180 @@
...
+      char* json = bson_as_canonical_extended_json(bson_doc, NULL);
+      DEFER(bson_free(json));
+      if (json == NULL) {
+        return ParseState::kInvalid;
+      }
+
+      // Find the type of command argument from the kind 0 section.
+      if (section.kind == 0) {
```
etep, line 123: Not sure if there is any semantic description of the section kind, but if so (i.e. in particular, if so in upstream), then we could have something like this:
```
if (section.kind == kBSONSection)
```

src/stirling/source_connectors/socket_tracer/protocols/mongodb/decode.cc
kpattaswamy: No I don't believe so, but there is a maximum document size the wire protocol mentions

ddelnano:
@kpattaswamy this lgtm from my end. Please ensure @etep has no further comments as well.

ddelnano (approved):

etep (approved):

JamesMBartlett (approved):

PR #1708, opened 2023-09-19
Add rr sub-tree, impl. perf buffer record & replay.

etep:
Summary: We add the `rr` sub-tree to implement BPF record & replay. In this PR, we include _only_ perf buffer record and replay. BPF maps & arrays will be included in a subsequent PR.

Type of change: /kind feature

Test Plan: We add a new test case: `rr_bpf_test`.

### DIFF
--- a/src/stirling/bpf_tools/rr/rr.cc
+++ b/src/stirling/bpf_tools/rr/rr.cc
@@ -80,12 +80,15 @@ void BPFReplayer::ReplayPerfBufferEvents(const PerfBufferSpec& perf_buffer_spec)
   while (playback_event_idx_ < n_events) {
     const auto event = events_proto_.event(playback_event_idx_);
     if (!event.has_perf_buffer_event()) {
+      // Return control to calling context. Replay will continue with a different eBPF data
+      // type, e.g. map or array.
       break;
     }
     const auto perf_buffer_event = event.perf_buffer_event();
     const auto data = perf_buffer_event.data();
     const auto name = perf_buffer_event.name();
     if (name != perf_buffer_spec.name) {
+      // Return control to calling context. Replay will continue with a different perf buffer.
       break;
     }
     const void* const_data = static_cast<const void*>(data.data());
--- a/src/stirling/bpf_tools/rr/rr.cc
+++ b/src/stirling/bpf_tools/rr/rr.cc
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2018- The Pixie Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "src/stirling/bpf_tools/rr/rr.h"
+
+#include <fstream>
+
+namespace px {
+namespace stirling {
+namespace bpf_tools {
+
+void BPFRecorder::WriteProto(const std::string& proto_buf_file_path) {
+  if (!recording_written_) {
+    LOG(INFO) << "Writing BPF events pb to file: " << proto_buf_file_path;
+
+    std::fstream outfile(proto_buf_file_path, std::ios::out | std::ios::trunc | std::ios::binary);
+    if (!outfile.is_open()) {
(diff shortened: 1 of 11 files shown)


src/stirling/bpf_tools/rr/rr_bpf_test.cc
```diff
@@ -0,0 +1,179 @@
...
+    ASSERT_OK(replaying_bcc_->OpenPerfBuffers(replaying_perf_buffer_specs));
+
+    const int64_t self_pid = getpid();
+    const std::filesystem::path self_path = GetSelfPath().ValueOrDie();
+    ASSERT_OK_AND_ASSIGN(auto elf_reader, ElfReader::Create(self_path.string()));
+    ASSERT_OK_AND_ASSIGN(auto converter, ElfAddressConverter::Create(elf_reader.get(), self_pid));
+
+    // For the uprobe spec.
+    const uint64_t foo_virt_addr = reinterpret_cast<uint64_t>(&::test::Foo);
+    const uint64_t bar_virt_addr = reinterpret_cast<uint64_t>(&::test::Bar);
```
ddelnano, line 119: In another place where we follow this similar pattern (probing a symbol within the current executable), we define c linkage ([source](https://github.com/pixie-io/pixie/blob/1e0654e33edaab4978b4b4e419f1e19d178aeeb0/src/stirling/source_connectors/socket_tracer/socket_trace_bpf_tables.cc#L36-L42)). It seems like that must be unnecessary since you were able to avoid declaring that? Do you know why that was done in the previous case?

src/stirling/bpf_tools/rr/testing/bpf/rr_test.c
```diff
@@ -0,0 +1,63 @@
...
+// A minimal BPF program to:
+// 1. Push something into a perf buffer.
+// 2. Populate a BPF array.
+// 3. Populate a BPF hash map.
+int push_something_into_a_perf_buffer(struct pt_regs* ctx) {
+  int count_idx = 0;
+  int* pcount = state.lookup(&count_idx);
+  if (pcount == NULL) {
+    return -1;
+  }
```
ddelnano, line 45: Is this condition not occuring because we are only recording perf buffer events at the moment? In #1693 we don't seem to be populating the state array, and so I don't follow how that test is passing in that PR.

ddelnano (commented):
A few comments and questions, but otherwise looks good.

src/stirling/bpf_tools/rr/rr.cc
```diff
@@ -0,0 +1,114 @@
...
+    return;
+  }
+
+  const auto n_events = events_proto_.event_size();
+
+  while (playback_event_idx_ < n_events) {
+    const auto event = events_proto_.event(playback_event_idx_);
+    if (!event.has_perf_buffer_event()) {
+      break;
+    }
```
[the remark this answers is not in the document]
etep: The recording will have different kinds of events. Therefore, "no" it is not guaranteed that the ... recording will only have perf buffer events.

We do want this to be a `break` statement so that control is returned to the containing context. For example, suppose that the socket tracer is polling perf buffers. This loop will consume all perf buffer events that match the perf buffer name, and then return control to the socket tracer. Next, the socket tracer might read a BPF map.

By breaking, we return control back to the caller when the appropriate perf buffer events are drained, i.e. so that the caller can resume and potentially read from other perf buffers or other BPF data structures.

src/stirling/bpf_tools/rr/rr_bpf_test.cc
etep: I think the c-linkage is not needed in the source code you cite.

src/stirling/bpf_tools/rr/testing/bpf/rr_test.c
etep: This test BPF probe was designed with a few more test cases in mind. It manipulates an array, a map, and a perf buffer (there is a comment about this, but in this PR we are only interested in the perf buffer).

This condition `pcount == NULL` is here to satisfy the eBPF verifier. We don't expect to hit this condition ever, i.e. because the state array has one entry and the index we are looking at is index zero.

src/stirling/bpf_tools/rr/rr.cc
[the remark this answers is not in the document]
etep: I added a comment for this.

src/stirling/bpf_tools/rr/testing/bpf/rr_test.c
ddelnano: Got it. I haven't used bpf array's from scratch, so at a glance I was expecting an array that wasn't explicitly inserted into would have returned `NULL`.

src/stirling/bpf_tools/rr/rr.cc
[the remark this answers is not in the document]
ddelnano: Got it. I see how that works now 👍

ddelnano (approved):

JamesMBartlett (approved):

PR #1709, opened 2023-09-20
fix: add return on err to prevent nil pointer deref panic in goroutine

caphrim007:
Summary: If no execScriptClient is returned due to there being an err, the codebase still attempts to use the client to call execScriptClient.Recv() shortly thereafter. This can result in a runtime panic, which appears to be able to bubble up and crash the QB. This patch returns early if there is a detected error such that the QB will continue processing.

Type of change: /kind fix

Test Plan: Reuse existing unit tests

### DIFF
--- a/src/vizier/services/query_broker/script_runner/script_runner.go
+++ b/src/vizier/services/query_broker/script_runner/script_runner.go
@@ -282,6 +282,7 @@ func (r *runner) runScript(scriptPeriod time.Duration) {
 	})
 	if err != nil {
 		log.WithError(err).Error("Failed to execute cronscript")
+		return
 	}
 	for {
 		resp, err := execScriptClient.Recv()


caphrim007:
some additional notes on the panic and traceback that this seeks to address

```
time="2023-09-20T00:54:06Z" level=error msg="Failed to execute cronscript" 
error="rpc error: code = Unavailable desc = last connection error: connection error: 
desc = \"transport: Error while dialing dial tcp [::1]:50300: connect: connection refused\""

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x47b9342]

goroutine 181 [running]:
px.dev/pixie/src/vizier/services/query_broker/script_runner.(*runner).runScript(0xc000635800, 0xc000508f80?)
	src/vizier/services/query_broker/script_runner/script_runner.go:287 +0x602
px.dev/pixie/src/vizier/services/query_broker/script_runner.(*runner).start.func1()
	src/vizier/services/query_broker/script_runner/script_runner.go:390 +0x91
created by px.dev/pixie/src/vizier/services/query_broker/script_runner.(*runner).start
	src/vizier/services/query_broker/script_runner/script_runner.go:383 +0xd3
```

JamesMBartlett (approved):

PR #1713, opened 2023-09-27
[standalone_pem] Add examples and add option to fully static link.

etep:
Summary: We add some example scripts, a README.md file, and update the build file with an option to fully static link. Static linking may be desirable for deployment to non-containerized environments, but is an experimental feature since it statically links in glibc. We considered static linking with musl, but cannot figure out how to do that with C++.

Type of change: /kind feature

Test Plan: Existing tests. And, copied the standalone pem binary to a GKE node, then deleted all libs, then ran the binary. The binary ran (but could not find linux headers).

### DIFF
--- a/src/experimental/standalone_pem/BUILD.bazel
+++ b/src/experimental/standalone_pem/BUILD.bazel
@@ -14,11 +14,28 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 load("@io_bazel_rules_docker//cc:image.bzl", "cc_image")
 load("@io_bazel_rules_docker//container:container.bzl", "container_image", "container_push")
 load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library")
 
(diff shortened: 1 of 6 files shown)


src/experimental/standalone_pem/BUILD.bazel
```diff
@@ -57,6 +57,7 @@ pl_cc_library(
 pl_cc_binary(
```
ddelnano, line 57: Did you consider linking a different libc like musl? I believe static linking glibc is generally [discouraged](https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged), although I don't have experience with these problems.

JamesMBartlett (changes requested):
Fully statically linking GCC will cause unexpected errors, especially with network calls. If we want to do fully statically linked binaries we have to use musl.

etep:
Agreed, needs to use `musl`. How much scope is there in switching to `musl` ... is that something like a new `cc_toolchain`?

etep:
It is not ok to statically link to glibc. Planning to setup a tool chain to use `musl` for this purpose.

PR #1715, opened 2023-09-29
[Fix CQL Stitcher 2/4] Update StitchFrames to use new interface

benkilimnik:
Summary: Fixes the head of line blocking issue discussed in #1375  by using a map of stream_ids to deque of frames, which we populate in the `conn_tracker`. CQL stitcher tests are updated to reflect the bug fix and new interface.

Related issues: #1375

Type of change: /kind bug

Test Plan: Tested all existing targets and updated the unittests for the CQL stitcher. `Note`: this PR relies on changes introduced in #1689.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
@@ -333,19 +333,16 @@ Status ProcessResp(Frame* resp_frame, Response* resp) {
   }
 }
 
-StatusOr<Record> ProcessReqRespPair(Frame* req_frame, Frame* resp_frame) {
+Status ProcessReqRespPair(Frame* req_frame, Frame* resp_frame, Record* r) {
   CTX_ECHECK_LT(req_frame->timestamp_ns, resp_frame->timestamp_ns);
 
-  Record r;
-  PX_RETURN_IF_ERROR(ProcessReq(req_frame, &r.req));
-  PX_RETURN_IF_ERROR(ProcessResp(resp_frame, &r.resp));
+  PX_RETURN_IF_ERROR(ProcessReq(req_frame, &r->req));
+  PX_RETURN_IF_ERROR(ProcessResp(resp_frame, &r->resp));
 
-  return r;
+  return Status::OK();
 }
 
-StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
-  Record r;
-
+Status ProcessSolitaryResp(Frame* resp_frame, Record* r) {
   // For now, Event is the only supported solitary response.
   // If this ever changes, the code below needs to be adapted.
   CTX_DCHECK(resp_frame->hdr.opcode == Opcode::kEvent);
@@ -353,15 +350,13 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
   // Make a fake request to go along with the response.
   // - Use REGISTER op, since that was what set up the events in the first place.
   // - Use response timestamp, so any calculated latencies are reported as 0.
-  r.req.op = ReqOp::kRegister;
-  r.req.msg = "-";
-  r.req.timestamp_ns = resp_frame->timestamp_ns;
+  r->req.op = ReqOp::kRegister;
+  r->req.msg = "-";
+  r->req.timestamp_ns = resp_frame->timestamp_ns;
 
   // A little inefficient because it will go through a switch statement again,
   // when we actually know the op. But keep it this way for consistency.
-  PX_RETURN_IF_ERROR(ProcessResp(resp_frame, &r.resp));
-
-  return r;
+  return ProcessResp(resp_frame, &r->resp);
 }
 
 // Currently StitchFrames() uses a response-led matching algorithm.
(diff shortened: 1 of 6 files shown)


src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
```diff
@@ -372,77 +372,120 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
...
-    // Search for matching req frame
-    for (auto& req_frame : *req_frames) {
-      if (resp_frame.hdr.stream == req_frame.hdr.stream) {
-        VLOG(2) << absl::Substitute("req_op=$0 msg=$1", magic_enum::enum_name(req_frame.hdr.opcode),
-                                    req_frame.msg);
-
-        StatusOr<Record> record_status = ProcessReqRespPair(&req_frame, &resp_frame);
+  // iterate through all deques of responses associated with a specific streamID and find the
+  // matching request
+  for (auto it = responses->begin(); it != responses->end(); it++) {
```
etep, line 383: Should we unpack this with a range iterator:
```
for (auto& [stream_id, resp_dequeue] : responses) {
```

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
```diff
@@ -372,77 +372,120 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
...
       }
+      continue;
+    }
+
+    // we found a potential set of requests for this stream ID
+    std::deque<cass::Frame>& req_frames = pos->second;
+    std::deque<uint64_t> req_timestamps = std::deque<uint64_t>();
+    // get just the timestamps for matching with responses
+    for (cass::Frame& req_frame : req_frames) {
+      req_timestamps.push_back(req_frame.timestamp_ns);
```
etep, line 419: There should be a way to run the timestamp filtering algorithm without building a separate timestamps dequeue. Also (have not thought about this...) does the timestamp dequeue need to be a dequeue. Can we pre-allocate a vector here?

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
```diff
@@ -372,77 +372,120 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
...
+          std::upper_bound(req_timestamps.begin(), req_timestamps.end(), resp_frame.timestamp_ns) -
+          1;
+      int req_index = stream_it - req_timestamps.begin();
+      // Responses should always have a more recent timestamp than the first request. If this
+      // condition is triggered we should not attempt to match this frame. Since responses are
+      // cleared during StitchFrames this will get cleaned up during the current iteration.
+      if (stream_it + 1 == req_timestamps.begin()) {
+        CTX_DCHECK(false) << "Unable to find request that is earlier than response: "
+                          << resp_frame.ToString();
+        continue;
```
etep, line 439: Is this actually even a `DHCECK` failure? I assume that with data loss any idiosyncratic thing can happen legitimately. Maybe `VLOG` warning?

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
```diff
@@ -372,77 +372,120 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
...
+    // go through the responses for this stream ID and check for requests
+    for (cass::Frame& resp_frame : resp_frames) {
+      latest_resp_ts = resp_frame.timestamp_ns;
+      // This returns the first request timestamp that is JUST BEFORE the response timestamp
+      // Upper bound returns the first request timestamps GREATER than the response timestamp; we
+      // want to get the one right before
+      auto stream_it =
+          std::upper_bound(req_timestamps.begin(), req_timestamps.end(), resp_frame.timestamp_ns) -
+          1;
+      int req_index = stream_it - req_timestamps.begin();
```
etep, line 432: `const size_t`.

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
```diff
@@ -372,77 +372,120 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
...
+      // Responses should always have a more recent timestamp than the first request. If this
+      // condition is triggered we should not attempt to match this frame. Since responses are
+      // cleared during StitchFrames this will get cleaned up during the current iteration.
+      if (stream_it + 1 == req_timestamps.begin()) {
+        CTX_DCHECK(false) << "Unable to find request that is earlier than response: "
+                          << resp_frame.ToString();
+        continue;
+      }
+      cass::Frame& req_frame = req_frames[req_index];
```
etep, line 441: Does this mean we look at only one `req_frame`? I thought if there were multiple request frames with time earlier than the response then we would look at all of those.

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
```diff
@@ -372,77 +372,120 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
...
+    std::deque<cass::Frame>& resp_frames = it->second;
+
+    bool kEventHandled = false;
+    for (cass::Frame& resp_frame : resp_frames) {
+      // Event responses are special: they have no request.
+      if (resp_frame.hdr.opcode == Opcode::kEvent) {
+        kEventHandled = true;
+        StatusOr<Record> record_status = ProcessSolitaryResp(&resp_frame);
         if (record_status.ok()) {
           entries.push_back(record_status.ConsumeValueOrDie());
```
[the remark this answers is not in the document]
benkilimnik: I've modified `ProcessSolitaryResp` to return `Status` and take an out parameter for Record. Then the check becomes:

```cpp
Record record;
Status record_status = ProcessSolitaryResp(&resp_frame, &record);
if (record_status.ok()) {
  entries.push_back(std::move(record));
} else {
  VLOG(1) << record_status.ToString();
  ++error_count;
}
```
This should be clearer.

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
benkilimnik: Good point, I've updated the logic to directly extract the timestamp from req_frames during search via a lambda comparator function passed to `upper_bound`.

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
benkilimnik: Changed to `VLOG`.

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
benkilimnik: The current logic assumes a one-to-one pairing, where each response matches the most recent preceding request. This is because, within a single stream, we expect to match a response with the most recent request.

So for each response frame (in the `resp_deque`), we try to find the most recent corresponding request frame (in `req_deque`) with a timestamp that's just before the response's timestamp.

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
```diff
@@ -372,76 +369,114 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
...
+      // StitchFrames iterations once they bubble up to the front of the deque and form a contiguous
+      // range with the consumed frames. This is done to avoid the bookeeping necessary to delete
+      // multiple ranges of indices in the deque (which occur when responses are lost).
+      if (frame.discarded || frame.timestamp_ns < latest_resp_ts) {
+        error_count++;
+        frame.discarded = true;
+      } else if (!found_unconsumed) {
+        delete_idx = idx;
+        found_unconsumed = true;
         break;
```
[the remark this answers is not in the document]
benkilimnik: Thanks for pointing this out, `found_unconsumed` was leftover from some previous logic.

src/stirling/source_connectors/socket_tracer/protocols/cql/types.h
```diff
@@ -102,6 +103,7 @@ struct Frame : public FrameBase {
   FrameHeader hdr;
   std::string msg;
   bool consumed = false;
+  bool discarded = false;
```
etep, line 106: Do we end up using this member (or did it become not-needed)?

etep (approved):

src/stirling/source_connectors/socket_tracer/protocols/cql/types.h
benkilimnik: Good catch, this is now unneeded. I've just removed it.

src/stirling/source_connectors/socket_tracer/protocols/cql/stitcher.cc
```diff
@@ -372,77 +372,120 @@ StatusOr<Record> ProcessSolitaryResp(Frame* resp_frame) {
...
-        // Found a match, so remove both request and response.
-        // We don't remove request frames on the fly, however,
-        // because it could otherwise cause unnecessary churn/copying in the deque.
-        // This is due to the fact that responses can come out-of-order.
-        // Just mark the request as consumed, and clean-up when they reach the head of the queue.
-        // Note that responses are always head-processed, so they don't require this optimization.
-        found_match = true;
-        req_frame.consumed = true;
-        break;
+    auto pos = requests->find(stream_id);
```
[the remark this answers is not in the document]
benkilimnik: I believe we decided to push this down the road after some discussion about the design changes required in the conn tracker / datastream buffer to make this a map from stream_id to its respective req/resp dequeues like
```absl_flat_hash_map<stream_id_t, std::pair<std::dequeue, std::dequeue>>```

vihangm (approved):

PR #1716, opened 2023-09-29
[Fix CQL Sticher 3/4] Populate map of streams to frames during parsing

### LINKED ISSUE #1375: Large CQL `Result` frames (> 1MiB) cause poor CQL protocol stitching

benkilimnik:
Summary: Modifies all protocol parsers to use a map of streams to deques by default. Protocols which do not have a notion of streams are encoded as single keys in a map. This completes the CQL stitcher fix and should simplify stitching frames for protocols with streams.

The final PR in this sequence #1732 populates a map of streamIDs to deque of frames in ParseFramesLoop instead of ParseFrames. This should provide a small efficiency boost, as we won't have to loop over the frames twice.

Related issues: Closes #1375

Type of change: /kind bug

Test Plan: Updated parsing tests to use new interface. `Note`: this PR relies on changes introduced in #1689 and #1715


src/stirling/source_connectors/socket_tracer/data_stream.h
```diff
@@ -254,22 +270,42 @@ class DataStream : NotCopyMoveable {
...
+        // TODO(yzhao): Benchmark with binary search and pick the faster one.
+        if (expiry_timestamp < frame_timestamp) {
+          LOG(INFO) << "Expired";
+          break;
+        }
       }
+
+      LOG(INFO) << "Distance " << std::distance(deque.begin(), iter);
+      deque.erase(deque.begin(), iter);
+      LOG(INFO) << "Size " << deque.size();
```
ddelnano, line 307: These logs statements appear to be debugging remnants.

benkilimnik: Removed in most recent commit

src/stirling/source_connectors/socket_tracer/protocols/common/event_parser.h
```diff
@@ -120,28 +120,32 @@ ParseResult ParseFrames(message_type_t type, DataStreamBuffer* data_stream_buffe
...
     msg.timestamp_ns = timestamp_ns_status.ValueOr(0);
   }
   result.end_position += start_pos;
 
+  // Parse frames into map
+  for (auto& frame : new_frames) {
+    // GetStreamID returns 0 by default if not implemented in protocol.
+    TKey key = GetStreamID<TKey, TFrameType>(&frame);
+    (*frames)[key].push_back(frame);
+  }
```
ddelnano, line 148: Have you tried pushing the `GetStreamID` call into the `ParseFramesLoop` function? It seems like we could avoid this loop if GetStreamID could be called [here](https://github.com/pixie-io/pixie/blob/782674dec0e3a6924b7a3c3002abe13bae10fdd0/src/stirling/source_connectors/socket_tracer/protocols/common/event_parser.h#L225) as the frame is pushed into the deque.

benkilimnik: I originally had it in `ParseFramesLoop`, but this required changing how `ParseResult` is used to update the timestamp of each frame in lines 130-141. The code became rather verbose and required changing various test files that use `ParseFramesLoop` like `src/stirling/source_connectors/socket_tracer/protocols/{cql,dns,kafka,mysql}/parse_test.cc`, so I opted to move it into `ParseFrames` instead.

Happy to discuss further if you think it would be better to move it back into ParseFramesLoop.

src/stirling/e2e_tests/stirling_wrapper_size_test.cc
```diff
@@ -30,7 +30,7 @@ namespace stirling {
 #ifdef __OPTIMIZE__
 constexpr uint64_t kFileSizeLimitMB = 118;
 #else
-constexpr uint64_t kFileSizeLimitMB = 295;
+constexpr uint64_t kFileSizeLimitMB = 297;
```
ddelnano, line 33: I think we should raise this by 5 MiB (300) to provide more of a buffer. While your change tripped the threshold any significant code added in stirling will inch us closer to this (adding libbson, the mongo parser/stitcher, which includes templates, etc). So while your change added more code and templates instantiations, I think it's likely that the mongo work has contributed to this as well.

benkilimnik: I've opened a separate PR for this #1731

src/stirling/source_connectors/socket_tracer/protocols/common/event_parser.h
benkilimnik: On second thought, I think it is better to have the call to `GetStreamID` in `ParseFramesLoop`. To avoid this PR getting too large, I've moved that change to #1732

src/stirling/source_connectors/socket_tracer/protocols/cql/types.h
```diff
@@ -103,6 +103,7 @@ struct Frame : public FrameBase {
   FrameHeader hdr;
   std::string msg;
   bool consumed = false;
+  bool discarded = false;
```
ddelnano, line 106: @benkilimnik this seems like a remnant from https://github.com/pixie-io/pixie/pull/1715

benkilimnik: Removed in most recent commit

ddelnano (approved):

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -786,9 +786,11 @@ void SocketTraceConnector::TransferDataImpl(ConnectorContext* ctx) {
...
       // If there's no transfer function, then the tracker should not be holding any data.
       // http::ProtocolTraits is used as a placeholder; the frames deque is expected to be
-      // std::monotstate.
-      ECHECK(conn_tracker->send_data().Empty<protocols::http::Message>());
-      ECHECK(conn_tracker->recv_data().Empty<protocols::http::Message>());
+      // std::monostate.
+      ECHECK((conn_tracker->send_data()
+                  .Empty<protocols::http::stream_id_t, protocols::http::Message>()));
+      ECHECK((conn_tracker->recv_data()
+                  .Empty<protocols::http::stream_id_t, protocols::http::Message>()));
```
etep, line 793: We should probably make this a `DCHECK`.

Also, is it possible to create a type alias so that the line breaks are avoided, e.g.
```
using stream_id_t = protocols::http::stream_id_t;
using message_t = protocols::http::Message;
.Empty<stream_id_t, message_t>();
```

src/stirling/source_connectors/socket_tracer/data_stream.h
```diff
@@ -134,10 +139,22 @@ class DataStream : NotCopyMoveable {
...
    * @return true if empty of all data.
    */
-  template <typename TFrameType>
+  template <typename TKey, typename TFrameType>
   bool Empty() const {
-    return data_buffer_.empty() && (std::holds_alternative<std::monostate>(frames_) ||
-                                    std::get<std::deque<TFrameType>>(frames_).empty());
+    bool data_buffer_empty = data_buffer_.empty();
+    bool monostate = std::holds_alternative<std::monostate>(frames_);
+    if (data_buffer_empty || monostate) {
```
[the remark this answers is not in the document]
benkilimnik: I believe the condition of the if statement depends on the runtime state of `data_buffer_` and `frames_`, so we can't use a `constexpr`.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
benkilimnik: Added type alias and changed to `DCHECK`

JamesMBartlett (approved):

PR #1718, opened 2023-09-30
Add experiment for `px-k8ssandra` to perf tool

benkilimnik:
Summary: Adds perf experiment for `px-k8ssandra` to new `k8ssandraExperimentSuite`

The perf tool does not currently install the [cert-manager](https://cert-manager.io/docs/installation/helm/) automatically, which is required for px-k8ssandra, so we have to spin up a cluster, install the cert-manager and run the perf tool with `--use-local-cluster`. For this reason, this PR creates a new `k8ssandraExperimentSuite` to avoid polluting nightly.

```sh
bazel run //src/e2e_test/perf_tool -- run --suite=k8ssandra --experiment_name=px-k8ssandra --commit_sha=$(git rev-parse HEAD) --api_key=$API_KEY --cloud_addr=$PL_CLOUD_ADDR --use_local_cluster
```

Another issue is that the perf tool currently does not properly clean up the demo; instead of running `px demo delete`, it just tries to delete the namespace. Since the k8ssandra demo uses CRDs, this could lead to orphaned resources or lingering finalizers that might complicate subsequent tests or deployments. A workaround is to comment out the lines in workload.go that delete the namespace and manually run `px demo delete`. Note that this is not an issue if perf tool deletes the entire cluster after a perf run (which it usually does).

Type of change: /kind feature

Test Plan: Ran the perf tool and verified that data showed up for `px-k8ssandra`

### DIFF
--- a/src/e2e_test/perf_tool/pkg/deploy/workload.go
+++ b/src/e2e_test/perf_tool/pkg/deploy/workload.go
@@ -145,6 +145,8 @@ func (w *workloadImpl) WaitForHealthCheck(ctx context.Context, clusterCtx *clust
 
 // Close stops the workload.
 func (w *workloadImpl) Close() error {
+	// TODO(@benkilimnik): Run demo delete for demos that use CRDs (e.g. px-k8ssandra)
+	// current approach can lead to orphaned resources or lingering finalizers that might complicate subsequent tests or deployments.
 	for ns := range w.namespacesToDelete {
 		log.WithField("workload", w.name).WithField("namespace", ns).Trace("deleting workload namespace")
 		od := k8s.ObjectDeleter{
--- a/src/e2e_test/perf_tool/pkg/suites/experiments.go
+++ b/src/e2e_test/perf_tool/pkg/suites/experiments.go
@@ -82,6 +82,56 @@ func HTTPLoadTestExperiment(
 	return e
 }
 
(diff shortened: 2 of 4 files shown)


etep:
Agreed with the PR and have no code comments, but I don't understand the implications of the cleanup issues. Does this add k8ssandra to the nightly suite? If so, do we have leaking resources?

Maybe this just adds the k8ssandra experiment as an option, but does not auto run the workload and thus there are no leaking resources. (If this is correct, can you please clarify in the PR description..).

ddelnano:
Assuming the perf tool cleans up the GKE cluster, I think cleaning up the resources is a requirement. However, the cert-manager dependency does seem like an issue. Won't that fail when it runs in GitHub actions?

benkilimnik:
@ddelnano Yes, I think the primary issue is that the perf tool does not currently install the [cert-manager](https://cert-manager.io/docs/installation/helm/) automatically, which is required for px-k8ssandra, so we have to spin up a cluster, install the cert-manager and run the perf tool with `--use-local-cluster`. 

For this reason, I think it would make sense to create a new experiment suite to avoid polluting nightly. I've updated the PR and description to reflect this suggestion.

etep (approved):

ddelnano (approved):

JamesMBartlett (approved):

Issue #238, opened 2021-05-04
Self-Hosted Pixie Install Script


zasgar:
**Is your feature request related to a problem? Please describe.**
We would like to have an install experience for the self-hosted version of Pixie that is as easy to use as the one hosted on withpixie.ai.

**Additional context**
Our team has been busy at work this month open sourcing Pixie's source code, docs, website, and other assets, We are also actively applying to be a [CNCF sandbox project](https://github.com/cncf/toc/issues/651)!  

One of our last remaining items is to publish an install script to deploy a self-hosted version of Pixie. 

## Who offers a hosted version of Pixie?

New Relic currently offers a 100% free hosted version of Pixie Cloud. This hosting has no contingencies and will be offered indefinitely to the Pixie Community. All the code used for hosting is open source, including out production manifest files.

## What will the Self-Hosted install script do?

The Self-Hosted install script will deploy Pixie Cloud so that you can use Pixie without any external dependencies. This is the exact version of Pixie Cloud we deploy, so it'll behave exactly as the hosted version, but will require management/configuration.

## What is the timeline? 

Good question. :) We had planned to open source this script by 5/4. Unfortunately, we didn’t make it. We need more time to ensure that the Pixie Cloud deploy script will be just as easy to install Pixie Cloud as it is to install the hosted version of Pixie (in < 2 minutes!)

## But I really want to run a Self-Hosted Pixie...now!

Technically you can build and run a self-hosted Pixie using Skaffold. Check out:

https://github.com/pixie-labs/pixie/blob/main/skaffold/skaffold_cloud.yaml
https://github.com/pixie-labs/pixie/tree/main/k8s/cloud
https://github.com/pixie-labs/pixie/tree/main/k8s/cloud_deps

These directions are not fully documented and the team is choosing to focus on quickly delivering the self-hosted install script. We'll constantly be iterating on the documentation to make the project more open source friendly.

shellfu:
Just curious how this is going :)

zasgar:
We are planning to discuss this in our Pixienaut meeting today!

It's most of the way there, some more finishing touches and it should be good to go by next week. It's mostly good to go today except for a small UI bug.

htroisi:
As of today, there are two ways to get Pixie:

[...]

PR #1728, opened 2023-10-06
Update create_cloud_secrets.sh

ksrikanthreddy40:
for the issue described in https://github.com/pixie-io/pixie/issues/238#issuecomment-850994948


aimichelle:
Hi @ksrikanthreddy40 ! Thanks for the contribution! These secrets should be created and managed by the elastic operator, so you shouldn't need to create them manually.

ksrikanthreddy40:
@aimichelle Thanks for the clarification

PR #1733, opened 2023-10-11
Enforce monotonicity for timestamps in event parser

benkilimnik:
Summary: Ensures that timestamps in the event parser are monotonically increasing.

The current stitching logic for CQL relies on the assumption that frame deques are chronologically sorted based on their timestamps. Timestamps from eBPF, however, may not always be monotonic due to CPU variances, with potential discrepancies of [up to 880 usec](https://lore.kernel.org/bpf/CAJD7tkYOs4LKa=j+xNRMRiK=ors7_uCBtAjp6axRNQo0NHQqWA@mail.gmail.com/). This change catches potential anomalies due to non-monotonic timestamps by ensuring that the next frame's timestamp will always be greater than the previous one.

The DataStreamBuffer itself relies on byte positions, not timestamps, for its internal ordering so it won't be affected by this timestamp issue directly. Although DataTables are sorted before being pushed to the TableStore, the timestamp variance from eBPF might still have implications. More experiments will be required to determine how frequently this condition is hit in the wild.

Type of change: /kind bug

Test Plan: Existing targets.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
+++ b/src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
@@ -20,6 +20,7 @@
 
 #include <gflags/gflags.h>
 
+#include <cstdint>
 #include <map>
 #include <memory>
 #include <string>
@@ -95,7 +96,21 @@ class DataStreamBuffer {
    * @param pos The logical position of the data.
    * @return The timestamp or error if the position does not contain valid data.
    */
-  StatusOr<uint64_t> GetTimestamp(size_t pos) const { return impl_->GetTimestamp(pos); }
+  StatusOr<uint64_t> GetTimestamp(size_t pos) {
+    StatusOr<uint64_t> timestamp_ns_status = impl_->GetTimestamp(pos);
+    if (!timestamp_ns_status.ok()) {
+      return timestamp_ns_status;
+    }
+    uint64_t current_timestamp_ns = timestamp_ns_status.ValueOr(0);
+    uint64_t prev_timestamp_ns = GetPrevTimestamp().ValueOr(0);
+    if (current_timestamp_ns < prev_timestamp_ns) {
+      LOG(WARNING)
+          << "Detected out-of-order or equal timestamp. Adjusting to be monotonically increasing.";
+      return GetPrevTimestamp();
+    }
+    SetPrevTimestamp(timestamp_ns_status);
+    return timestamp_ns_status;
+  }
 
   /**
    * Remove n bytes from the head of the buffer.
@@ -147,7 +162,10 @@ class DataStreamBuffer {
   void ShrinkToFit() { impl_->ShrinkToFit(); }
 
  private:
+  StatusOr<uint64_t> GetPrevTimestamp() const { return prev_timestamp_ns_; }
+  void SetPrevTimestamp(const StatusOr<uint64_t>& timestamp) { prev_timestamp_ns_ = timestamp; }
   std::unique_ptr<DataStreamBufferImpl> impl_;
+  StatusOr<uint64_t> prev_timestamp_ns_;
 };
 
 }  // namespace protocols
--- a/src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
+++ b/src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
@@ -92,23 +92,22 @@ class DataStreamBuffer {
 
   /**
    * Get timestamp recorded for the data at the specified position.
+   * If less than previous timestamp, timestamp will be adjusted to be monotonically increasing.
    * @param pos The logical position of the data.
    * @return The timestamp or error if the position does not contain valid data.
    */
   StatusOr<uint64_t> GetTimestamp(size_t pos) {
     StatusOr<uint64_t> timestamp_ns_status = impl_->GetTimestamp(pos);
-    if (!timestamp_ns_status.ok()) {
-      return timestamp_ns_status;
-    }
-    uint64_t current_timestamp_ns = timestamp_ns_status.ValueOr(0);
-    uint64_t prev_timestamp_ns = GetPrevTimestamp().ValueOr(0);
+    if (!timestamp_ns_status.ok()) return timestamp_ns_status;
+    uint64_t current_timestamp_ns = timestamp_ns_status.ConsumeValueOrDie();
+    uint64_t prev_timestamp_ns = GetPrevTimestamp();
     if (current_timestamp_ns < prev_timestamp_ns) {
-      LOG(WARNING)
-          << "Detected out-of-order or equal timestamp. Adjusting to be monotonically increasing.";
-      return GetPrevTimestamp();
+      LOG(WARNING) << "Detected non-monotonically increasing timestamp " << current_timestamp_ns
+                   << ". Adjusting to previous timestamp + 1: " << prev_timestamp_ns + 1;
+      return prev_timestamp_ns + 1;
     }
-    SetPrevTimestamp(timestamp_ns_status);
-    return timestamp_ns_status;
+    SetPrevTimestamp(current_timestamp_ns);
+    return current_timestamp_ns;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/stirling/source_connectors/socket_tracer/protocols/common/event_parser.h
```diff
@@ -129,6 +129,7 @@ ParseResult ParseFrames(message_type_t type, DataStreamBuffer* data_stream_buffe
   VLOG(1) << absl::Substitute("Parsed $0 new frames", frames->size() - prev_size);
 
   // Match timestamps with the parsed frames.
+  uint64_t prev_timestamp_ns = 0;
   for (size_t i = 0; i < result.frame_positions.size(); ++i) {
```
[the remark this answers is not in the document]
benkilimnik: I've added a field to the datastream buffer `StatusOr<uint64_t> prev_timestamp_ns_;`.  Whenever a non-monotonic timestamp is encountered, we log a warning, and return the previous timestamp.

It may be cleaner to move the error handling in `event_parser.h` to `data_stream_buffer.h`.
```cpp
LOG_IF(ERROR, !timestamp_ns_status.ok()) << timestamp_ns_status.ToString();
```
Currently `GetTimestamp` simply returns a non-OK statusOr to be handled by the caller.
```cpp
if (!timestamp_ns_status.ok()) {
  return timestamp_ns_status;
}
```
Curious to hear your thoughts on this.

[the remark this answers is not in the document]
benkilimnik: Another question is whether we want to `ValueOr(0)` here, as we do in `event_parser`
```cpp
uint64_t current_timestamp_ns = timestamp_ns_status.ValueOr(0);
uint64_t prev_timestamp_ns = GetPrevTimestamp().ValueOr(0);
```

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
```diff
@@ -95,7 +95,21 @@ class DataStreamBuffer {
    * @param pos The logical position of the data.
    * @return The timestamp or error if the position does not contain valid data.
    */
-  StatusOr<uint64_t> GetTimestamp(size_t pos) const { return impl_->GetTimestamp(pos); }
+  StatusOr<uint64_t> GetTimestamp(size_t pos) {
```
ddelnano, line 98: Adding a chunk with a non monotonic timestamp through `DataStream::Add` should be easy to test this behavior. Can we add an new test case to [DataStreamBufferTest.Timestamp](https://github.com/pixie-io/pixie/blob/3e02189116d9c3c5f488504894d015809d0bcf19/src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer_test.cc#L131)?

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
```diff
@@ -95,7 +95,21 @@ class DataStreamBuffer {
...
+    StatusOr<uint64_t> timestamp_ns_status = impl_->GetTimestamp(pos);
+    if (!timestamp_ns_status.ok()) {
+      return timestamp_ns_status;
+    }
+    uint64_t current_timestamp_ns = timestamp_ns_status.ValueOr(0);
+    uint64_t prev_timestamp_ns = GetPrevTimestamp().ValueOr(0);
```
ddelnano, line 104: I'm open to other's thoughts, but I think it's cleaner here to use `ConsumeValueOrDie()`. The implementation of that function checks the same field as `StatusOr::ok()`, so it should be safe unless we feel that the object will be changed in between these calls.

@etep what do you think?

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
```diff
@@ -147,7 +161,10 @@ class DataStreamBuffer {
   void ShrinkToFit() { impl_->ShrinkToFit(); }
 
  private:
+  StatusOr<uint64_t> GetPrevTimestamp() const { return prev_timestamp_ns_; }
+  void SetPrevTimestamp(const StatusOr<uint64_t>& timestamp) { prev_timestamp_ns_ = timestamp; }
   std::unique_ptr<DataStreamBufferImpl> impl_;
+  StatusOr<uint64_t> prev_timestamp_ns_;
```
ddelnano, line 167: Did you consider storing the `uint64_t` directly? I have a slight preference to unwrap it and keep the integer value rather than the `StatusOr` object.

src/stirling/source_connectors/socket_tracer/protocols/common/event_parser.h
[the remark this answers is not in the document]
ddelnano: I'm fine with the error handling in the current code, however, I could see a case for moving it from `event_parser.h` to `data_stream_buffer.h`. As far as I can tell, the only caller of `GetTimestamp` (outside of tests) is the event parser. So we could keep the existing `ValueOr(0)` semantic entirely within the buffer. No strong opinion one way or the other 

Regarding the `ValueOr(0)`, see my comment about possibly using `ConsumeValueOrDie()`.

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
benkilimnik: Changed to `uint64_t`

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
benkilimnik: Added!

ddelnano (approved):
lgtm as long as @etep has no additional comments. Really appreciate all the investment on this to make sure we get it right!

benkilimnik:
I did notice that this particular CI test took quite a while to run (1.5 hours) in comparison to the others: [build-and-test (BPF opt (4.14.254), kvm, --config=qemu-bpf --compilation_mode=opt](https://github.com/pixie-io/pixie/actions/runs/6511408388/job/17687148129?pr=1733#logs)

Do we know if that is unusual behavior?

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
```diff
@@ -92,10 +92,23 @@ class DataStreamBuffer {
...
-  StatusOr<uint64_t> GetTimestamp(size_t pos) const { return impl_->GetTimestamp(pos); }
+  StatusOr<uint64_t> GetTimestamp(size_t pos) {
+    StatusOr<uint64_t> timestamp_ns_status = impl_->GetTimestamp(pos);
+    if (!timestamp_ns_status.ok()) return timestamp_ns_status;
+    uint64_t current_timestamp_ns = timestamp_ns_status.ConsumeValueOrDie();
+    uint64_t prev_timestamp_ns = GetPrevTimestamp();
+    if (current_timestamp_ns < prev_timestamp_ns) {
+      LOG(WARNING) << "Detected non-monotonically increasing timestamp " << current_timestamp_ns
+                   << ". Adjusting to previous timestamp + 1: " << prev_timestamp_ns + 1;
+      return prev_timestamp_ns + 1;
```
etep, line 107: Probably we want:
```
if (current < prev) {
  current += 1;
}
SET_PREVIOUS(current);
return current;
```

benkilimnik: I believe that incrementing `current` by one may not be enough to make it greater than the previous timestamp. Did you mean `prev += 1`?

benkilimnik: I've adopted the structure you outlined, replacing `current += 1` with `current = prev + 1`

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
```diff
@@ -92,10 +92,23 @@ class DataStreamBuffer {
...
   /**
    * Get timestamp recorded for the data at the specified position.
+   * If less than previous timestamp, timestamp will be adjusted to be monotonically increasing.
    * @param pos The logical position of the data.
    * @return The timestamp or error if the position does not contain valid data.
    */
-  StatusOr<uint64_t> GetTimestamp(size_t pos) const { return impl_->GetTimestamp(pos); }
+  StatusOr<uint64_t> GetTimestamp(size_t pos) {
+    StatusOr<uint64_t> timestamp_ns_status = impl_->GetTimestamp(pos);
+    if (!timestamp_ns_status.ok()) return timestamp_ns_status;
```
etep, line 101: My strong bias is towards always including the curly braces.

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer.h
```diff
@@ -92,10 +92,23 @@ class DataStreamBuffer {
...
+    uint64_t current_timestamp_ns = timestamp_ns_status.ConsumeValueOrDie();
+    uint64_t prev_timestamp_ns = GetPrevTimestamp();
+    if (current_timestamp_ns < prev_timestamp_ns) {
+      LOG(WARNING) << "Detected non-monotonically increasing timestamp " << current_timestamp_ns
+                   << ". Adjusting to previous timestamp + 1: " << prev_timestamp_ns + 1;
+      current_timestamp_ns = prev_timestamp_ns + 1;
+    }
+    SetPrevTimestamp(current_timestamp_ns);
```
etep, line 109: IIUC we do not need the setter/getter pattern for the previous timestamp member. Would recommend just directly accessing the private member here.

etep (approved):

JamesMBartlett (approved):

PR #1735, opened 2023-10-14
uprobe manager: Use ns_pid when deriving .so paths

danielkleinstein:


danielkleinstein:
See https://github.com/pixie-io/pixie/issues/1736

ddelnano:
Closing since #1736 fixes the original regression and adds test coverage for it. Thanks for the very detailed report and identifying this issue @danielkleinstein!

Issue #1736, opened 2023-10-14
libssl.so.3 observability is broken


danielkleinstein:
**Describe the bug**
OpenSSL visibility seems to be broken, at least with `libssl.so.3`.

- On one of my cluster nodes, I am running a Python 3.10 program that performs outgoing HTTPS traffic against an API endpoint

- The relevant PEM's log shows the following line:
> uprobe_manager.cc:624] Attaching OpenSSL uprobes on dynamic library failed for PID 226977: Internal : libssl not found [path = /proc/226977/root/proc/226977/root/usr/lib/x86_64-linux-gnu/libssl.so.3]
    i.e. it looks like it's trying to access /proc/226977/root/proc/226977/root/usr/lib/x86_64-linux-gnu/libssl.so.3 and failing

- On the same node that the Python/PEM are running in, I ran a privileged K8s pod that allowed me to run some commands in the host's namespaces - this allowed me to pinpoint the problem: /proc/226977/root/proc/226977/ does not exist. 226977 is indeed the PID of the Python program relative to the host, but it shouldn't appear twice - the last PID should be the PID relative to the container.

**To Reproduce**

- Run `kubectl run ubuntu --image=ubuntu --rm -it -- bash`
- Inside that container, create the following Python script as `http-traffic.py` and run `./http-traffic.py` (installing dependencies as necessary):
```python
#!/usr/bin/env python3

import argparse
import requests
import sys
import time

DEFAULT_ENDPOINT = "https://api.sampleapis.com/countries/countries"

def fetch_endpoint(endpoint, interval=1, count=None, quiet=False):
    global total_requests
    total_requests = 0

    while True:
        response = requests.get(endpoint)
        total_requests += 1
        if not quiet:
            print(response.text)

        if count is not None and total_requests >= count:
            break

        time.sleep(interval)
        if not quiet:
            sys.stderr.write(f"Total requests made: {total_requests}\n")

if __name__ == "__main__":
    parser = argparse.ArgumentParser(description="Access an endpoint at regular intervals.")
    parser.add_argument("-e", "--endpoint", default=DEFAULT_ENDPOINT, help=f"URL of the endpoint to access (default is {DEFAULT_ENDPOINT}).")
    parser.add_argument("-i", "--interval", type=float, default=1, help="Interval in seconds between requests (default is 1 second).")
    parser.add_argument("-c", "--count", type=int, help="Limit the number of requests to this count. If not provided, the script will run indefinitely.")
[...]

PR #1740, opened 2023-10-16
Fix duplicate `/proc/<pid>/root` prefix bug that formed wrong path to `libssl.so` files.

etep:
Summary: Fixes a bug that prevented SSL tracing on dynamically loaded SSL libraries. When probing open ssl libs found from a pod's mapped libraries, we erroneously prefixed library paths with `/proc/<pid>/root` twice, i.e. such that a target library might look like this:
```
/proc/<pid>/root/proc/<pid>/root/<path>/some-lib.so
```
Normal target processes run in their own pid namespace. This meant that the *second* instance of `<pid>` in the path above would not exist in the target process namespace. Hence probing dynamic libraries using this path failed.

We also update the open-ssl test case to not use host pid namespace, i.e. such that those tests expose this bug. Previously, this bug was masked in those test cases because the test container processes ran with host pid namespace.

Related issues: #1736.

Test Plan: Show that tests fail when we disable use of host pid namespace in SSL test containers, see #1744.

Changelog Message:
```release-note
Fixed a bug that prevented SSL tracing on dynamically linked SSL libraries. 
```

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
+++ b/src/stirling/source_connectors/socket_tracer/openssl_trace_bpf_test.cc
@@ -104,7 +104,7 @@ class OpenSSLTraceTest : public SocketTraceBPFTestFixture</* TClientSideTracing 
     // Run the nginx HTTPS server.
     // The container runner will make sure it is in the ready state before unblocking.
     // Stirling will run after this unblocks, as part of SocketTraceBPFTest SetUp().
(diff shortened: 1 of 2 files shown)


ddelnano (approved):
@etep thanks for tracking this down and fixing the issue.

I think we should include a `Changelog Message:` in the PR explaining that a bug was fixed that prevented dynamically linked OpenSSL tracing from working. In addition to that, can you update the `Title:` to match the PR's.

JamesMBartlett (approved):

PR #1743, opened 2023-10-17
Add support to stitch MongoDB frames

kpattaswamy:
Summary: This PR adds functionality to stitch MongoDB frames together. It relies on state to iterate over the order of streamIDs parsed and uses a response led matching approach.

Type of change: /kind feature

Test Plan: Added tests

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
@@ -35,21 +35,31 @@ namespace stirling {
 namespace protocols {
 namespace mongodb {
 
+void FlattenSections(mongodb::Frame* frame) {
+  for (const auto& section : frame->sections) {
+    for (const auto& doc : section.documents) {
+      frame->frame_body.append(doc).append(" ");
+    }
+  }
+  frame->sections.clear();
+}
+
 RecordsWithErrorCount<mongodb::Record> StitchFrames(
     absl::flat_hash_map<mongodb::stream_id_t, std::deque<mongodb::Frame>>* reqs,
     absl::flat_hash_map<mongodb::stream_id_t, std::deque<mongodb::Frame>>* resps, State* state) {
-
   std::vector<mongodb::Record> records;
   int error_count = 0;
 
-  for (const auto& [idx, stream_id] : Enumerate(state->transaction_stream_order)) {
+  for (auto stream_id_it = state->stream_order.begin(); stream_id_it != state->stream_order.end();
+       stream_id_it++) {
+    auto& stream_id_pair = *stream_id_it;
+    auto stream_id = stream_id_pair.first;
+
     // Find the stream ID's response deque.
     auto resp_it = resps->find(stream_id);
     if (resp_it == resps->end()) {
-      VLOG(1) << absl::Substitute(
-          "Did not find a response deque with the stream ID: $0",
-          stream_id);
-      error_count++;
+      VLOG(1) << absl::Substitute("Did not find a response deque with the stream ID: $0",
+                                  stream_id);
       continue;
     }
 
@@ -59,9 +69,7 @@ RecordsWithErrorCount<mongodb::Record> StitchFrames(
     // Find the stream ID's response deque.
     auto req_it = reqs->find(stream_id);
     if (req_it == reqs->end()) {
-      VLOG(1) << absl::Substitute(
-          "Did not find a request with the stream ID = $0",
-          stream_id);
+      VLOG(1) << absl::Substitute("Did not find a request with the stream ID = $0", stream_id);
       error_count++;
       continue;
     }
@@ -72,7 +80,8 @@ RecordsWithErrorCount<mongodb::Record> StitchFrames(
     // Track the latest response timestamp to compare against request frame's timestamp later.
     uint64_t latest_resp_ts = 0;
 
-    // Loop over each frame in the response deque and match the oldest response frame with the oldest request that occured just before the response.
+    // Loop over each frame in the response deque and match the oldest response frame with the
+    // oldest request that occured just before the response.
     for (const auto& [idx, resp_frame] : Enumerate(resp_deque)) {
       if (resp_frame.consumed) {
         continue;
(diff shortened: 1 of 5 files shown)


src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
```diff
@@ -0,0 +1,198 @@
...
+      if (req_frame_it == req_deque.begin() &&
+          req_frame_it->timestamp_ns > resp_frame.timestamp_ns) {
+        VLOG(1) << absl::Substitute(
+            "Did not find a request frame that is earlier than the response. Response's "
+            "responseTo: $0",
+            resp_frame.response_to);
+        resp_deque.erase(resp_deque.begin() + idx);
+        error_count++;
+        break;
+      }
```
ddelnano, line 148: I was expecting this logic to come before the more to come parsing. Don't we need to identify the correct response with the timestamp heuristic before the 1 or n responses can be parsed?

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher_test.cc
```diff
@@ -0,0 +1,368 @@
...
+  mongodb::Section section;
+  section.documents.push_back(doc);
+
+  frame.sections.push_back(section);
+
+  return frame;
+}
+
+// Uncomment following two functions once upstream testing functions are merged.
+bool areAllDequesEmpty(
```
ddelnano, line 55: This should be PascalCase per our [style guide ](https://github.com/pixie-io/pixie/blob/e670eb63cb18368d671169fb630567d78cd3f8c8/styleguide/cc_style.md#c-style-guide-proposalsopens) -- `AreAllDequeuesEmpty`. The particular part is in the [Google C++ guide](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules).

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher_test.cc
```diff
@@ -0,0 +1,368 @@
...
+    const absl::flat_hash_map<mongodb::stream_id_t, std::deque<mongodb::Frame>>& frame_map) {
+  for (const auto& pair : frame_map) {
+    if (!pair.second.empty()) {
+      return false;
+    }
+  }
+  return true;
+}
+
+size_t totalDequeSize(
```
ddelnano, line 65: Same here.

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
```diff
@@ -0,0 +1,198 @@
...
+      if (resp_frame.consumed) {
+        continue;
+      }
+
+      latest_resp_ts = resp_frame.timestamp_ns;
+
+      auto curr_resp = resp_frame;
+
+      // Find and insert all of the moreToCome frame(s) section data to the head response frame.
+      while (curr_resp.more_to_come) {
```
ddelnano, line 96: I think putting the more to come logic in a dedicated function would help with the readability of this if it's possible.

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
```diff
@@ -0,0 +1,198 @@
...
+          break;
+        }
+
+        mongodb::Frame& next_resp = *next_resp_it;
+        resp_frame.sections.insert(std::end(resp_frame.sections), std::begin(next_resp.sections),
+                                   std::end(next_resp.sections));
+        next_resp.consumed = true;
+        latest_resp_ts = next_resp.timestamp_ns;
+        curr_resp = next_resp;
+        next_resp_deque.erase(next_resp_it);
```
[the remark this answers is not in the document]
kpattaswamy: I initially thought to erase the `more to come` frame as it was consumed so the `error count` could be updated with the number of response frames remaining in each deque after stitching. But I think the better approach would be to set the `more to come` frame as consumed and updating the error count with the number of unconsumed frames in each response deque

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
```diff
@@ -0,0 +1,210 @@
...
+      latest_resp_ts = resp_frame.timestamp_ns;
+
+      // Find the corresponding request frame for the head response frame.
+      auto req_frame_it = std::upper_bound(req_deque.begin(), req_deque.end(), latest_resp_ts,
+                                           [](const uint64_t ts, const mongodb::Frame& frame) {
+                                             return ts < frame.timestamp_ns;
+                                           }) -
+                          1;
+      if (req_frame_it == req_deque.begin() &&
+          req_frame_it->timestamp_ns > resp_frame.timestamp_ns) {
```
ddelnano, line 146: @kpattaswamy have you validated that this condition works? If there are no earlier requests, I think we will get `req_deque.begin() - 1` because we are subtracting 1 from the `std::upper_bound` output.

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
```diff
@@ -0,0 +1,210 @@
...
+    // current response frame's timestamp.
+    auto next_resp_it = std::upper_bound(
+        next_resp_deque.begin(), next_resp_deque.end(), *latest_resp_ts,
+        [](const uint64_t ts, const mongodb::Frame& frame) { return ts < frame.timestamp_ns; });
+    if (next_resp_it->timestamp_ns < *latest_resp_ts) {
+      VLOG(1) << absl::Substitute(
+          "Did not find a response extending the prior more to come response. RequestID: $0",
+          curr_resp->request_id);
+      (*error_count)++;
+      break;
```
ddelnano, line 72: Since this is inside of a function now, I think it would be better to return rather than break.
```suggestion
      return;
```

ddelnano (approved):
These changes lgtm. Appreciate all the hard work on the various iterations we investigated!

Please make sure any comments @etep has are addressed as well.

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
```diff
@@ -0,0 +1,212 @@
...
+  frame->sections.clear();
+}
+
+RecordsWithErrorCount<mongodb::Record> StitchFrames(
+    absl::flat_hash_map<mongodb::stream_id_t, std::deque<mongodb::Frame>>* reqs,
+    absl::flat_hash_map<mongodb::stream_id_t, std::deque<mongodb::Frame>>* resps, State* state) {
+  std::vector<mongodb::Record> records;
+  int error_count = 0;
+
+  for (auto stream_id_it = state->stream_order.begin(); stream_id_it != state->stream_order.end();
```
etep, line 104: Can this be written as:
```
for (const auto& [foo ,bar] : state->stream_order ) {
...
```

etep (approved):

JamesMBartlett (approved):

PR #1738, opened 2023-10-16
Revert "Added cgid capture to stirling with testing (#1681)"

benkilimnik:
Summary: Revert cgid capture

This change reverts capturing Cgroup ids from https://github.com/pixie-io/pixie/commit/2a2349f4853c90848437441ceed4bad634505767. Many of the 4.14 kernel tests have [BPF compilation errors](https://app.buildbuddy.io/invocation/075de8d0-d161-4ff6-8b74-123c6da66778#): `HINT: bpf_get_current_cgroup_id missing (added in Linux 4.18)`. Perplexingly, GitHub seems to show these tests as [passing despite the failures](https://github.com/pixie-io/pixie/actions/runs/6366478258/job/17320729511).

Relevant Issues: #1638 #1739

Type of change: /kind bug

Test Plan: Running CI

PR #1748, opened 2023-10-20
Add cgroup bcc flag for compilation

AdityaAtulTewari:
This is to ensure that the reversions in #1738 are not necessary for #1638.

### DIFF
--- a/src/stirling/bpf_tools/bcc_wrapper_bpf_test.cc
+++ b/src/stirling/bpf_tools/bcc_wrapper_bpf_test.cc
@@ -27,6 +27,7 @@
 #include "src/stirling/bpf_tools/macros.h"
 #include "src/stirling/obj_tools/address_converter.h"
 #include "src/stirling/obj_tools/testdata/cc/test_exe_fixture.h"
+#include "src/shared/metadata/cgroup_path_resolver.h"
 
 // A function which we will uprobe on, to trigger our BPF code.
 // The function itself is irrelevant, but it must not be optimized away.
@@ -43,6 +44,7 @@ namespace bpf_tools {
 
 using ::px::testing::BazelRunfilePath;
 using ::px::testing::TempDir;
+
 
 constexpr char kBCCProgram[] = R"BCC(
   int foo(struct pt_regs* ctx) {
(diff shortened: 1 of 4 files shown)


src/stirling/bpf_tools/testdata/get_current_cgroup_id.c
```diff
@@ -0,0 +1,13 @@
+#include "src/stirling/bpf_tools/bcc_bpf/task_struct_utils.h"
+
+BPF_ARRAY(cgroup_id_output, uint64_t, 1);
+
+int probe_bpf_get_current_cgroup_id(struct pt_regs* ctx) {
+  uint64_t cgroup_id = UINT64_MAX;
+#if GET_CGROUP_ID_ENABLED
+  cgroup_id = bpf_get_current_cgroup_id();
+#endif
```
ddelnano, line 9: While this `BCCWrapper` test verifies our compilation works, I think it would be better to have this `#if` check within library code. Otherwise we have to rewrite the same `#if` check in the socket tracer BPF code (meaning it's not benefitting from this test coverage).

My thinking is that we should add this "cgroup function" to `src/stirling/bpf_tools/bcc_bpf/utils.h`. Something along the lines of this:

```
// src/stirling/bpf_tools/bcc_bpf/utils.h

#ifndef GET_CGROUP_ID_ENABLED
uint64_t bpf_get_current_cgroup_id() {
  return UINT64_MAX;
}
#endif

```

Rather than only defining it for older kernels, I could also see a light wrapper making sense (like you did here).

We can then update `probe_bpf_get_current_cgroup_id` to include that `utils.h` header file and call that function. That will ensure that the version the socket tracer uses is exercised by this test..

src/stirling/bpf_tools/bcc_wrapper_bpf_test.cc
```diff
@@ -169,6 +170,52 @@ TEST(BCCWrapperTest, GetTGIDStartTime) {
   EXPECT_EQ(proc_pid_start_time, expected_proc_pid_start_time);
 }
 
+TEST(BCCWrapperTest, GetTGIDStartTime) {
```
ddelnano, line 173:
```suggestion
TEST(BCCWrapperTest, GetCurrentCgroupId) {
```

src/stirling/bpf_tools/bcc_wrapper_bpf_test.cc
```diff
@@ -169,6 +170,52 @@ TEST(BCCWrapperTest, GetTGIDStartTime) {
...
+  // Force the TaskStructResolver to run,
+  // since we're trying to check that it correctly gets the task_struct offsets.
```
ddelnano, line 175: Seems to be copied from the other test case and should be removed.

src/stirling/bpf_tools/bcc_wrapper_bpf_test.cc
```diff
@@ -169,6 +170,52 @@ TEST(BCCWrapperTest, GetTGIDStartTime) {
...
+  std::vector<std::string> cflags = {};
+  bool requires_linux_headers = true;
+  bool always_infer_task_struct_offsets = true;
+
+  BCCWrapperImpl bcc_wrapper;
+  ASSERT_OK(bcc_wrapper.InitBPFProgram(get_tgid_start_time_bcc_script, cflags,
```
ddelnano, line 181: This doesn't seem like it's attaching the correct BPF program. We should have an additional `OBJ_STRVIEW(get_current_cgroup_id)` and pass in `get_current_cgroup_id_bcc_script`.

src/stirling/bpf_tools/testdata/get_current_cgroup_id.c
AdityaAtulTewari: I see this makes a lot of sense. I'll make these changes.

ddelnano:
@AdityaAtulTewari sorry for the late reply here. The solution I envisioned was what I've implemented in https://github.com/pixie-io/pixie/commit/3bab3f0bdf03b8ffddb7a58138c2f4d3a64b3ec4. That change isn't production ready, but shows how the `pl_bpf_cc_resource` bazel rule can be enhanced to support delaying some c preprocessing logic. I've verified that this prevents the bcc compilation issues and that the resulting socket_trace.c file still contains the `#ifdef` macros.

```
ddelnano@pixie-dev:~/code/pixie (ddelnano/add-cgroup-short-proc-fix-behind-feature-toggle) $ bazel build  src/stirling/source_connectors/socket_tracer\/bcc_bpf:socket_trace_bpf_preprocess
Starting local Bazel server and connecting to it...
INFO: Analyzed target //src/stirling/source_connectors/socket_tracer/bcc_bpf:socket_trace_bpf_preprocess (87 packages loaded, 1828 targets configured).
INFO: Found 1 target...
INFO: Deleting stale sandbox base /home/ddelnano/.cache/bazel/_bazel_ddelnano/b2e8b093089e374aa7d44dfed4f9953b/sandbox
Target //src/stirling/source_connectors/socket_tracer/bcc_bpf:socket_trace_bpf_preprocess up-to-date:
  bazel-bin/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace_bpf_preprocess.c
INFO: Elapsed time: 13.225s, Critical Path: 0.18s
INFO: 2 processes: 1 internal, 1 linux-sandbox.
INFO: Build completed successfully, 2 total actions


ddelnano@pixie-dev:~/code/pixie (ddelnano/add-cgroup-short-proc-fix-behind-feature-toggle) $ head -n 11 bazel-bin/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace_bpf_preprocess.c
#ifdef GET_CGROUP_ID_ENABLED
uint64_t pl_bpf_get_current_cgroup_id() {
    return bpf_get_current_cgroup_id();
}
#else

uint64_t pl_bpf_get_current_cgroup_id() {
    // TODO(ddelnano): UINT64_MAX doesn't work in BPF. Find another way to represent this.
    return (uint64_t)18446744073709551615;
}
#endif
```

In order to move forward with this, I think we will need some feedback on this approach from @vihangm and @JamesMBartlett. If this is something we are comfortable with, then we can shape up what I've prototyped.

Issue #341, opened 2021-10-25
Support MQTT protocol(s)


mak3r:
**Describe the solution you'd like**
It would be useful to add MQTT to the [list of protocols](https://docs.px.dev/about-pixie/data-sources#supported-protocols) which support tracing. In particular, it would be beneficial to have  support for [Request Tracing](https://docs.px.dev/tutorials/pixie-101/request-tracing) and [Service Performance](https://docs.px.dev/tutorials/pixie-101/service-performance) of the MQTT protocol. 

MQTT is one of the most common protocols for device communication from edge nodes such as microcontrollers and IOT sensors. While these types of devices may not yet be capable of hosting container orchestration components like the kubelet themselves, they are a big part of the ecosystem for cloud-native at the edge. Projects like Node Feature Discovery (NFD), Akri and Smarter Device Manager are exposing non-standard resources at the edge and it would be a great addition to the developing toolset to enable Pixie to help with the practical issues facing developers who need to get existing systems integrated into a cloud-native ecosystem.

**Describe alternatives you've considered**
There are other message broker alternatives however I suspect it would be useful to implement a variety of options.

**Additional context**
There are two protocol standards today. Implementation of v5 which has been available since March of 2019 seems the most logical however there is a good chance that a large proportion of the devices using MQTT today are still using the older protocol. I do not currently have data to support this or refute it. Here are links to the different protocols.
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.pdf
https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.pdf

A description of MQTT from the abstract of version 5:
*MQTT is a Client Server publish/subscribe messaging transport protocol. It is light weight, open, simple, and designed to be easy to implement. These characteristics make it ideal for use in many situations, including constrained environments such as for communication in Machine to Machine (M2M) and Internet of Things (IoT) contexts where a small code footprint is required and/or network bandwidth is at a premium. The protocol runs over TCP/IP, or over other network protocols that provide ordered, lossless, bidirectional connections.*

Shubhamlmp:
Hello @mak3r I am interested in working on this project. and I want to discuss more about this project with you.

deprov447:
[...]

PR #1756, opened 2023-10-31
Enable support for MQTT Parser in stirling

ChinmayaSharma-hue:
Summary: This PR adds the parser component of MQTT (v5), a newly added protocol.

Related issues: #341

Type of change: /kind feature

Test Plan: Added tests


src/stirling/source_connectors/socket_tracer/protocols/mqtt/stitcher.h
```diff
@@ -0,0 +1,44 @@
...
+
+namespace px {
+namespace stirling {
+namespace protocols {
+
+template <>
+inline RecordsWithErrorCount<mqtt::Record> StitchFrames(std::deque<mqtt::Message>* req_messages,
+                                                        std::deque<mqtt::Message>* resp_messages,
+                                                        NoState* /*state*/) {
+  return StitchMessagesWithTimestampOrder<mqtt::Record>(req_messages, resp_messages);
```
ddelnano, line 39: The stitcher should be the next PR after the parser changes, so I don't want to dive into this too much now. However, it seems that message ordering is only guaranteed within a given QoS level. I think using `StitchMessagesWithTimestampOrder` will result in invalid protocol traces if a client used multiple QoS levels.

Screenshot 2023-11-01 at 8 54 26 AM

In the next PR we will add tests for the Stitcher and we can model that situation. We will likely need to perform a similar process as `StitchMessagesWithTimestampOrder`, making sure that we only match frames within the same QoS since that will guarantee its [assumptions hold](https://github.com/pixie-io/pixie/blob/b6cf0ac3e4723a409b41cc0b603f3b8ddc1001b3/src/stirling/source_connectors/socket_tracer/protocols/common/timestamp_stitcher.h#L32-L35). This is why it doesn't work with HTTP pipelining.

We can either use the QoS field as the map key to our new stitcher interface. This hasn't been documented yet as it's in the process of being merged (#1716) or we can leverage the protocol state to make sure we have the ordering correct. My initial thinking is that the former would be best, but we will have to see.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse_test.cc
```diff
@@ -0,0 +1,914 @@
...
+
+class MQTTParserTest : public ::testing::Test {
+};
+
+TEST_F(MQTTParserTest, Properties) {
+    Message frame;
+    ParseState result_state;
+    std::string_view frame_view;
+
+    uint8_t payload_format_indicator_publish[] = {
```
ddelnano, line 38: Did these payloads come from real packet captures or are they handcrafted?

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse_test.cc
```diff
@@ -0,0 +1,914 @@
...
+    ASSERT_EQ(result_state, ParseState::kSuccess);
+    EXPECT_EQ(frame.control_packet_type, "CONNECT");
+    EXPECT_EQ(frame.header_fields["remaining_length"], (size_t) 16);
+    EXPECT_EQ(frame.header_fields["username_flag"], (unsigned long) 0);
+    EXPECT_EQ(frame.header_fields["password_flag"], (unsigned long) 0);
+    EXPECT_EQ(frame.header_fields["will_retain"], (unsigned long) 0);
+    EXPECT_EQ(frame.header_fields["will_qos"], (unsigned long) 0);
+    EXPECT_EQ(frame.header_fields["will_flag"], (unsigned long) 0);
+    EXPECT_EQ(frame.header_fields["clean_start"], (unsigned long) 1);
+    EXPECT_EQ(frame.header_fields["keep_alive"], (unsigned long) 60);
```
ddelnano, line 800: Please use `UL` to form the appropriate literals or remove the casting entirely. The tests seem to pass without these, so my preference is the latter. This applies to the other casting done in this file.

ddelnano (commented):
@ChinmayaSharma-hue really appreciate all your hard work on this and very excited to have MQTT support within Pixie! I still haven't made it the whole way through this since I'm new to MQTT. However, I wanted to post the feedback I have so far.

src/stirling/source_connectors/socket_tracer/bcc_bpf/BUILD.bazel
```diff
@@ -82,6 +82,7 @@ pl_cc_test(
         "ENABLE_NATS_TRACING=true",
         "ENABLE_MONGO_TRACING=true",
         "ENABLE_AMQP_TRACING=true",
+        "ENABLE_MQTT_TRACING=true",
     ],
     deps = [
         "//src/stirling/bpf_tools/bcc_bpf:headers",
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Sure, Noted! Will update in the next commit.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+        ],
+    ),
+    hdrs = glob(
+        [
+            "*.h",
+        ],
+    ),
+    deps = [
+        "//src/common/json:cc_library",
+        "//src/common/zlib:cc_library",
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Sure, Noted! Will update in the next commit.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/types.h
```diff
@@ -0,0 +1,124 @@
...
+
+struct State {
+  bool conn_closed = false;
+};
+
+struct StateWrapper {
+  State global;
+  std::monostate send;
+  std::monostate recv;
+};
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: I did this in the beginning when I wasn't sure if this was needed, will remove this.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,678 @@
...
+    }
+    return ParseState::kSuccess;
+}
+
+ParseState ParseVariableHeader(Message* result, BinaryDecoder* decoder, MqttControlPacketType& control_packet_type) {
+    switch (control_packet_type) {
+        case MqttControlPacketType::CONNECT: {
+            PX_ASSIGN_OR_RETURN_ERROR(uint8_t protocol_name_length, decoder->ExtractBEInt<uint16_t>());
+            PX_ASSIGN_OR_RETURN_ERROR(std::string_view protocol_name,decoder->ExtractString((size_t)protocol_name_length));
+            CTX_DCHECK(protocol_name == "MQTT");
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Is MQIsdp used for MQTTv3? I am not sure if the implementation will work for mqttv3 since mqtt v3 does not have properties. There might be other v5 specific code that might not work for v3 since I have not thoroughly gone through v3. I need to look more into this.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,678 @@
...
+    uint8_t control_packet_flags = control_packet_code_flags & 0x0F;
+
+    MqttControlPacketType control_packet_type = GetControlPacketType(control_packet_code);
+    result->control_packet_type = ControlPacketTypeStrings[control_packet_type];
+
+    // Saving the flags if control packet type is PUBLISH
+    if (control_packet_type == MqttControlPacketType::PUBLISH) {
+        result->header_fields["dup"] = (control_packet_flags >> 3) != 0;
+        result->header_fields["retain"] = (control_packet_flags & 0x1) != 0;
+        result->header_fields["qos"] = (control_packet_flags >> 1) & 0x3;
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Except for QOS right? Because it would be helpful to know the exact QOS (0,1 or 2) instead of just knowing whether or not the qos is 0 or not.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse_test.cc
ChinmayaSharma-hue: They came from real packet captures. I used mosquitto to generate packets with different properties and payloads.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/stitcher.h
ChinmayaSharma-hue: Makes sense. I had initially used `StitchMessagesWithTimestampOrder` as a placeholder as I was not sure what would work. I will focus more on the stitcher in the stitcher PR. Thanks for the additional context, it's very helpful.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/types.h
```diff
@@ -0,0 +1,124 @@
...
+#pragma once
+
+#include "src/common/base/utils.h"
+#include "src/stirling/source_connectors/socket_tracer/protocols/common/event_parser.h"  // For FrameBase.
+
+namespace px {
+namespace stirling {
+namespace protocols {
+namespace mqtt {
+
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Noted!

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,678 @@
...
+            return ParseState::kSuccess;
+        default:
+            return ParseState::kInvalid;;
+    }
+}
+
+ParseState ParseFrame(message_type_t type, std::string_view* buf,
+                  Message* result) {
+    CTX_DCHECK(type == message_type_t::kRequest || type == message_type_t::kResponse);
+    if (buf->size() < 2) {
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Would it not work to return kNeedMoreData for all cases of `ExtractUVarInt` errors? If buffer size is 3 bytes and expected packet length is 4 then `ExtractUVarInt` would return an error and based on this kNeedsMoreData would be returned. I have added another function that checks whether or not the number returned by `ExtractUVarInt` is over 4 bytes, which would cause the function to return kInvalid. So both the cases would be taken care of in this way right?
Also, I am not sure why it would be easier to change `<2` to `<5`. Wouldn't this eliminate cases where the buffer size is 2? (PINGREQ and PINRESP) are only 2 bytes, with remaining length set to 0.)

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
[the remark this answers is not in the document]
ddelnano: Yea, I glossed over that the qos field wasn't a bool. The data type should fit whatever we need to store for each field.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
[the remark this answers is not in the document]
ddelnano: When I made this comment it was before I read the MQTT v5 docs (I had trouble finding it first so I went with v3). It does seem like that specifically mentions `"MQTT"`, so let's just drop this.

I agree that we would need some changes to support MQTT v3. I think it's fine to target the latest version and not increase the scope of this initial implementation.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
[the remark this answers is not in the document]
ChinmayaSharma-hue: Are you talking about the case when ExtractUVarInt would return insufficient number of bytes error which would happen as it goes over 4 bytes to parse (as it does not know that the limit is 4 bytes), which would prompt my code to return kNeedsMoreData when in fact it is kInvalid?

src/stirling/binaries/stirling_wrapper.cc
```diff
@@ -62,7 +62,7 @@ DEFINE_string(trace, "",
               "Dynamic trace to deploy. Either (1) the path to a file containing PxL or IR trace "
               "spec, or (2) <path to object file>:<symbol_name> for full-function tracing.");
 DEFINE_string(print_record_batches,
-              "http_events,mysql_events,pgsql_events,redis_events,cql_events,dns_events",
+              "http_events,mysql_events,pgsql_events,redis_events,cql_events,dns_events,mqtt_events",
```
ddelnano, line 65: We should hold off on updating this until the MQTT changes are released (one of the final changes).

src/stirling/source_connectors/socket_tracer/protocols/BUILD.bazel
```diff
@@ -45,5 +45,6 @@ pl_cc_library(
         "//src/stirling/source_connectors/socket_tracer/protocols/nats:cc_library",
         "//src/stirling/source_connectors/socket_tracer/protocols/pgsql:cc_library",
         "//src/stirling/source_connectors/socket_tracer/protocols/redis:cc_library",
+        "//src/stirling/source_connectors/socket_tracer/protocols/mqtt:cc_library",
```
ddelnano, line 48: This should be saved for a later change. Most likely once we introduce the "mqtt trace bpf test"

src/stirling/source_connectors/socket_tracer/socket_trace_connector.h
```diff
@@ -65,6 +65,7 @@ DECLARE_int32(stirling_enable_nats_tracing);
 DECLARE_int32(stirling_enable_kafka_tracing);
 DECLARE_int32(stirling_enable_mux_tracing);
 DECLARE_int32(stirling_enable_amqp_tracing);
+DECLARE_int32(stirling_enable_mqtt_tracing);
```
ddelnano, line 68: Changes to this file should also come in a later PR (when the MQTT trace bpf test is added).

src/stirling/source_connectors/socket_tracer/protocols/types.h
```diff
@@ -49,7 +50,8 @@ using FrameDequeVariant = std::variant<std::monostate,
                                        std::deque<redis::Message>,
                                        std::deque<kafka::Packet>,
                                        std::deque<nats::Message>,
-                                       std::deque<amqp::Frame>>;
+                                       std::deque<amqp::Frame>,
+                                       std::deque<mqtt::Message>>;
```
ddelnano, line 54: Changes to this file should also come in a later PR (when the MQTT trace bpf test is added).

src/stirling/source_connectors/socket_tracer/protocols/mqtt/types.h
```diff
@@ -0,0 +1,119 @@
...
+          header_fields_str, payload_str, properties_str);
+    }
+};
+
+//-----------------------------------------------------------------------------
+// Table Store Entry Level Structs
+//-----------------------------------------------------------------------------
+
+/**
+ *  Record is the primary output of the http stitcher.
```
ddelnano, line 98:
```suggestion
 *  Record is the primary output of the MQTT stitcher.
```

src/stirling/source_connectors/socket_tracer/protocols/stitchers.h
```diff
@@ -30,3 +30,4 @@
 #include "src/stirling/source_connectors/socket_tracer/protocols/nats/stitcher.h"  // IWYU pragma: export
 #include "src/stirling/source_connectors/socket_tracer/protocols/pgsql/stitcher.h"  // IWYU pragma: export
 #include "src/stirling/source_connectors/socket_tracer/protocols/redis/stitcher.h"  // IWYU pragma: export
+#include "src/stirling/source_connectors/socket_tracer/protocols/mqtt/stitcher.h" // IWYU pragma: export
```
ddelnano, line 33: We should save this for the stitcher PR.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "src/stirling/source_connectors/socket_tracer/protocols/common/interface.h"
+#include "src/stirling/source_connectors/socket_tracer/protocols/mqtt/types.h"
+
+namespace px {
+namespace stirling {
+namespace protocols {
+
+/**
+ * Parses a single HTTP message from the input string.
```
ddelnano, line 33:
```suggestion
 * Parses a single MQTT message from the input string.
```

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,725 @@
...
+#include <memory>
+#include <string_view>
+#include <iomanip> 
+
+#include "src/common/base/base.h"
+#include "src/stirling/utils/parse_state.h"
+#include "src/stirling/utils/binary_decoder.h"
+#include "src/stirling/source_connectors/socket_tracer/protocols/mqtt/types.h"
+
+#define PX_ASSIGN_OR_RETURN_ERROR(expr, val_or) \
```
ddelnano, line 28: Can we rename this to `PX_ASSIGN_OR_RETURN_NEEDS_MORE_DATA`? Imo `kNeedsMoreData` isn't an error because it's an indicator that we should continue to retry. We also have the same macro defined in the pgsql parser and this would ensure our naming is consistent ([source](https://github.com/pixie-io/pixie/blob/62b8080699371b0bcd83e2c3ddda265d22d0eb40/src/stirling/source_connectors/socket_tracer/protocols/pgsql/parse.cc#L41C28-L41C28)).

src/stirling/source_connectors/socket_tracer/protocols/mqtt/types.h
```diff
@@ -0,0 +1,119 @@
...
+                result += entry.second;
+            }
+            result += ", ";
+        }
+        if (!inputMap.empty()) {
+            result = result.substr(0, result.size() - 2); // Remove the trailing ", "
+        }
+        result += "}";
+        return result;
+    }
```
ddelnano, line 60: Could this be replaced with the [ToJSONString function?](https://github.com/pixie-io/pixie/blob/62b8080699371b0bcd83e2c3ddda265d22d0eb40/src/common/json/json.h#L119) It seems like this is similar to json encoding and we should use `rapidjson` rather than building it by hand.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/types.h
```diff
@@ -0,0 +1,119 @@
...
+        for (const auto& entry : properties) {
+            properties_str += entry.first + ": " + entry.second + ", ";
+        }
+        properties_str += "}";
+
+        std::string payload_str = "{";
+        for (const auto& entry : properties) {
+            payload_str += entry.first + ": " + entry.second + ", ";
+        }
+        payload_str += "}";
```
ddelnano, line 83: Same question regarding json encoding here.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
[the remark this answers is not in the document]
ddelnano: > Would it not work to return kNeedMoreData for all cases of ExtractUVarInt errors?

I don't think it will work because if only 2 bytes are available a UVarInt decode could return a complete value or it could return an insufficient number of bytes error. Since UVarInt's of 2-4 bytes in length are valid in MQTT, we can't use that error as an indicator unless we know that a 4 byte UVarInt is guaranteed to parse.

> Also, I am not sure why it would be easier to change <2 to <5. Wouldn't this eliminate cases where the buffer size is 2? (PINGREQ and PINRESP) are only 2 bytes, with remaining length set to 0.)

Good call. Since `PINGREQ` and `PINGRESP` are a maximum of 2 bytes, that would be a problem and is another case we need to handle.

> Are you talking about the case when ExtractUVarInt would return insufficient number of bytes error which would happen as it goes over 4 bytes to parse

Correct, that would be similar to case 1 from my original comment. If we check for 5 bytes, that would allow us to treat any UVarInt decoding as `kInvalid` since it would guarantee that it would be 5+ bytes in size.

Could we handle the `PINGREQ` and `PINGRESP` cases once the `control_packet_code_flags` variable is populated? Then we can check to see if the buffer contains 5 bytes? I'm thinking something like the following:

```
    if (buf->size() < 2) {
      return ParseState::kNeedsMoreData;
    }

    PX_ASSIGN_OR_RETURN_ERROR(uint8_t control_packet_code_flags, decoder.ExtractBEInt<uint8_t>());
    uint8_t control_packet_code = control_packet_code_flags >> 4;
    uint8_t control_packet_flags = control_packet_code_flags & 0x0F;

    MqttControlPacketType control_packet_type = GetControlPacketType(control_packet_code);
    result->control_packet_type = ControlPacketTypeStrings[control_packet_type];

    if (control_packet_type == MqttControlPacketType::PINGREQ || 
        control_packet_type == MqttControlPacketType::PINGRESP) {
          // Decode UVarInt, check it's 1 byte in length and return success
          // if decoding fails or it is > 1 byte, return kInvalid
    }
    
    // With the control messages less than 5 bytes in size handled, we can do the following validation.
    // This would then allow us to treat any insufficient byte errors from UVarInt decoding as kInvalid since
    // there shouldn't be a UVarInt with more than 4 bytes returned.
    if (buf->size() < 5) {
      return ParseState::kNeedsMoreData;
    }
    PX_ASSIGN_OR(size_t remaining_length, decoder.ExtractUVarInt(), return ParseState::kInvalid)
```

ddelnano:
@ChinmayaSharma-hue just posted my second round of feedback and this is shaping up very nicely! Thanks again for all the hard work on this!

In addition to the comments, can you please update the PR description to match our [GitHub template](https://github.com/pixie-io/pixie/tree/62b8080699371b0bcd83e2c3ddda265d22d0eb40/.github/PULL_REQUEST_TEMPLATE)? This would have been surfaced once the GitHub actions for this PR are permitted to run (I don't have that permission, but I can get that triggered today).

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
ChinmayaSharma-hue: So what that would mean is only in the main parsing function before parsing the variable header or the payload, checks are done to make sure remaining length is equal to the buffer size and then for any subsequent bit extractions, `kInvalid` should be returned? 
Right now for every extraction error `kNeedsMoreData` is being returned. I realise that extraction error happening is unlikely (unless remaining length field is incorrect) because we have already validated that the remaining length is equal to the buffer size so the full packet is already in the buffer. Would you suggest having PX_ASSIGN_OR_RETURN_INVALID to return `kInvalid` in variable header parsing and payload parsing? Because an extraction error in variable header/payload parsing, after we have already validated remaining length to be equal to buffer size, means that the remaining length field was wrong, which would make it an invalid MQTT packet.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -112,6 +112,9 @@ DEFINE_int32(stirling_enable_mux_tracing,
 DEFINE_int32(stirling_enable_amqp_tracing,
              gflags::Int32FromEnv("PX_STIRLING_ENABLE_AMQP_TRACING", px::stirling::TraceMode::On),
              "If true, stirling will trace and process AMQP messages.");
+DEFINE_int32(stirling_enable_mqtt_tracing,
+             gflags::Int32FromEnv("PX_STIRLING_ENABLE_MQTT_TRACING", px::stirling::TraceMode::On),
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Changes in this entire file should be removed and made in a later PR right?

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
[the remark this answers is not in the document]
ChinmayaSharma-hue: Thanks for explaining that in detail and providing the code to make it more clear! I apologise if I took too long to get it.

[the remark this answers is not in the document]
ChinmayaSharma-hue: So currently I have decided to do this,
```
  // Decoding the variable encoding of remaining length field
  size_t remaining_length;
  if (control_packet_type == MqttControlPacketType::PINGREQ ||
      control_packet_type == MqttControlPacketType::PINGRESP) {
    PX_ASSIGN_OR_RETURN_INVALID(remaining_length, decoder.ExtractUVarInt());
    if (remaining_length > 0) {
      return ParseState::kInvalid;
    }
  }

  // Eliminating cases where kNeedsMoreData needs to be returned
  // If buffer size is less tan 4, there are chances that the remaining length is not present in its entirety
  if (decoder.BufSize() < 4) {
      // Checking if buffer is complete
      PX_ASSIGN_OR_RETURN_NEEDS_MORE_DATA(remaining_length, decoder.ExtractUVarInt());
      // if remaining length is greater than 3 (4 if remaining length is included), then incomplete buffer, otherwise buffer is complete
      if (remaining_length > 3) {
          return ParseState::kNeedsMoreData;
      }
  } else {
      PX_ASSIGN_OR_RETURN_INVALID(remaining_length, decoder.ExtractUVarInt());
      if (!VariableEncodingNumBytes(remaining_length).ok()) {
          return ParseState::kInvalid;
      }
  }

  // Making sure buffer is complete according to remaining length
  if (decoder.BufSize() < remaining_length) {
    return ParseState::kNeedsMoreData;
  }
  ```
  This is not perfect since it can cause problems in this section,
  ```
        // if remaining length is greater than 3 (4 if remaining length is included), then incomplete buffer, otherwise buffer is complete
      if (remaining_length > 3) {
          return ParseState::kNeedsMoreData;
      }
```
where I am deciding that buffer is incomplete if remaining length is greater than 3, when it could be that remaining length is greater than 3 simply because of an encoding error and the full data is still present in the buffer. But this works for most of the cases discussed.

Also `<5` is replaced with `<4` because the first byte is already extracted and I am using decoder's buffer.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,678 @@
...
+        result->header_fields["retain"] = (control_packet_flags & 0x1) != 0;
+        result->header_fields["qos"] = (control_packet_flags >> 1) & 0x3;
+    }
+
+    // Decoding the variable encoding of remaining length field
+    size_t remaining_length;
+    std::tie(remaining_length, std::ignore) = VariableLengthEncodingDecoder(&decoder);
+    if (remaining_length < 0) {
+        return ParseState::kInvalid;
+    }
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: I have added this check and then returned kInvalid and all the future decoding errors.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,774 @@
...
+    return 3;
+  }
+  return 4;
+}
+
+ParseState ParseProperties(Message* result, BinaryDecoder* decoder, size_t& properties_length) {
+  uint8_t property_code;
+  while (properties_length > 0) {
+    // Extracting the property code
+    PX_ASSIGN_OR_RETURN_INVALID(property_code, decoder->ExtractBEInt<uint8_t>());
```
ddelnano, line 143: In order to avoid all the `static_cast<uint8_t>` cases, we could use `magic_enum::enum_cast`.

```
uint8_t property_code;
PX_ASSIGN_OR_RETURN_INVALID(property_code, decoder->ExtractBEInt<uint8_t>());
auto property = magic_enum::enum_cast<PropertyCode>(property_code);

if (!property.has_value()) {
  return ParseState::kInvalid;
}

switch (property.value()) {
    case PropertyCode::PayloadFormatIndicator: {
...

```

Given there are quite a few cases, I think it will help with the readability.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,774 @@
...
+  BinaryDecoder decoder(*buf);
+
+  // Parsing the fixed header
+  // Control Packet Type extracted from first four bits of the first byte
+  PX_ASSIGN_OR_RETURN_NEEDS_MORE_DATA(uint8_t control_packet_code_flags,
+                                      decoder.ExtractBEInt<uint8_t>());
+  uint8_t control_packet_code = control_packet_code_flags >> 4;
+  uint8_t control_packet_flags = control_packet_code_flags & 0x0F;
+
+  MqttControlPacketType control_packet_type = GetControlPacketType(control_packet_code);
```
ddelnano, line 701: `GetControlPacketType` can be replaced with `magic_enum::enum_cast<MqttControlPacketType>`. If the resulting `std::optional<MqttControlPacketType>` doesn't have a value that should be equivalent to the `MqttControlPacketType::INVALID` type. It seems like we would want to return `kInvalid` for that case anyway?

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,774 @@
...
+  PUBREC = 5,
+  PUBREL = 6,
+  PUBCOMP = 7,
+  SUBSCRIBE = 8,
+  SUBACK = 9,
+  UNSUBSCRIBE = 10,
+  UNSUBACK = 11,
+  PINGREQ = 12,
+  PINGRESP = 13,
+  DISCONNECT = 14,
```
ddelnano, line 52: Is there a reason you didn't want to add the `AUTH` control packet type? It seems like that is the only one omitted from this list.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
ddelnano: Your latest changes are consistent with what I described above, so this lgtm.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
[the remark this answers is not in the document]
ddelnano: I see. Thanks for the detailed explanation on those cases.

If we can special case any control codes that are known to be short in length (like we did with `PINGREQ` and `PINGRESP`, it may reduce the cases of treating some of these incorrectly. For now I'm fine with keeping the implementation as is unless you see an opportunity for special casing any of them.

ddelnano (commented):
@ChinmayaSharma-hue I'm hoping this will be my last round of comments. I know there have been a variety of things to address, but making this initial implementation solid will be worth it in the long run!

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
ChinmayaSharma-hue: I forgot to include this, sorry about this. I have added this with tests in the new commit.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
ChinmayaSharma-hue: I have replaced this with magic_enum.

ChinmayaSharma-hue:
> @ChinmayaSharma-hue I believe this will be my last round of comments. I know there have been a variety of things to address, but making this initial implementation solid will be worth it in the long run!
> 
> Also please check the [linter build errors](https://github.com/pixie-io/pixie/actions/runs/6822021844/job/18589702739?pr=1756). Those will need to be resolved as well.

I haven't gotten to making all the required changes as of yet including the linter error fixes, I will let you know as soon as all the changes are made. There were some instances where I made silly errors which should have been fixed without alerting on your part. I hope to make less of these in the future! Thanks a lot for the guidance and support.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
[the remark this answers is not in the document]
ChinmayaSharma-hue: Special casing only works for PINGREQ and PINGRESP as we can guarantee their sizes to be below 5. With all the other control packets there are chances that properties are present in variable header that can make them bigger.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,774 @@
...
+                       result->header_fields["variable_header_length"];
+      while (payload_length > 0) {
+        PX_ASSIGN_OR_RETURN_INVALID(reason_code, decoder->ExtractBEInt<uint8_t>());
+        if (result->payload["reason_code"].empty()) {
+          result->payload["reason_code"] += std::to_string(reason_code);
+        } else {
+          result->payload["reason_code"] += ", " + std::to_string(reason_code);
+        }
+        payload_length -= 1;
+      }
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Wouldn't the else statement be executed once result->payload["reason_code"] is filled with one reason code? The UNSUBACK control type packet can contain multiple reason codes in its payload, and I took this approach to form a readable string as value for payload["reason_code"]. It would start with the first reason code when the string is empty and then concatenate the rest of them to this string.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,774 @@
...
+          (result->header_fields.find("variable_header_length") == result->header_fields.end())) {
+        return ParseState::kInvalid;
+      }
+
+      result->payload["reason_code"] = "";
+      payload_length = result->header_fields["remaining_length"] -
+                       result->header_fields["variable_header_length"];
+      while (payload_length > 0) {
+        PX_ASSIGN_OR_RETURN_INVALID(reason_code, decoder->ExtractBEInt<uint8_t>());
+        if (result->payload["reason_code"].empty()) {
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Yeah I have now fixed this in the rest of the code too.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,774 @@
...
+      if ((result->header_fields.find("remaining_length") == result->header_fields.end()) ||
+          (result->header_fields.find("variable_header_length") == result->header_fields.end())) {
+        return ParseState::kInvalid;
+      }
+
+      result->payload["topic_filter"] = "";
+      payload_length = result->header_fields["remaining_length"] -
+                       result->header_fields["variable_header_length"];
+      while (payload_length > 0) {
+        PX_ASSIGN_OR_RETURN_INVALID(topic_filter_length, decoder->ExtractBEInt<uint16_t>());
```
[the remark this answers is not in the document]
ChinmayaSharma-hue: Yeah I have now fixed this in the rest of the code too.

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,774 @@
...
+        PX_ASSIGN_OR_RETURN_INVALID(std::string_view topic_filter,
+                                    decoder->ExtractString(topic_filter_length));
+        if (result->payload["topic_filter"].empty()) {
+          result->payload["topic_filter"] += std::string(topic_filter);
+        } else {
+          result->payload["topic_filter"] += ", " + std::string(topic_filter);
+        }
+        payload_length -= (2 + topic_filter_length);
+      }
```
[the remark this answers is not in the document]
ddelnano: Sorry, I misunderstood this on my previous read. This lgtm 👍

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,750 @@
...
+  SubscriptionIdentifiersAvailable = 0x29,
+  SharedSubscriptionAvailable = 0x2A
+};
+
+constexpr int kMaxVarInt8 = 128;
+constexpr int kMaxVarInt16 = 16384;
+constexpr int kMaxVarInt24 = 2097152;
+constexpr int kMaxVarInt32 = 268435456;
+
+static inline StatusOr<size_t> VariableEncodingNumBytes(unsigned long integer) {
```
ddelnano, line 93:
```suggestion
static inline StatusOr<size_t> VariableEncodingNumBytes(uint64_t integer) {
```

src/stirling/source_connectors/socket_tracer/protocols/mqtt/parse.cc
```diff
@@ -0,0 +1,750 @@
...
+        PX_ASSIGN_OR_RETURN_INVALID(std::string_view correlation_data,
+                                    decoder->ExtractString(property_length));
+        result->properties["correlation_data"] =
+            BytesToString<bytes_format::HexAsciiMix>(correlation_data);
+        ;
+        properties_length -= property_length;
+        break;
+      }
+      case PropertyCode::SubscriptionIdentifier: {
+        PX_ASSIGN_OR_RETURN_INVALID(unsigned long subscription_id, decoder->ExtractUVarInt());
```
ddelnano, line 171:
```suggestion
        PX_ASSIGN_OR_RETURN_INVALID(uint64_t subscription_id, decoder->ExtractUVarInt());
```

ddelnano (commented):
One comment regarding one of the changes from the previous round of feedback and two suggestions for fixing the `runtime/int` linter warnings.

ChinmayaSharma-hue:
I have fixed the linting errors (I ran arc lint locally), and I have fixed the other issues with the code. Sorry for the delay.

ddelnano:
@pixie-io/maintainers can we kick off the github actions for this PR? This is ready for its final review and I will be approving once the build passes.

ddelnano (approved):
Thanks so much for this contribution @ChinmayaSharma-hue and looking forward to working together on the upcoming ones!

JamesMBartlett (approved):

JamesMBartlett:
Thanks for the contribution @ChinmayaSharma-hue. We require all of our commits to be GPG signed. Could you please follow this guide and sign your commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

ChinmayaSharma-hue:
The commits are now gpg signed and verified.

PR #1760, opened 2023-11-02
Clear the frame contents from the buffer in the case that we return k…

kpattaswamy:
Summary: This PR removes the frame's contents from the buffer in the case we return `kIgnored` when the frame is a type (opcode) we do not parse. When running the BPF test we noticed that the program would stall when the parser encountered a frame with an opcode it does not support. This was due to to the parser returning `kIgnored` to `ParseFramesLoop` and it not moving the buffer forward before calling `ParseFrame` again. This change will update the buffer position before returning `kIgnored` to `ParseFramesLoop` so that the remaining frames in the buffer can be parsed.

Type of change: /kind bug

Test Plan: Modified the existing test checking for the unsupported opcode type.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse.cc
@@ -67,16 +67,19 @@ ParseState ParseFrame(message_type_t type, std::string_view* buf, Frame* frame, 
   // Parser will ignore Op Codes that have been deprecated/removed from version 5.0 onwards.
   if (!(frame_type == Type::kOPMsg || frame_type == Type::kOPCompressed ||
         frame_type == Type::kReserved)) {
+    buf->remove_prefix(frame->length);
     return ParseState::kIgnored;
   }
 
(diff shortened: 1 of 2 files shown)


src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse.cc
```diff
@@ -67,16 +67,19 @@ ParseState ParseFrame(message_type_t type, std::string_view* buf, Frame* frame,
   // Parser will ignore Op Codes that have been deprecated/removed from version 5.0 onwards.
   if (!(frame_type == Type::kOPMsg || frame_type == Type::kOPCompressed ||
         frame_type == Type::kReserved)) {
```
ddelnano, line 69: Can we remove `Type::kReserved` and `Type::kOPCompressed` from this if statement to avoid adding the `else if` branch added in this PR?

I ran the updated unit test and it seems to pass still.

kpattaswamy: Will modify the comment as well

ddelnano (approved):

aimichelle (approved):

PR #1732, opened 2023-10-11
[Fix CQL Sticher 4/4] Move StreamID assignment to ParseFramesLoop

benkilimnik:
Summary: Populates a map of streamIDs to deque of frames in `ParseFramesLoop` instead of `ParseFrames`. This should provide a small efficiency boost, as we won't have to loop over the frames twice. This PR relies on #1761 due to the way timestamps are updated using `ParseResult`.

Related issues: https://github.com/pixie-io/pixie/issues/1375

Type of change: /kind cleanup

Test Plan: Updated parsing tests to use new interface

PR #1761, opened 2023-11-03
Move timestamp monotonicity check to `DataStreamBuffer::Head()` to accommodate in #1733

benkilimnik:
Summary: Preemptively adapts the timestamp monotonicity change introduced in #1733 to the last stitcher api PR #1732, which modifies `ParseResult.frame_positions` to be an unordered `flat_hash_map`. This changes the order in which `GetTimestamp` is called because we are now iterating over an unordered map of streamIDs to positions when matching timestamps with the parsed frames in the [event_parser](https://github.com/pixie-io/pixie/blob/e6bfab707f1f4871f4b7b8ed53321ec9e7b5807d/src/stirling/source_connectors/socket_tracer/protocols/common/event_parser.h#L138C29-L138C36).

Previously, we were always iterating over the `frame_position` with the oldest timestamp first, meaning that `prev_timestamp_` in the datastream buffer was set correctly. With `frame_positions` being an unordered map, we no longer have this guarantee.

To address this, we move the monotonicity check to the `Head()` implementation of the datastream buffer and enforce increasing timestamps for the contiguous chunk returned by `Head()` only.

Type of change: /kind bug

Test Plan: Extended the data stream buffer test + existing targets.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/common/always_contiguous_data_stream_buffer_impl.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/common/always_contiguous_data_stream_buffer_impl.cc
@@ -263,14 +263,41 @@ std::map<size_t, size_t>::const_iterator AlwaysContiguousDataStreamBufferImpl::G
   return iter;
 }
 
-std::string_view AlwaysContiguousDataStreamBufferImpl::Get(size_t pos) const {
+void AlwaysContiguousDataStreamBufferImpl::EnforceTimestampMonotonicity(size_t pos,
+                                                                        size_t chunk_end) {
+  // Get timestamp for chunk which is <= pos.
+  auto it = timestamps_.upper_bound(pos);
+  if (it == timestamps_.begin()) {
+    return;
+  }
+  --it;
+
+  // Loop from chunk_start up to but not including chunk_end.
+  // The next element after chunk_end should not be part of the contiguous block.
+  prev_timestamp_ = 0;
+  for (; it != timestamps_.end() && it->first < chunk_end; ++it) {
+    if (prev_timestamp_ > 0 && it->second < prev_timestamp_) {
+      LOG(WARNING) << "For chunk pos " << it->first << ", "
+                   << "detected non-monotonically increasing timestamp " << it->second
+                   << ". Adjusting to previous timestamp + 1: " << prev_timestamp_ + 1;
+      it->second = prev_timestamp_ + 1;
+    }
+    prev_timestamp_ = it->second;
+  }
+}
+
+std::string_view AlwaysContiguousDataStreamBufferImpl::Get(size_t pos) {
   auto iter = GetChunkForPos(pos);
   if (iter == chunks_.cend()) {
     return {};
   }
 
-  size_t chunk_pos = iter->first;
-  size_t chunk_size = iter->second;
+  size_t chunk_pos = iter->first;    // start of contiguous head
+  size_t chunk_size = iter->second;  // size of contiguous (already merged in Add)
+
+  // since we only call Get() in Head() and the event parser fully processes a contiguous head,
+  // we need only enforce timestamp monotonicity once per head.
+  EnforceTimestampMonotonicity(chunk_pos, chunk_pos + chunk_size);
 
   ssize_t bytes_available = chunk_size - (pos - chunk_pos);
   DCHECK_GT(bytes_available, 0);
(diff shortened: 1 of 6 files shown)


src/stirling/source_connectors/socket_tracer/protocols/common/always_contiguous_data_stream_buffer_impl.cc
```diff
@@ -263,14 +263,41 @@ std::map<size_t, size_t>::const_iterator AlwaysContiguousDataStreamBufferImpl::G
...
+  --it;
+
+  // Loop from chunk_start up to but not including chunk_end.
+  // The next element after chunk_end should not be part of the contiguous block.
+  prev_timestamp_ = 0;
+  for (; it != timestamps_.end() && it->first < chunk_end; ++it) {
+    if (prev_timestamp_ > 0 && it->second < prev_timestamp_) {
+      LOG(WARNING) << "For chunk pos " << it->first << ", "
+                   << "detected non-monotonically increasing timestamp " << it->second
+                   << ". Adjusting to previous timestamp + 1: " << prev_timestamp_ + 1;
```
ddelnano, line 282: I realized we missed this on the previous review, but our [style guide](https://github.com/pixie-io/pixie/blob/main/styleguide/cc_style.md#formatted-strings) disallows `<<` for string formatting. We should use `absl::Substitute` here instead.

src/stirling/source_connectors/socket_tracer/protocols/common/lazy_contiguous_data_stream_buffer_impl.cc
```diff
@@ -151,9 +151,17 @@ void LazyContiguousDataStreamBufferImpl::MergeContiguousEventsIntoHead() {
...
 
   it = events_.begin();
+  // end_it stopped at the first non-contiguous event (at the end of current head)
   while (it != end_it) {
     size_t event_size = it->second.data.size();
     memcpy(new_buffer->Data() + offset, it->second.data.data(), event_size);
+    // Ensure that the event timestamps are monotonically increasing for a given contiguous head
+    if (prev_timestamp_ > 0 && it->second.timestamp < prev_timestamp_) {
+      LOG(WARNING) << "Detected non-monotonically increasing timestamp " << it->second.timestamp
+                   << ". Adjusting to previous timestamp + 1: " << (prev_timestamp_ + 1);
```
ddelnano, line 161: Same comment here about the string formatting.

src/stirling/source_connectors/socket_tracer/protocols/common/data_stream_buffer_test.cc
```diff
@@ -152,6 +152,15 @@ TEST_P(DataStreamBufferTest, Timestamp) {
   EXPECT_EQ(stream_buffer.Head(), "123456789");
   EXPECT_OK_AND_EQ(stream_buffer.GetTimestamp(8),
                    5);  // timestamp is adjusted to previous timestamp + 1
```
ddelnano, line 154: Is the test code we added last time relevant still? Do we need 3 non monotonic `Add` calls or are the ones you added in this PR enough?

src/stirling/source_connectors/socket_tracer/protocols/common/lazy_contiguous_data_stream_buffer_impl.cc
```diff
@@ -151,9 +151,19 @@ void LazyContiguousDataStreamBufferImpl::MergeContiguousEventsIntoHead() {
   }
 
   it = events_.begin();
+  // end_it stopped at the first non-contiguous event (at the end of current head)
```
ddelnano, line 154: Are we guaranteed to call `MergeContiguousEventsIntoHead` for the situations we care about? I noticed that `Head` only calls this function if `IsHeadAndEventsMergeable` is true ([source](https://github.com/pixie-io/pixie/blob/62b8080699371b0bcd83e2c3ddda265d22d0eb40/src/stirling/source_connectors/socket_tracer/protocols/common/lazy_contiguous_data_stream_buffer_impl.cc#L169)).

benkilimnik: `IsHeadAndEventsMergeable` is true whenever we have events that line up with the end of the previous `head_` such that we can merge at least one event from `events_`.

Since we only check for monotonicity when merging events into `head_`, any existing `head_` will have monotonic timestamps.

Does that answer your question?

ddelnano: Yea, just wanted to double check my understanding of that merge logic 👍

ddelnano (approved):

etep (approved):

vihangm (approved):

PR #1762, opened 2023-11-03
Add support for connecting standalone PEM in pxapi.

etep:
Summary: Add support for connecting standalone PEM in pxapi.

Type of change: /kind feature

Test Plan: Added an example program to connect standalone PEM.


RagalahariP (approved):
Looks good.

src/api/go/pxapi/client.go
```diff
@@ -89,7 +92,11 @@ func (c *Client) init(ctx context.Context) error {
 	tlsConfig := &tls.Config{InsecureSkipVerify: c.disableTLSVerification}
 	creds := credentials.NewTLS(tlsConfig)
 
-	conn, err := grpc.Dial(c.cloudAddr, grpc.WithTransportCredentials(creds))
+	if c.insecureDirect {
+		creds = insecure.NewCredentials()
```
JamesMBartlett, line 96: @ddelnano I know you said there was some internal pentest that led to your `WithDisableTLSVerification` changes. Do you think this option would come up on a similar pentest

ddelnano: There isn't another planned pentest, but their goal was to remove any areas where TLS verification or insecure connections were used as a default. In addition to that they wanted the method to disable it documented.

This should be fine since it isn't the default and the `ClientOption` godoc string explains that it disables TLS.

JamesMBartlett (approved):

Issue #1739, opened 2023-10-16
False positive Github build/test status for 4.14 kernel


benkilimnik:
**Describe the bug**
In https://github.com/pixie-io/pixie/commit/2a2349f4853c90848437441ceed4bad634505767, many of the 4.14 kernel tests have [BPF compilation errors](https://app.buildbuddy.io/invocation/075de8d0-d161-4ff6-8b74-123c6da66778#): `HINT: bpf_get_current_cgroup_id missing (added in Linux 4.18)`. GitHub shows these tests as [passing despite the failures](https://github.com/pixie-io/pixie/actions/runs/6366478258/job/17320729511). As a result, the PR was merged despite the build errors.

**To Reproduce**
1. View [BPF compilation errors](https://app.buildbuddy.io/invocation/075de8d0-d161-4ff6-8b74-123c6da66778#) in buildbuddy

**Expected behavior**
GitHub CI should show failing builds/tests.

**Screenshots**

<img width="1149" alt="image" src="https://github.com/pixie-io/pixie/assets/47846691/008b2cb7-1515-47c0-a4c0-b385ae74ce80">

<img width="1687" alt="image" src="https://github.com/pixie-io/pixie/assets/47846691/202be716-374d-4224-bbb3-87f74aa86c82">

ddelnano:
This false positive was due to the fact that the BEP upload failed when there were failing tests. This is a case we explicitly ignore the error code for ([source](https://github.com/pixie-io/pixie/blob/62b8080699371b0bcd83e2c3ddda265d22d0eb40/scripts/bazel_ignore_codes.sh#L25-L26)).

<img width="1723" alt="Screenshot 2023-11-07 at 8 51 10 AM" src="https://github.com/pixie-io/pixie/assets/5855593/09373078-7724-43c8-8916-11d4b5f9f23f">

PR #1766, opened 2023-11-08
Parse test results XML

vihangm:
Summary: Our bazel test script ignores some non-zero status. This
is to ensure that BES upload failures don't fail the build. However
if you have a test failure on a run with a BES upload failure, the
test failure gets ignored. Parsing the test output should help
catch broken tests.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Check Github Actions output on this PR.


ddelnano (approved):
Can we add #1739 as a relevant issue in the PR description?

JamesMBartlett (approved):

PR #1776, opened 2023-11-20
Add support to ignore MongoDB handshaking frames

kpattaswamy:
Summary: When end to end testing MongoDB tracing, the parser encountered `OP_MSG` handshaking frames and was not able to parse them. This was due to the old parser searching for a CRUD command in the `OP_MSG` frame's top level key in the payload and `OP_MSG` handshaking frames not containing a key the old parser could interpret.

This PR adds support to parse top level handshaking keys and discards those frames at stitching time. It also correctly prefixes `const` variables in the `types.h` file.

The motivation to parse these frames instead of immediately ignoring them at parsing time is to account for the different possibilities of responses for a handshaking request. A handshaking request could lead to a `more_to_come` response where each frame in the `more_to_come` response may not be identifiable as a handshaking frame. It's also possible that the response of a handshaking request may contain an `ok` key which can be misidentified as a non handshaking response and left stale in the map of response deques. Identifying all of the handshaking request/response frames at stitching time but not inserting it to the records would ensure that all handshaking frames are cleared from the map and not pushed to the data table.

Type of change: /kind bug

Test Plan: Added a stitcher test

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
@@ -162,6 +162,12 @@ RecordsWithErrorCount<mongodb::Record> StitchFrames(
       resp_frame.consumed = true;
       FlattenSections(&req_frame);
       FlattenSections(&resp_frame);
+
+      // Ignore stitching the request/response if either one is a handshaking frame.
+      if (req_frame.is_handshake || resp_frame.is_handshake) {
+        break;
+      }
+
       records.push_back({std::move(req_frame), std::move(resp_frame)});
       break;
     }
--- a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
@@ -165,6 +165,8 @@ RecordsWithErrorCount<mongodb::Record> StitchFrames(
 
       // Ignore stitching the request/response if either one is a handshaking frame.
       if (req_frame.is_handshake || resp_frame.is_handshake) {
+        req_frame.consumed = true;
+        resp_frame.consumed = true;
         break;
       }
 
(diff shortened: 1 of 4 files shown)


src/stirling/source_connectors/socket_tracer/protocols/mongodb/types.h
```diff
@@ -116,6 +121,9 @@ constexpr int32_t kMaxBSONObjSize = 16000000;
  * https://github.com/mongodb/specifications/blob/e09b41df206f9efaa36ba4c332c47d04ddb7d6d1/source/message/OP_MSG.rst#command-arguments-as-payload
  *
  * There can be 0 or more documents in a section of kind 1 without a separator between them.
+ *
+ * Information about MongoDB handshaking messages can be found here:
+ * https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst
```
ddelnano, line 126: Can we make this a permalink so that any future changes to the file will link to the current copy?

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
```diff
@@ -162,6 +162,12 @@ RecordsWithErrorCount<mongodb::Record> StitchFrames(
       resp_frame.consumed = true;
       FlattenSections(&req_frame);
       FlattenSections(&resp_frame);
+
+      // Ignore stitching the request/response if either one is a handshaking frame.
+      if (req_frame.is_handshake || resp_frame.is_handshake) {
+        break;
+      }
```
ddelnano, line 169: I have a slight preference to set the `consumed` member on these frames to be more explicit about the handling of these frames. I know that our logic still works either way since we clear up stale frames.

ddelnano (commented):
Two minor comments but otherwise lgtm.

src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
kpattaswamy: Do you mean setting the `consumed` field in the request/response frame to `true` again before breaking?

ddelnano: Yea.

ddelnano (approved):

vihangm (approved):

Issue #1755, opened 2023-10-27
Capture partial frames for syscalls which exceed bpf `LOOP`/`CHUNK` limits


benkilimnik:
**Problem Description**
Pixie is unable to capture syscalls with an [iovcnt](https://github.com/pixie-io/pixie/blob/755f138bac4e97f5628a44f58454bfce2ab2f9ad/src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c#L38) `>42` or a [message size](https://github.com/pixie-io/pixie/blob/755f138bac4e97f5628a44f58454bfce2ab2f9ad/src/stirling/source_connectors/socket_tracer/bcc_bpf_intf/socket_trace.h#L130) `>120 KiB` . 

These variables are set conservatively to keep the instruction count below BPF's limit for version 4 kernels (4096 per probe). These limits, however, result in data loss and incomplete syscall tracing. For example, in a [community-shared NodeJS application](https://github.com/vincent99/acornfiles/blob/d556fc28b49f49c65a24ff81b125088c9cc1eb70/netgraph/index.js) transferring just 10kB of data, the iovec array contained 257 entries, which is well beyond the current `LOOP_LIMIT` of 42. We've also seen the message size (`CHUNK_LIMIT`) exceeded in k8ssandra deployments. This is very likely an issue across all protocols. 

**Proposed Solutions**
1. Dynamically increase the loop limit for newer kernels with higher instruction limits (1 million for kernels > 5.1). This could mitigate the issue, though it would likely persist for large messages/iovecs. (This approach could be combined with option 2. to increase both the amount of data and number of frames Pixie can trace)
- Each PEM now tracks its kernel version due to changes introduced in [#1685](https://github.com/pixie-io/pixie/pull/1685). We could pass this version in as a compile time flag / preprocessor directive.
- Note that even if bpf were able to trace everything for large messages, Pixie would still truncate the data (e.g. based on `FLAGS_max_body_bytes` for HTTP). However, capturing complete metadata could still be invaluable, conveying headers, response codes, and other important information. 

--------

2. For each event where data loss occurs due to `LOOP/CHUNK` limits, pass metadata to the event parser, which attempts to process a partial frame. For this to work, protocol parsers must be modified to work lazily, parsing as far as possible and returning a new parseState `kPartialFrame` when they've processed enough bytes to capture essential metadata. 

[...]

PR #1787, opened 2023-11-29
[Parse incomplete chunks 2/9] Add chunk type annotations, record gap sizes and filler allocations

benkilimnik:
Summary: Adds metadata to `socket_data_event_t` to keep track of gaps from bpf along with the number of bytes we miss. This metadata will be fed to prometheus in a future PR to keep track of the prevalence and distribution of gaps on a per protocol basis. More details in #1755.

Type of change: /kind feature

Test Plan: Existing targets


src/stirling/source_connectors/socket_tracer/bcc_bpf_intf/socket_trace.hpp
```diff
@@ -134,28 +146,51 @@ struct SocketDataEvent {
...
+    // Note that msg_size - msg_buf_size != bytes_missed in the case where we exceed LOOP_LIMIT
+    // in perf_submit_iovecs, because one call to perf_submit_buf takes only the size of the current
+    // iovec into account, ommitting the rest of the iovecs which could not be submitted.
+    // As a result, we need to use bytes_missed to determine the size of the filler event.
+
+    // For kernels < 5.1, we cannot track the bytes missed in socket_trace.c properly and thus
+    // preserve the previous behavior of encoding the bytes missed via the msg_size.
+    // If our loop and chunk limits are at most 42 and 4, then we know that we can
+    // stay below the verifier instruction limit for kernels < 5.1.
+    if (LOOP_LIMIT <= 42 && CHUNK_LIMIT <= 4) {
```
etep, line 158: Please consider here and in the eBPF c code, creating a semantically meaningful version of this. Perhaps:
```
const bool kernelNewerThan5dot1 = LOOP_LIMIT > 42 || CUNK_LIMIT > 4;
```

benkilimnik: Added

JamesMBartlett:
@etep Could you give an approval if you think this is good to go

@benkilimnik If you've done additional testing to ensure that this works (eg. on kernels older than 5.1 and kernels newer than 5.1), could you add it to the Test Plan in the PR description. If we haven't done that, I think it would be good to ensure that things work end to end, when we change bpf code.

PR #1795, opened 2023-11-29
Raise loop and chunk limit for newer kernels

benkilimnik:
Summary: Dynamically increase the loop limit for newer kernels with higher instruction limits (1 million for kernels > 5.1) by 21x to reduce data loss and raise ingest. More details in #1755.

One open question is whether we want to add vizier flag to toggle this behavior in case there are unforseen performance bottlenecks for certain clusters.

Type of change: /kind feature

Test Plan: Existing targets + perf/demo tests outlined in #1755.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
+++ b/src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
@@ -31,6 +31,7 @@
 #include <google/protobuf/text_format.h>
 #include <google/protobuf/util/delimited_message_util.h>
 #include <magic_enum.hpp>
+#include "src/common/system/kernel_version.h"
 
 #include "src/common/base/base.h"
 #include "src/common/base/utils.h"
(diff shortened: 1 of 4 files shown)


src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -431,6 +432,20 @@ auto SocketTraceConnector::InitPerfBufferSpecs() {
 }
 
 Status SocketTraceConnector::InitBPF() {
+  // set BPF loop limit and chunk limit based on kernel version
+  auto kernel = system::GetCachedKernelVersion();
+  int loop_limit = 42;
```
JamesMBartlett, line 437: These should be gflags.

benkilimnik: Changed to gflags

etep (approved):

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -431,6 +441,18 @@ auto SocketTraceConnector::InitPerfBufferSpecs() {
 }
 
 Status SocketTraceConnector::InitBPF() {
+  // set BPF loop limit and chunk limit based on kernel version
+  auto kernel = system::GetCachedKernelVersion();
+  if (kernel.version >= 5 || (kernel.version == 5 && kernel.major_rev >= 1)) {
+    // Kernels >= 5.1 have higher BPF instruction limits (1 million for verifier).
+    // This enables a 21x increase to our loop and chunk limits
+    FLAGS_stirling_bpf_loop_limit = 882;
+    FLAGS_stirling_bpf_chunk_limit = 84;
```
JamesMBartlett, line 450: How did you pick these numbers? How close are we to the instruction limit with these numbers?

JamesMBartlett (approved):

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
benkilimnik: I manually raised the loop/chunk limits to see when the verifier would [throw an error](https://github.com/iovisor/bcc/blob/1a784cfe1a2936f2c5cf0a8ee045dcdfd841ba5f/src/cc/libbpf.c#L917) and chose values just below that upper bound, which is around 22x our previous limits.
```sh
bpf: Argument list too long. Program too large (... insns), at most 4096 insns
```
Confusingly, the `bpf(2)` syscall returns the same error whether the program size or its complexity exceeds the limits. See [stackoverflow](https://stackoverflow.com/questions/70147464/program-too-large-threshold-greater-than-actual-instruction-count)

PR #1799, opened 2023-11-30
Add the pxl script to visualize MongoDB data

kpattaswamy:
Summary: This PR adds the pxl script to visualize the MongoDB data table on the UI.

This is how the table looks like
Screenshot 2023-11-30 at 11 28 47 AM

Type of change: /kind feature

Test Plan: Ran the pxl script with `vis.json` in the scratch pad section of the UI

### DIFF
--- a/src/pxl_scripts/px/mongodb_data/mongodb_data.pxl
+++ b/src/pxl_scripts/px/mongodb_data/mongodb_data.pxl
@@ -1,0 +1,116 @@
+# Copyright 2018- The Pixie Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+''' MongoDB Data Tracer
+
+Shows the most recent MongoDB messages in the cluster.
+'''
+import px
+
+
+def mongodb_data(start_time: str, source_filter: str, destination_filter: str, num_head: int):
+
+    df = px.DataFrame(table='mongodb_events', start_time=start_time)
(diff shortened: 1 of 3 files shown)


src/pxl_scripts/px/mongodb_data/mongodb_data.pxl
```diff
@@ -0,0 +1,116 @@
...
+def mongodb_data(start_time: str, source_filter: str, destination_filter: str, num_head: int):
+
+    df = px.DataFrame(table='mongodb_events', start_time=start_time)
+    df = add_source_dest_columns(df)
+
+    # Filter out entities as specified by the user.
+    df = df[px.contains(df.source, source_filter)]
+    df = df[px.contains(df.destination, destination_filter)]
+
+    # Add additional filters below:
```
etep, line 33: This might be a stray comment.

src/pxl_scripts/px/mongodb_data/mongodb_data.pxl
```diff
@@ -0,0 +1,116 @@
...
+
+
+def mongodb_data(start_time: str, source_filter: str, destination_filter: str, num_head: int):
+
+    df = px.DataFrame(table='mongodb_events', start_time=start_time)
+    df = add_source_dest_columns(df)
+
+    # Filter out entities as specified by the user.
+    df = df[px.contains(df.source, source_filter)]
+    df = df[px.contains(df.destination, destination_filter)]
```
etep, line 31: This could be my knowledge gap, but what happens when the filter string is empty? Or is the default something like `*`?

kpattaswamy: I believe the default source filter is already to set to empty so it would display all origins of mongo requests

src/pxl_scripts/px/mongodb_data/mongodb_data.pxl
kpattaswamy: I think the purpose of that comment is to point out the location in the script where a user would add additional filters if necessary. That's what my understanding was when looking through other the pxl scripts

src/pxl_scripts/px/mongodb_data/mongodb_data.pxl
ddelnano: That's correct. Another thing I wanted to mention was that was something we didn't have a compiler optimization for until recently to prune "empty contains" calls ([pixie#1679](https://github.com/pixie-io/pixie/pull/1679))

PR #1807, opened 2023-12-21
[Add Local Addr & Port 2/3] Add columns to socket tracer

benkilimnik:
Summary: Adds data columns for local IP address and port to the socket tracer, which are populated by #1808 and #1809. This will support standalone pem entity relationships.

Type of change: /kind feature

Test Plan: Existing targets

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/conn_tracker.h
+++ b/src/stirling/source_connectors/socket_tracer/conn_tracker.h
@@ -338,6 +338,11 @@ class ConnTracker : NotCopyMoveable {
   const SockAddr& remote_endpoint() const { return open_info_.remote_addr; }
 
   /**
+   * Get local IP endpoint of the connection.
+   */
+  const SockAddr& local_endpoint() const { return open_info_.local_addr; }
+
+  /**
    * Get the connection information (e.g. remote IP, port, PID, etc.) for this connection.
    */
   const SocketOpen& conn() const { return open_info_; }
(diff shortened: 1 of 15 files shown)


src/stirling/source_connectors/socket_tracer/conn_tracker.h
```diff
@@ -799,6 +806,8 @@ std::string DebugString(const ConnTracker& c, std::string_view prefix) {
   info += absl::Substitute("state=$0\n", magic_enum::enum_name(c.state()));
   info += absl::Substitute("$0remote_addr=$1:$2\n", prefix, c.remote_endpoint().AddrStr(),
                            c.remote_endpoint().port());
+  info += absl::Substitute("$0remote_addr=$1:$2\n", prefix, c.local_endpoint().AddrStr(),
+                           c.local_endpoint().port());
```
etep, line 810: I think this was supposed to have the text `local_addr=$1:$2`.

benkilimnik: Fixed

etep (approved):

oazizi000:
We've historically relied on only capturing the remote addr, and using the PID along with K8s metadata to determine the local IP, but I do see value in capturing the local IP directly. So this looks good, except that this can't be landed in isolation. Will be looking at the follow-up PRs soon.

JamesMBartlett (approved):
I think this will lead to empty strings in the relevant tables, but as long as we don't release before the other diffs in this series land I think its ok.

JamesMBartlett:
Will hold off on landing until I have more confidence the other diffs will land.

oazizi000:
> I think this will lead to empty strings in the relevant tables, but as long as we don't release before the other diffs in this series land I think its ok.

The other way to deal with this is to order things differently. It should be safe to do the data collection in bpf without actually pushing to the tables. And then have this diff come after that. Might require a bit of reshuffling, but logically that's how I would go about it (start from BPF code and layer diffs all the way up to pushing into the tables).

benkilimnik:
Thanks for the feedback @oazizi000 @JamesMBartlett 
I've reordered the PRs such that data collection in bpf comes first #1808

Issue #932, opened 2023-03-01
security_socket_sendmsg not exist and can't attach kprobe


ChenLingPeng:
**Describe the bug**
when kernel build with CONFIG_SECURITY_NETWORK not set, security_socket_sendmsg is inlined. 
In this situation, socket_tracer will not work

```
I20230301 07:05:24.697929 26969 bcc_wrapper.cc:170] Initializing BPF program ...
I20230301 07:05:33.953138 26969 scoped_timer.h:48] Timer(init_bpf_program) : 9.26 s
cannot attach kprobe, probe entry may not exist
W20230301 07:05:37.417877 26969 stirling.cc:411] Source Connector (registry name=socket_tracer) not instantiated, error: Internal : Unable to attach kprobe for security_socket_sendmsg using probe_entry_security_socket_sendmsg
```

using sock_sendmsg maybe another option

**To Reproduce**

**Expected behavior**

**Screenshots**
If applicable, add screenshots to help explain your problem. Please make sure the screenshot does not contain any sensitive information such as API keys or access tokens.

**Logs**
Please attach the logs by running the following command:
```
./px collect-logs
```

**App information (please complete the following information):**
- Pixie version
- K8s cluster version
- Node Kernel version
- Browser version

**Additional context**

ddelnano:
Hi @ChenLingPeng, thanks for reporting this issue and opening up a PR to address it!

I was able to verify that the functions we probe are dependent on `CONFIG_SECURITY_NETWORK` (as you mentioned). I believe we need to add probes to the entry and return of `sock_sendmsg`, however, it was intentional that we are tracing `security_sock_sendmsg`.

In Google's [Container optimized OS](https://cloud.google.com/container-optimized-os/docs), we've noticed that it is possible for the kernel to call call `security_socket_sendmsg` rather than `socket_sendmsg` (https://github.com/pixie-io/pixie/commit/508686ff248f9210bc34bc4580dd371a7b3072a7). If we change which function we probe, we will miss the probe on COS in some cases.

I believe we will need to do something more dynamic such as inspecting the kernel config or available probes to determine the correct function to attach to. Let me discuss with the rest of the team to get consensus, but I'm fairly confident that's what the ideal change would do.

ddelnano:
@ChenLingPeng I've confirmed with the team that we will need a dynamic solution for this so that we ensure the COS case is still handled properly.

Can you give more detail on what kernel you are using? Is this custom built or are you using an upstream kernel?

[...]

PR #1811, opened 2024-01-05
feat: support fallback probes when a probe attach fails

CameronHall:
Summary: Support fallback probes when a probe attach fails

This should address the issue mentioned in https://github.com/pixie-io/pixie/issues/932. I ran into this while trying to run Stirling on WSL.

Type of change: /kind bug

Test Plan: TBD


src/stirling/bpf_tools/bcc_wrapper.cc
```diff
@@ -181,6 +181,11 @@ Status BCCWrapperImpl::AttachKProbe(const KProbeSpec& probe) {
       bpf_.attach_kprobe(GetKProbeTargetName(probe), std::string(probe.probe_fn), 0 /* offset */,
                          static_cast<bpf_probe_attach_type>(probe.attach_type), kKprobeMaxActive);
 
+  if (!status.ok() && probe.fallback_probe != nullptr) {
+    VLOG(1) << "kprobe attach failed... attempting fallback."
```
ddelnano, line 185: @CameronHall while we wait for approval to run the GitHub actions, there is a missing comma here that will fix the test build.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -332,10 +332,13 @@ const auto kProbeSpecs = MakeArray<bpf_tools::KProbeSpec>(
...
-      /*is_syscall*/ false},
-     {"security_socket_recvmsg", ProbeType::kEntry, "probe_entry_security_socket_recvmsg",
-      /*is_syscall*/ false}});
+     {"security_socket_sendmsg", ProbeType::kEntry, "probe_entry_socket_sendmsg",
+     /*is_syscall*/ false, /* is_optional */ false,
+     new bpf_tools::KProbeSpec{"sock_sendmesg", ProbeType::kEntry, "probe_entry_socket_sendmsg", false, true}},
+     {"security_socket_recvmsg", ProbeType::kEntry, "probe_entry_socket_recvmsg",
+     /*is_syscall*/ false, /* is_optional */ false,
+     new bpf_tools::KProbeSpec{"sock_recvmsg", ProbeType::kEntry, "probe_entry_socket_recvmsg", false, true}},
+  });
```
ddelnano, line 341: There are also some linting errors here that can be identified by running `arc lint` (assuming your development environment is setup).

ddelnano (commented):
Thanks for the contribution @CameronHall! A few minor comments, but this approach lgtm for handling this fallback behavior.

CameronHall:
Thanks @ddelnano. I've addressed the linting issues

src/stirling/bpf_tools/probe_specs/probe_specs.h
```diff
@@ -75,6 +75,9 @@ struct KProbeSpec {
   // kernels.
   bool is_optional = false;
 
+  // If the kernel function is not found, then this fallback function will be used instead.
+  struct KProbeSpec* fallback_probe = nullptr;
```
JamesMBartlett, line 79: This should be a managed ptr (`std::unique_ptr`) otherwise the memory will be leaked when the struct is deallocated.

JamesMBartlett (commented):
Sorry for the delay in reviewing. One minor change and then we should be able to get this through.

CameronHall:
@JamesMBartlett sorry it took so long. Updated it now.

src/stirling/bpf_tools/probe_specs/probe_specs.h
```diff
@@ -75,6 +75,9 @@ struct KProbeSpec {
   // kernels.
   bool is_optional = false;
 
+  // If the kernel function is not found, then this fallback function will be used instead.
+  struct KProbeSpec* fallback_probe = unique_ptr;
```
JamesMBartlett, line 79: Since this is still a pointer to a struct you will still leak memory. What I was imagining with my comment is:
```
std::unique_ptr<KProbeSpec> fallback_probe = nullptr;
```

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -293,49 +293,54 @@ void SocketTraceConnector::InitProtocolTransferSpecs() {
...
+    {"recvmsg", ProbeType::kReturn, "syscall__probe_ret_recvmsg"},
+    {"recvmmsg", ProbeType::kEntry, "syscall__probe_entry_recvmmsg"},
+    {"recvmmsg", ProbeType::kReturn, "syscall__probe_ret_recvmmsg"},
+    {"close", ProbeType::kEntry, "syscall__probe_entry_close"},
+    {"close", ProbeType::kReturn, "syscall__probe_ret_close"},
+    {"mmap", ProbeType::kEntry, "syscall__probe_entry_mmap"},
+    {"sock_alloc", ProbeType::kReturn, "probe_ret_sock_alloc", /*is_syscall*/ false},
+    {"security_socket_sendmsg", ProbeType::kEntry, "probe_entry_socket_sendmsg",
+     /*is_syscall*/ false, /* is_optional */ false,
+     new bpf_tools::KProbeSpec{"sock_sendmesg", ProbeType::kEntry, "probe_entry_socket_sendmsg",
```
JamesMBartlett, line 337: And then over here you can do `std::make_unique<bpf_tools::KProbeSpec>(...)`

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -293,49 +293,54 @@ void SocketTraceConnector::InitProtocolTransferSpecs() {
...
+                               false, true}},
+    {"security_socket_recvmsg", ProbeType::kEntry, "probe_entry_socket_recvmsg",
+     /*is_syscall*/ false, /* is_optional */ false,
+     new bpf_tools::KProbeSpec{"sock_recvmsg", ProbeType::kEntry, "probe_entry_socket_recvmsg",
```
JamesMBartlett, line 341: ditto

JamesMBartlett (commented):
Sorry for the delay on my end and the misunderstanding with my previous comment. Hopefully, I described what I was thinking a little more clearly this time.

ddelnano:
@CameronHall thanks again for getting this PR open. There was recent interest in this change from a community slack discussion, so I went ahead and added on top of your change to address the remaining feedback.

@JamesMBartlett `std::unique_ptr` runs into problems since [this](https://github.com/pixie-io/pixie/blob/79886a48cd2f8a1b77a57b2a314dfe708454f8ce/src/stirling/bpf_tools/bcc_wrapper.cc#L190) relies on a copy constructor being present. Instead of trying to working around that, I chose to use a `std::shared_ptr`. Let me know what you think.

JamesMBartlett:
> @JamesMBartlett `std::unique_ptr` runs into problems since [this](https://github.com/pixie-io/pixie/blob/79886a48cd2f8a1b77a57b2a314dfe708454f8ce/src/stirling/bpf_tools/bcc_wrapper.cc#L190) relies on a copy constructor being present. Instead of trying to working around that, I chose to use a `std::shared_ptr`. Let me know what you think.

I think we should reserve using shared pointers for when the memory is actually shared between two different owners (in this case its just allocated at init), so IMO we should just change the usage in BCCWrapper. Looking at it, it seems that the probe specs are only ever used as a const reference so we should just change the class to store const references instead of copying the probe spec.

ddelnano:
> it seems that the probe specs are only ever used as a const reference so we should just change the class to store const references instead of copying the probe spec.

@JamesMBartlett are you talking about changing the `kprobes_` member to that data type? I believe `std::vector` elements need to be assignable and after trying that I haven't been able to find an option that compiles. I feel like `std::vector` is probably not the right fit and a read only container would make it easier to accomplish what you described.

If that wasn't what you were thinking, let me know and I'm happy to try it.

JamesMBartlett (approved):
Discussed offline. Decided that its easier to just go with the shared ptr.

PR #1821, opened 2024-01-18
Fix bug that threw an error while searching for linux headers through a symlink

kpattaswamy:
Summary: This PR removes a condition which checks to see if a host path exists in the filesystem of a process when trying to resolve symlinks.

The function `fs::Exists()` returned false for a broken symlink when the expectation was to return true because the path existed although it was a broken symlink. This function would later on fix the broken symlink path by reading its target host and converting the broken path into the actual path. This makes the code removal safe as the next check: `is_symlink` will return an error code if there is no file.

Type of change: /kind bug

Test Plan: Existing tests pass and deployed the PEM with the updated fix and saw that the Linux header files were found.

### DIFF
--- a/src/stirling/utils/linux_headers.cc
+++ b/src/stirling/utils/linux_headers.cc
@@ -214,7 +214,8 @@ StatusOr<std::filesystem::path> ResolvePossibleSymlinkToHostPath(const std::file
   std::error_code ec;
   const bool is_symlink = std::filesystem::is_symlink(p, ec);
   if (ec) {
-    return error::Internal(ec.message());
+    return error::NotFound(absl::Substitute("Did not find the host headers at path: $0, $1.",
+                                            p.string(), ec.message()));
   }
 
   if (!is_symlink) {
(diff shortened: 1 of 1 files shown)


etep:
We should update the error message when `is_symlink` returns an error. Something like:
```
return error::NotFound(absl::Substitute("Did not find host headers at path: $0, $1.", p.string(), ec.message()));
```

Also, I think the following:
```
  if (!is_symlink) {
    // Not a symlink, we are good now.
    return p;
  }
```

Should return `system::Config::GetInstance().ToHostPath(p)`.

etep (commented):
Can we try a PR description along the lines of:

Method `fs::Exists` returned false for a broken symlink, but its expected behavior was to return true, i.e. because something existed (even if that something was a broken symlink). This function later fixes the broken symlink by reading its target and converting that target into a host path.

Removing the `fs::Exists` check solves the bug where the function early exits before resolving a symlink (see above explanation) and is safe because the next check, `is_symlink` will return an error code if there is no file.

etep:
Also, we can update the test plan, something like:

deployed pem with bug fix and observed correct behavior.

etep (approved):

etep:
My own comment:

The condition, removed in this PR, was fully redundant to the next check `is_symlink`. Not only was it redundant, but it was also wrong in the case of a broken symlink.

etep (approved):

oazizi000 (approved):

PR #1820, opened 2024-01-16
Address compilation issues with auto generated amqp files

ddelnano:
Summary: Address compilation issues with auto generated amqp files

When reviewing #1816, I realized that there was drift from the amqp template files and the resulting code. The updated stitcher interface was propagated to the [auto generated files](https://github.com/pixie-io/pixie/commit/e04a764d714fa4c9618312b567697b287b9ec3b1#diff-cece0214fed75080ef2a368fa37e9addb7c1c407c0463038c18c7c48af297ce7R210), but not the template itself. As for the other changes, I wasn't able to track down why they didn't exist in the template file but these changes were necessary to get the code compiling again.

Once this is merged, #1816 should be easy to merge afterwards.

Relevant Issues: #1816

Type of change: /kind cleanup

Test Plan: amqp trace bpf test passes

PR #1827, opened 2024-01-24
Automate amqp code generation and add test to catch drift

ddelnano:
Summary: Automate amqp code generation and add test to catch drift

This is a work in progress change to implement what was described [here](https://github.com/pixie-io/pixie/pull/1820#issuecomment-1905273586).

Relevant Issues: https://github.com/pixie-io/pixie/pull/1820#issuecomment-1905273586

Type of change: /kind cleanup

Test Plan: `diff_file` rules should fail until linting issues are resolved


vihangm:
I think it's okay to skip linting for autogenerated files. Might be worth adding them to the exclusion list and calling this done.

vihangm:
I think if you reduce the tab/spaces in the returns for `gen_method_enum_select_case`, `gen_method_enum_declr` etc to the expected output, the generated code is probably mostly formatted as expected.

There's a lot of them though, so unsure if it's worth the trouble.

ddelnano:
I was expecting the diff 'as is' to be pretty nasty, which is why I tried to keep the code formatting intact initially. I'll see how difficult it is to get the code generation to output relatively formatted code.

ddelnano:
@vihangm my latest approach gets the files formatted with minor whitespace differences. It required quite a few changes to the amqp code generation file. The changes to `amqp_code_gen.py` could be cleaner, but I wanted to get a feel for how many lines needed to be touched to make the diff reasonable. It required enough whitespace changes that my preference would be to stick with the initial approach that formatted the output files according to our clang-format config.

Let me know what you think.

vihangm:
Given that you have spent a bunch of time here, I trust your judgement. If you think running `clang-format` as part of the generation isn't too much trouble, that approach seems fine too. I was worried that running the linter would be more painful and hence suggested skipping lint/updating the code generation. :)

ddelnano:
Sounds good :+1:. I've updated the branch based on the initial approach and it's ready for review now.

PR #1833, opened 2024-02-02
Parse local address from socket info if connect/accept syscall not traced (and remote addr is known)

benkilimnik:
Summary: Add condition to `InferConnInfo` and parse the local address from socket information if its missing when the remote address is known.

Type of change: /kind feature

Test Plan: E2E tests. Skaffolded to a cluster and inspected pem logs. Ran and compiled standalone pem.


src/stirling/source_connectors/socket_tracer/conn_tracker.cc
```diff
@@ -797,12 +797,17 @@ void ConnTracker::IterationPreTick(
...
+  const bool laddr_found = open_info_.local_addr.family != SockAddrFamily::kUnspecified;
+  const bool info_mgr_ok = socket_info_mgr != nullptr;
 
+  if (!raddr_found && info_mgr_ok) {
+    InferConnInfo(proc_parser, socket_info_mgr);
     // TODO(oazizi): If connection resolves to SockAddr type "Other",
     //               we should mark the state in BPF to Other too, so BPF stops tracing.
     //               We should also mark the ConnTracker for death.
+  } else if (!laddr_found && info_mgr_ok) {
+    InferConnInfo(proc_parser, socket_info_mgr, true);
```
oazizi000, line 810: I like the idea of checking for both missing local and remote addresses, though I am curious about something: 

Can you comment on how this is specific to long-lived connections where we miss the `connect`/`accept`? (as mentioned in the description).

If we've missed the initial connection, we should be missing both local and remote address. This change seems to be specific to the case where we've somehow caught the remote address, but not the local address from BPF. I don't doubt that it can happen, but I'm just trying to understand the scenario that triggered this.

src/stirling/source_connectors/socket_tracer/conn_tracker.cc
```diff
@@ -962,7 +982,9 @@ void ConnTracker::InferConnInfo(system::ProcParser* proc_parser,
     bool success = conn_resolver_->Setup();
     if (!success) {
       conn_resolver_.reset();
-      conn_resolution_failed_ = true;
+      if (!parse_local_addr_only) {
+        conn_resolution_failed_ = true;
+      }
```
oazizi000, line 987: Why is this conditional on `!parse_local_addr_only`?

src/stirling/source_connectors/socket_tracer/conn_tracker.cc
```diff
@@ -1024,13 +1046,21 @@ void ConnTracker::InferConnInfo(system::ProcParser* proc_parser,
...
+  if (!parse_local_addr_only) {
+    Status s = ParseSocketInfoRemoteAddr(socket_info, &open_info_.remote_addr);
+    if (!s.ok()) {
+      // reset only if remote address is not found
+      conn_resolver_.reset();
+      conn_resolution_failed_ = true;
+      CONN_TRACE(2) << absl::Substitute("Remote address (type=$0) parsing failed. Message: $1",
+                                        socket_info.family, s.msg());
+      return;
+    }
```
oazizi000, line 1063: For a bit of additional symmetry, would it make sense to write this as:

```
if (open_info_.local_addr.empty()) { // empty() is pseudo-code
  Status s = ParseSocketInfoLocalAddr(socket_info, &open_info_.local_addr);
  ...
}

if (open_info_.remote_addr.empty()) { // empty() is pseudo-code
  Status s = ParseSocketInfoRemoteAddr(socket_info, &open_info_.local_addr);
  ...
}
```

Also, wondering again about why local and remote addresses are treated differently in terms of `conn_resolution_failed_` treatment. Have you seen cases where remote addresses can be inferred by not local ones?

oazizi000 (commented):
Looking forward to this, but trying to understand some things first. See comments in the code.

src/stirling/source_connectors/socket_tracer/conn_tracker.cc
benkilimnik: Thanks for the feedback @oazizi000. You're right that this is not specific to long-lived connections. I believe it is most likely to occur when we capture a `connect` syscall which lacks socket information for us trace the local IP from bpf. I've updated the description accordingly.

src/stirling/source_connectors/socket_tracer/conn_tracker.cc
benkilimnik: We hit this condition when the `FDResolver` fails to [read](https://github.com/pixie-io/pixie/blob/07672ff59b9441242a7adf8993a6493572b457ac/src/stirling/source_connectors/socket_tracer/fd_resolver.cc#L39) the `proc/pid/fd` link of the connection. If we set `conn_resolution_failed_` for cases where we already know the remote IP, then we'll end up [disabling](https://github.com/pixie-io/pixie/blob/6b418c2d4a3b69e8b37b76b6b4ff38bcde80fb8b/src/stirling/source_connectors/socket_tracer/conn_tracker.cc#L729) [tracing](https://github.com/pixie-io/pixie/blob/6b418c2d4a3b69e8b37b76b6b4ff38bcde80fb8b/src/stirling/source_connectors/socket_tracer/conn_tracker.cc#L754) in `ConnTracker::UpdateState` and skipping inference.

That said, I’m unsure why the FD might be volatile for connections where we know the remote addr already.

Before I added this conditional, a couple bpf tests started failing on [conn_resolver_->Setup()](https://github.com/pixie-io/pixie/blob/6b418c2d4a3b69e8b37b76b6b4ff38bcde80fb8b/src/stirling/source_connectors/socket_tracer/conn_tracker.cc#L982).
```sh
pgsql_trace_bpf_test:PostgreSQLTraceTest.SelectQuery
socket_trace_bpf_test:NonVecSyscalls/NonVecSyscallTests.NonVecSyscalls/1
```
See [buildbuddy](https://app.buildbuddy.io/invocation/2c6c36ad-ae6d-4d34-a7d7-d4fc07e36a2e?target=%2F%2Fsrc%2Fstirling%2Fsource_connectors%2Fsocket_tracer%3Apgsql_trace_bpf_test&targetStatus=6#@214) for additional context.

src/stirling/source_connectors/socket_tracer/conn_tracker.cc
benkilimnik: Thanks for the suggestion @oazizi000. I've just updated the conditions for simplicity. The main motivation behind treating local/remote addresses differently is to prevent the connection from being dropped when we already have a remote endpoint, but fail to parse a local address. See previous comment for additional details.

oazizi000 (approved):
Looks good. Thanks for the updates!

PR #1878, opened 2024-04-20
[Mongo] Clear the streamID from the request/response maps after stitching

kpattaswamy:
Summary: This PR modifies mongo's stitcher logic to clear streamID’s from request/response maps once all frames of a streamID have been consumed. We observed high CPU use allocated to `FramesSize()` & `EraseExpiredFrames()`, this was due to the size of the maps increasing with new streamIDs and having to continuously loop over those growing maps to cleanup. Clearing the streamID's from the maps after stitching significantly reduces the CPU allocated to the cleanup logic for mongo, the exact details for mongo's streamID reuse needs to be determined to further adapt the stitcher logic.

Type of change: /kind bug

Test Plan: Existing tests still pass, ran the px-mongo demo and observed lower CPU allocated to mongo in the PEM through flamegraph.

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
@@ -183,7 +183,8 @@ RecordsWithErrorCount<mongodb::Record> StitchFrames(
       ++erase_until_iter;
     }
 
-    // Determine whether to clear the StreamID from the map or only the frames that have been consumed.
+    // Determine whether to clear the StreamID from the map or only the frames that have been
+    // consumed.
     if (erase_until_iter == req_deque.end()) {
       reqs->erase(stream_id);
     } else {
--- a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
+++ b/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
@@ -183,7 +183,12 @@ RecordsWithErrorCount<mongodb::Record> StitchFrames(
       ++erase_until_iter;
     }
 
-    req_deque.erase(req_deque.begin(), erase_until_iter);
+    // Determine whether to clear the StreamID from the map or only the frames that have been consumed.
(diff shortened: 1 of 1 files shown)


src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher.cc
```diff
@@ -195,9 +201,11 @@ RecordsWithErrorCount<mongodb::Record> StitchFrames(
         error_count++;
       }
     }
-    resp_deque.clear();
   }
 
+  // Clear the response map.
+  resps->clear();
+
```
etep, line 208: In the previous version, where we had `resp_deque.clear();`, it seemed "ok" to iterate over the dequeues and add up the error count. Now it seems like potentially wasted iteration. In the main loop, starting on line 130 above, we iterate over the responses in the a dequeue. Can we make sure that the error count is counted up in that code above and just remove lines 196 to 200?

kpattaswamy: The code in line 130 loops over a `resp_deque` found from the response map using the `stream_id` from `state->stream_order`. This `stream_id` is inserted to the `stream_order` vector at the time of parsing a [request side frame](https://github.com/pixie-io/pixie/blob/48d20c7271dce3daeb6356d115229f37ec5dd27a/src/stirling/source_connectors/socket_tracer/protocols/mongodb/parse.cc#L78C1-L79C1). In an edge case where we parse a response frame but miss its corresponding request frame, the response frame's `stream_id` would not exist in `stream_order` meaning the `resp_deque` it's in would not be found/iterated across

This [test case](https://github.com/pixie-io/pixie/blob/e9be6fef18fbe37b9f480c965c42c5ef9a69dc6e/src/stirling/source_connectors/socket_tracer/protocols/mongodb/stitcher_test.cc#L201) is an example of that, I think we’d still need the final loop to calculate the number of unconsumed response frames to handle that edge case

etep (approved):

ddelnano (approved):

PR #1883, opened 2024-04-25
Make sure altstack_size_ is a multiple of the system page size

### LINKED ISSUE #1882: The PEM fails to start up on 7th gen AMI using bottlerocket
**Describe the bug**
The PEM is unable to run on a EKS cluster containing 7th gen instances running bottlerocket.

**To Reproduce**
Deploy pixie to a EKS cluster with a `m7i.large` instance type node and inspect the vizier-pem logs to see an error.

**Expected behavior**
Pixie should start up on EKS clusters running a 7th gen AMI using bottlerocket. 

**Logs**
```
Defaulted container "pem" out of: pem, qb-wait (init)
I20231103 21:12:02.837999 62316 env.cc:47] Started: /app/src/vizier/services/agent/pem/pem
F20231103 21:12:02.838059 62316 signal_action.cc:149] Check failed: mprotect(altstack_ + guard_size_ + altstack_size_, guard_size_, 0x0) == 0 (-1 vs. 0)
```

**App information (please complete the following information):**
- PEM version: 0.14.8
- K8s cluster version: 1.28
- Node Kernel version: 6.1.82

**Additional context**
It seems that this issue arises due to trying to `mprotect` a region of memory in the `altstack_` that isn't page aligned

kpattaswamy:
Summary: This PR modifies the logic that sets the value of `altstack_size_`. When deploying Pixie on a 7th gen AMI running bottle rocket, the value of `altstack_size_` was set to `MINSIGSTKSZ`. This value was 39616 and was not a multiple of the system page size.

In this [line of code](https://github.com/pixie-io/pixie/blob/0888c38df2dc414a36cc84198c7b159c39eea0e0/src/common/signal/signal_action.cc#L149) `mprotect` was then applied on a non page aligned region since its address was set as `altstack_ + guard_size_ + altstack_size_` leading to an error. This region in `altstack_` was also not the tailing `guard_size_` area as [mmap](https://github.com/pixie-io/pixie/blob/0888c38df2dc414a36cc84198c7b159c39eea0e0/src/common/signal/signal_action.cc#L145) would have added extra bytes to `altstack_` in order to page align it.

As per the docs, the address passed to `mprotect` needs to be aligned to a page boundary. The function in this PR makes sure that the value of `altstack_size_` is a multiple of the page size so that the address calculated/passed to `mprotect` is aligned to a page boundary and that it protects the tailing `guard_size_` region of `altstack_`

Relevant Issues: Fixes #1882

Type of change: /kind bug

Test Plan: Skaffolded pixie on EKS clusters with m7i.large, r7i.large instances and on a GKE cluster with e2-standard-4 nodes and saw the PEM start up

### DIFF
--- a/src/common/signal/signal_action.h
+++ b/src/common/signal/signal_action.h
@@ -35,7 +35,7 @@ class SignalAction : public NotCopyable {
  public:
   SignalAction()
       : guard_size_(px::system::Config::GetInstance().PageSizeBytes()),
-        altstack_size_(std::max(guard_size_ * 4, static_cast<size_t>(MINSIGSTKSZ))) {
+        altstack_size_(DetermineAltStackSize()) {
     MapAndProtectStackMemory();
     InstallSigHandlers();
   }
@@ -71,6 +71,23 @@ class SignalAction : public NotCopyable {
    * Additionally, two guard pages will be allocated to bookend the usable area.
    */
   const size_t altstack_size_;
+  /**
+   * Determine the number of bytes to allocate to altstack_size_.
+   * Makes sure that it is a multiple of the system page size.
+   */
+  size_t DetermineAltStackSize() const {
+    auto min_stack_size = static_cast<size_t>(MINSIGSTKSZ);
+
+    if (guard_size_ * 4 > min_stack_size) {
+      return guard_size_ * 4;
+    }
+
+    if (min_stack_size % guard_size_ != 0) {
+      min_stack_size += guard_size_ - (min_stack_size % guard_size_);
+    }
+
+    return min_stack_size;
+  }
   /**
    * Signal handlers will be installed for these signals which have a fatal outcome.
    */
--- a/src/common/signal/signal_action.h
+++ b/src/common/signal/signal_action.h
@@ -73,15 +73,13 @@ class SignalAction : public NotCopyable {
   const size_t altstack_size_;
   /**
    * Determine the number of bytes to allocate to altstack_size_.
-   * Makes sure that it is a multiple of the system page size.
    */
   size_t DetermineAltStackSize() const {
-    // The size of altstack_ should be at least 4 * page size or size MINSIGSTKSZ if it is
+    // The size of altstack_size_ should be at least 4 * guard size or size MINSIGSTKSZ if it is
     // greater. This size needs to be a multiple of the system page size.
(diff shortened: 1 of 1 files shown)


src/common/signal/signal_action.h
```diff
@@ -71,6 +71,23 @@ class SignalAction : public NotCopyable {
...
+    if (guard_size_ * 4 > min_stack_size) {
+      return guard_size_ * 4;
+    }
+
+    if (min_stack_size % guard_size_ != 0) {
+      min_stack_size += guard_size_ - (min_stack_size % guard_size_);
+    }
+
+    return min_stack_size;
+  }
```
etep, line 90: Nice work in tracking this down!

If we use page size instead of guard size, then it becomes clear that we are aligning to the page size.

Perhaps something along the lines of:
```
const size_t min_alt_stack_size = 4 * page_size;
const size_t stack_size_page_count = IntRoundUpDivide(min_stack_size, page_size);
const size_t stack_size_bytes = page_size * stack_size_page_count;
return std::max(stack_size_bytes, min_alt_stack_size);
```

I'm not completely in sync with the names of the variables above. If possible, can we find better names that capture the semantics of the values we calculate (e.g. there is min stack size and also 4x of that ... do we have a semantic distinction in these concepts?).
```

Comment (kpattaswamy) [reply src/common/signal/signal_action.h]: My understanding from the original imp was that `altstack_` should always be at least 4*page size, but if `MINSIGSTKSZ` is greater we default to that, so I kept the same condition to choose between the two in the new logic

File: src/common/signal/signal_action.h [hunk 2]
```diff
@@ -71,6 +71,20 @@ class SignalAction : public NotCopyable {
...
    */
   const size_t altstack_size_;
+  /**
+   * Determine the number of bytes to allocate to altstack_size_.
+   * Makes sure that it is a multiple of the system page size.
+   */
+  size_t DetermineAltStackSize() const {
+    // The size of altstack_ should be at least 4 * page size or size MINSIGSTKSZ if it is
+    // greater. This size needs to be a multiple of the system page size.
+    const size_t page_size = px::system::Config::GetInstance().PageSizeBytes();
```
ddelnano, line 81: `guard_size_` is set to the page size. Can we reuse that rather than creating another variable for referencing page size?

etep (approved):

PR #1884, opened 2024-04-26
Optimize TCP Stats Connector

benkilimnik:
Summary: Makes a couple of optimizations to the TCP stats connector. In particular, we pass by reference instead of copy in `TCPStats::UpdateStats` and simplify the cleanup logic in `TransferDataImpl`.

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Bpf test works. Ran heap profiler and observed fewer allocations.

One remaining mystery is a single 384Mb allocation that seems to always arise in the heap profile for `AcceptTcpEvent`. Even commenting out the body of the function does not change this. Originally thought this might be related to the `BPF_HASH(sock_store, uint32_t, struct sock*, 10240);`, but changing its size had no affect on the allocation.

image


etep (approved):
Easy button. Nice & thank you!

RagalahariP (approved):
Looks good. Thank you.

ddelnano (approved):

PR #1885, opened 2024-04-27
Upgrade dev and linter containers to Ubuntu 24.04

ddelnano:
Summary: Upgrade dev and linter containers to Ubuntu 24.04

Relevant Issues: N/A

Test Plan: CI should pass when using the new images
- [x] BPF tests on all kernels pass -- `#ci:bpf-build-all-kernels` build ([link](https://github.com/pixie-io/pixie/actions/runs/9045655568/job/24855673228?pr=1885))
- [x] All non BPF tests pass -- `#ci:ignore-deps` build ([link](https://github.com/pixie-io/pixie/actions/runs/9045355544/job/24855061659))
- [x] Builds and test runs on VM provisioned with chef are successful

Changelog Message: Upgrade development environment to Ubuntu 24.04


tools/chef/cookbooks/px_dev_extras/attributes/linux_debs.rb
```diff
@@ -24,6 +24,15 @@
 default['clang']['deb_sha256'] =
   'a7c6aa046cb3a75fae2f61e1ed43abbcce6514dd91f933bdc20f6633113994ed'
```
vihangm, line 25: just realized that the location for clang needs to be updated to point to https://github.com/pixie-io/dev-artifacts/releases/tag/clang%2F15.0-pl12

ddelnano: The clang-linters deb is in the same state. I'll update that one as well.

ddelnano: This is fixed in the latest revision.

I also was able to upload the libtinfo5 release, so the `tools/docker/Makefile` changes are included along with the corresponding chef install from the dev-artifacts URL.

tools/chef/cookbooks/px_dev/recipes/linters.rb
```diff
@@ -26,7 +26,7 @@
 end
 
 execute 'install py linters' do
-  command 'python3 -m pip install flake8 flake8-mypy yamllint --no-cache-dir && python3 -m pip cache purge'
+  command 'python3 -m pip install --break-system-packages flake8 flake8-mypy yamllint --no-cache-dir && python3 -m pip cache purge'
```
vihangm, line 29: What system package does this install conflict with? can we side step this by doing a user install?
This also seems ok, but the flag naming usually alarms folks. :)

vihangm (approved):

tools/chef/cookbooks/px_dev/recipes/linters.rb
ddelnano: I don't think these packages actually conflict with anything currently. I believe Ubuntu is marking its python as externally managed according to [PEP 668](https://peps.python.org/pep-0668/) and so any system wide pip install is going to complain with Ubuntu's python.

My understanding is that we would need to install the ubuntu apt versions of these dependencies:
* [yamllint](https://packages.ubuntu.com/en/noble/yamllint)
* [flake8](https://packages.ubuntu.com/en/noble/flake8)
* flake8-mypy [upstream](https://github.com/ambv/flake8-mypy) is deprecated so that should probably be replaced

What if we replaced flake8 and yamllint with Ubuntu's packages and deferred "fixing" flake8-mypy until it is replaced with a different linter integration?

PR #1886, opened 2024-04-30
Ensure a hermetic python install is used for `pip_parse` and upgrade `rules_python`

ddelnano:
Summary: Ensure a hermetic python install is used for `pip_parse` and upgrade `rules_python`

This is prerequisite work for migrating our docker dev image  to Ubuntu 24.04 (#1885). In that PR, I've been running builds from within the containers built via packer to verify the upgrade is successful. This change is required to address the following issues:
* The `pip_parse` for `ubuntu_package_deps` comes before the hermetic python definition. This results in a `ModuleNotFoundError` error without the `rules_python` upgrade (since python 3.12, shipped with 24.04, [removed disutils](https://peps.python.org/pep-0632/)).
* The remaining `pip_parse` calls fail when run within a 24.04 container since it causes python to upgrade to 3.12.

Note: `rules_python` 0.26.0 is the latest version we can easily upgrade to. The next release has issues with our current version of `rules_docker`. Since `rules_docker` is deprecated and we are at the latest version, upgrading beyond 0.26.0 seemed like it would be a more significant change.

Relevant Issues: N/A

Test Plan: CI build should pass and running builds in containers from #1885 succeed

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -142,6 +142,9 @@ pl_model_files()
 
 python_register_toolchains(
     name = "python3_10",
+    # Allow the root user to build the code base since this is a current requirement for
+    # building in a containerized environment. See https://github.com/bazelbuild/rules_python/pull/713
+    # for more details.
     ignore_root_user_error = True,
     # Available versions are listed in @rules_python//python:versions.bzl.
     # We recommend using the same version your team is already standardized on.
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -80,6 +80,10 @@ load("//bazel:pl_workspace.bzl", "pl_container_images", "pl_model_files", "pl_wo
 
 pl_workspace_setup()
 
+load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
+
+py_repositories()
+
 # The pip_deps rule cannot be loaded until we load all the basic packages in the Pixie
 # workspace. Also, bazel requires that loads are done at the top level (not in a function), so
 # we need to pull it out over here.
@@ -136,10 +140,9 @@ tf_workspace0()
 
 pl_model_files()
 
-load("@rules_python//python:repositories.bzl", "python_register_toolchains")
-
 python_register_toolchains(
     name = "python3_10",
+    ignore_root_user_error = True,
     # Available versions are listed in @rules_python//python:versions.bzl.
     # We recommend using the same version your team is already standardized on.
     python_version = "3.10",
@@ -160,6 +163,7 @@ vizier_api_install_deps()
 
 pip_parse(
     name = "pxapi_python_doc_deps",
+    python_interpreter_target = interpreter,
     requirements_lock = "//src/api/python/doc:requirements.bazel.txt",
 )
 
@@ -250,6 +254,7 @@ go_download_sdk(
 
 pip_parse(
     name = "amqp_gen_reqs",
+    python_interpreter_target = interpreter,
     requirements_lock = "//src/stirling/source_connectors/socket_tracer/protocols/amqp/amqp_code_generator:requirements.bazel.txt",
 )
 
@@ -259,6 +264,7 @@ amp_gen_install_deps()
 
 pip_parse(
     name = "protocol_inference",
+    python_interpreter_target = interpreter,
     requirements_lock = "//src/stirling/protocol_inference:requirements.bazel.txt",
 )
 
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -140,6 +144,7 @@ load("@rules_python//python:repositories.bzl", "python_register_toolchains")
 
 python_register_toolchains(
     name = "python3_10",
+    ignore_root_user_error = True,
```
ddelnano, line 147: This was required to avoid the following error:

```
root@px-dev-docker-nas-metal-ubuntu22-l1-kvm:/px/src/px.dev/pixie# bazel build //...
DEBUG: /root/.cache/bazel/_bazel_root/54060b0ed2e63c063d495ae4fb1a7d19/external/rules_python/python/pip.bzl:49:10: pip_install is deprecated. Please switch to pip_parse. pip_install will be removed in a future release.
INFO: Repository python3_10_x86_64-unknown-linux-gnu instantiated at:
  /px/src/px.dev/pixie/WORKSPACE:145:27: in <toplevel>
  /root/.cache/bazel/_bazel_root/54060b0ed2e63c063d495ae4fb1a7d19/external/rules_python/python/repositories.bzl:580:26: in python_register_toolchains
Repository rule python_repository defined at:
  /root/.cache/bazel/_bazel_root/54060b0ed2e63c063d495ae4fb1a7d19/external/rules_python/python/repositories.bzl:404:36: in <toplevel>
INFO: repository @python3_10_x86_64-unknown-linux-gnu' used the following cache hits instead of downloading the corresponding file.
 * Hash '5d0429c67c992da19ba3eb58b3acd0b35ec5e915b8cae9a4aa8ca565c423847a' for https://github.com/indygreg/python-build-standalone/releases/download/20231002/cpython-3.10.13+20231002-x86_64-unknown-linux-gnu-install_only.tar.gz
If the definition of 'repository @python3_10_x86_64-unknown-linux-gnu' was updated, verify that the hashes were also updated.
ERROR: An error occurred during the fetch of repository 'python3_10_x86_64-unknown-linux-gnu':
   Traceback (most recent call last):
        File "/root/.cache/bazel/_bazel_root/54060b0ed2e63c063d495ae4fb1a7d19/external/rules_python/python/repositories.bzl", line 218, column 25, in _python_repository_impl
                fail("The current user is root, please run as non-root when using the hermetic Python interpreter. See https://github.com/bazelbuild/rules_python/pull/713.")
Error in fail: The current user is root, please run as non-root when using the hermetic Python interpreter. See https://github.com/bazelbuild/rules_python/pull/713.
ERROR: /px/src/px.dev/pixie/WORKSPACE:145:27: fetching python_repository rule //external:python3_10_x86_64-unknown-linux-gnu: Traceback (most recent call last):
        File "/root/.cache/bazel/_bazel_root/54060b0ed2e63c063d495ae4fb1a7d19/external/rules_python/python/repositories.bzl", line 218, column 25, in _python_repository_impl
                fail("The current user is root, please run as non-root when using the hermetic Python interpreter. See https://github.com/bazelbuild/rules_python/pull/713.")
Error in fail: The current user is root, please run as non-root when using the hermetic Python interpreter. See https://github.com/bazelbuild/rules_python/pull/713.
ERROR: Error computing the main repository mapping: Encountered error while reading extension file 'requirements.bzl': no such package '@vizier_api_python_deps//': no such package '@python3_10_x86_64-unknown-linux-gnu//': The current user is root, please run as non-root when using the hermetic Python interpreter. See https://github.com/bazelbuild/rules_python/pull/713.
Loading:
```

In order to avoid this, I believe we have two options:
* Use user namespaces to remap the container root user to a non root uid. This would mean our CI and adhoc use of the dev container would need to supply that at runtime.
* Update our packer built images to include an additional user and ensure that all the adhoc and CI use cases work under this new user.

vihangm: Maybe leave a note in the code saying that this is to allow the root user to build the code base which often happens when building in a containerized environment.

WORKSPACE
```diff
@@ -160,6 +163,7 @@ vizier_api_install_deps()
 
 pip_parse(
     name = "pxapi_python_doc_deps",
+    python_interpreter_target = interpreter,
```
vihangm, line 166: What's going on here?

ddelnano: It ensures that the `pip_parse` uses the toolchain's pip rather than the system pip. Without this upgrading the dev container to 24.04 opts us into a newer pip and python version and causes a ton of build failures (lots of wheels fail to build for the versions pinned in the `requirements.bazel.txt` files).

We already did this for [vizier_api_python_deps](https://github.com/pixie-io/pixie/blob/0888c38df2dc414a36cc84198c7b159c39eea0e0/WORKSPACE#L153), it just wasn't maintained for the new `pip_parse` rules added since the 22.04 upgrade.

WORKSPACE
ddelnano: Updated. Let me know if that doesn't address your feedback.

vihangm (approved):

PR #1894, opened 2024-05-06
Fix negative table size calculation when socket tracer disabled via stirling flag

benkilimnik:
Summary: When some or all of the tables `http_events, stirling_error, probe_status, proc_exit_events` are disabled via stirling flags (`PL_STIRLING_SOURCES=...`) the `other_table_size` calculation returns a negative output if `num_tables < 4`. It also attempts division by zero when `num_tables = 4`

```cpp
  int64_t other_table_size = (memory_limit - http_table_size - stirling_error_table_size -
                              probe_status_table_size - proc_exit_events_table_size) /
                             (num_tables - 4);
```

This PR streamlines the table size calculation and fixes the edge condition.

Type of change: /kind bug

Test Plan: deployed to node and manually checked calculated table sizes.

### DIFF
--- a/src/experimental/standalone_pem/standalone_pem_manager.cc
+++ b/src/experimental/standalone_pem/standalone_pem_manager.cc
@@ -169,22 +169,31 @@ Status StandalonePEMManager::InitSchemas() {
   bool has_http_events = false, has_stirling_error = false, has_probe_status = false,
        has_proc_exit_events = false;
   for (const auto& relation_info : relation_info_vec) {
-    if (relation_info.name == "http_events")
+    if (relation_info.name == "http_events") {
       has_http_events = true;
-    else if (relation_info.name == "stirling_error")
+    } else if (relation_info.name == "stirling_error") {
       has_stirling_error = true;
-    else if (relation_info.name == "probe_status")
+    } else if (relation_info.name == "probe_status") {
       has_probe_status = true;
-    else if (relation_info.name == "proc_exit_events")
+    } else if (relation_info.name == "proc_exit_events") {
       has_proc_exit_events = true;
+    }
   }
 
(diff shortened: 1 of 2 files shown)


src/experimental/standalone_pem/standalone_pem_manager.cc
```diff
@@ -164,32 +164,53 @@ Status StandalonePEMManager::InitSchemas() {
...
   for (const auto& relation_info : relation_info_vec) {
     std::shared_ptr<table_store::Table> table_ptr;
+    int64_t table_size = other_table_size;  // Default size
+
     if (relation_info.name == "http_events") {
       // Special case to set the max size of the http_events table differently from the other
       // tables. For now, the min cold batch size is set to 256kB to be consistent with previous
       // behaviour.
-      table_ptr = std::make_shared<table_store::Table>(relation_info.name, relation_info.relation,
-                                                       http_table_size, 256 * 1024);
```
ddelnano, line 178: The http compacted batch size of `256 * 1024` is no longer being applied. Was that intentional?

I think we should maintain the same behavior unless there is a strong reason to change it.

benkilimnik: Thanks for catching that! I agree, we should keep the http batch size. Fixed in latest commit.

ddelnano (approved):

benkilimnik:
Just pushed an update to also fix table size calculations in `pem_manager.cc` and add an error condition for when the pem flags are set such that there is not enough space for the default tables.
```cpp
  if (remaining_memory < 0) {
    return error::Internal("Table store data limit is too low to store the tables.");
  }
```

vihangm (approved):

PR #1935, opened 2024-05-30
chore: Upgrade go-git

vihangm:
Summary: This is to address a CVE.

Relevant Issues: CVE-2023-49569

Type of change: /kind cve

Test Plan: Ran the versions gen script, ensured it still works.


bazel/external/circl.patch
```diff
@@ -0,0 +1,70 @@
...
+     name = "fp448",
+     srcs = [
+@@ -11,6 +13,7 @@ go_library(
+         "fp_generic.go",
+         "fp_noasm.go",
+     ],
++    cgo = True,
+     importpath = "github.com/cloudflare/circl/math/fp448",
+     visibility = ["//visibility:public"],
+     deps = [
```
ddelnano, line 70: What's the purpose of this patch? Does this work around the fact that our go toolchain is old?

vihangm: Not really, newer bazel tooling will make the patching easier, but the main issue is that `circl` has a odd cgo setup that gazelle doesn't handle automatic build file generation for. So we have to patch what gazelle generates.

See https://github.com/bazelbuild/bazel-gazelle/issues/1543 and https://github.com/google/go-github/issues/2932 for other folks who ran into this.

PR #1941, opened 2024-06-06
Add x86_64 sysroot profile to cloud and operator skaffold configs

ddelnano:
Summary: Add x86_64 sysroot profile to cloud and operator skaffold configs

Deploying builds from a noble (Ubuntu 24.04) environment requires using the x86_64 sysroot at all times. Add this to the config to make this addition easy to opt into.

Relevant Issues: N/A

Type of change: /kind compatibility

Test Plan: Deployed a new cloud and got it successfully running with `skaffold run -p ${env},x86_64_sysroot -f skaffold/skaffold_cloud.yaml`

### DIFF
--- a/skaffold/skaffold_cloud.yaml
+++ b/skaffold/skaffold_cloud.yaml
@@ -153,3 +153,9 @@ profiles:
     path: /manifests/kustomize/paths
     value:
     - k8s/cloud/public/base
+- name: x86_64_sysroot
+  patches:
+  - op: add
+    path: /build/artifacts/context=./bazel/args
+    value:
+    - --config=x86_64_sysroot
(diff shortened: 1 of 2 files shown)


skaffold/skaffold_cloud.yaml
```diff
@@ -153,3 +153,9 @@ profiles:
     path: /manifests/kustomize/paths
     value:
     - k8s/cloud/public/base
+- name: x86_64_sysroot
+  patches:
+  - op: add
+    path: /build/artifacts/context=./bazel/args
+    value:
+    - --config=x86_64_sysroot
```
vihangm, line 161: I this this doesn't stack so if you do prod and sysroot, you'll lose opt and stamping etc.
You might want to add this to the prod profile

ddelnano: Good call. I glossed over that. I've updated it for testing, staging and prod and used a yaml anchor to reduce the duplication.

ddelnano:
@vihangm can you give this another look? Your previous comment should be addressed with my most recent change.

vihangm (approved):

PR #1953, opened 2024-06-24
Replace deprecated react-cookie-banner with react-cookie-consent

ddelnano:
Summary: Replace deprecated react-cookie-banner with react-cookie-consent

The react-cookie-banner library was deprecated in March 2024 (see [Github page](https://github.com/buildo/react-cookie-banner)). This replaces it with a similarly minimal library that is maintained.

Relevant Issues: N/A

Test Plan: Skaffold'ed a cloud and verified the new cookie banner works properly

### DIFF
--- a/src/ui/package.json
+++ b/src/ui/package.json
@@ -116,7 +116,7 @@
     "prop-types": "^15.5.4",
     "query-string": "^6.14.0",
     "react": "^18.1.0",
-    "react-cookie-banner": "^4.1.0",
+    "react-cookie-consent": "^9.0.0",
     "react-dom": "^18.1.0",
     "react-draggable": "^4.4.5",
     "react-grid-layout": "^0.18.2",
--- a/src/ui/package.json
+++ b/src/ui/package.json
@@ -116,7 +116,6 @@
     "prop-types": "^15.5.4",
     "query-string": "^6.14.0",
     "react": "^18.1.0",
-    "react-cookie-consent": "^9.0.0",
     "react-dom": "^18.1.0",
     "react-draggable": "^4.4.5",
     "react-grid-layout": "^0.18.2",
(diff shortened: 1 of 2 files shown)


src/ui/package.json
```diff
@@ -116,7 +116,7 @@
     "prop-types": "^15.5.4",
     "query-string": "^6.14.0",
     "react": "^18.1.0",
-    "react-cookie-banner": "^4.1.0",
+    "react-cookie-consent": "^9.0.0",
```
vihangm, line 119: minor nit: I wonder if we should move this to the `optional` section of deps in `package.json`

vihangm (approved):

src/ui/package.json
ddelnano: Being optional sounds correct for this type of dependency, however, there doesn't seem to be a way to enforce that yarn/npm installs an optional dependency. For the case when the banner should be included, the build should fail if the optional dependency is missing.

Do you have any ideas on how that could be accomplished? From my brief research, it seems like this might require more thought in order to avoid the issue where the cookie package is silently missing when it's needed.

vihangm: yarn/npm will always install optional deps unless you tell them to omit them.
also the build will fail if the dep is missing since the js bundler should try to resolve all `requires`.

you can quickly test that by doing the following in the UI dir: `rm -r node_modules` `yarn install --ignore-optional` and then trying to `yarn build_prod`. the last step should fail if you have an override that uses the cookie banner.

I'm also okay punting on this change for now.

ddelnano: ah, right. I was thinking it would be a runtime error, but webpack should have a problem with it.

Sounds good to me. I've updated the PR to reflect the optional dep.

vihangm (approved):

PR #1957, opened 2024-06-27
Refactor base colors to configurables directory. Allow customizing primary color for light and dark mode

ddelnano:
Summary: Refactor base colors to configurables directory. Allow customizing primary color for light and dark mode

This change makes it possible to have different main colors for light and dark mode. The contrast difference between the light mode background and the text color scores poorly with the [contrast checker](https://webaim.org/resources/contrastchecker/?fcolor=12D6D6&bcolor=F6F6F6).

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Tested UI to verify that it looks the same in light and dark mode
- [x] Verified all references to `COLORS.PRIMARY[500]` have been updated with new structure

### DIFF
--- a/src/ui/src/configurables/base/theme-colors.ts
+++ b/src/ui/src/configurables/base/theme-colors.ts
@@ -23,10 +23,8 @@ export const COLORS = {
     '200': '#A1F7F7',
     '300': '#6DF3F3',
     '400': '#35EEEE',
-    '500': { // Brand primary color!
-      dark: '#12D6D6',
-      light: '#12D6D6',
-    },
+    '500': '#12D6D6', // Brand primary color!
+    '550': '#12D6D6', // Primary for light mode
     '600': '#0DA0A0',
     '700': '#096C6C',
     '800': '#053838',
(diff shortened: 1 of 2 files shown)


src/ui/src/configurables/base/theme-colors.ts
```diff
@@ -0,0 +1,86 @@
...
+export const COLORS = {
+  PRIMARY: {
+    '100': '#D0FBFB',
+    '200': '#A1F7F7',
+    '300': '#6DF3F3',
+    '400': '#35EEEE',
+    '500': { // Brand primary color!
+      dark: '#12D6D6',
+      light: '#12D6D6',
+    },
```
NickLanam, line 29: **Nit (non-blocking)**: Instead of breaking convention here, how about a '500' and a '550' (then you can follow the normal pattern up in `mui-theme.tsx` as well)?

NickLanam (approved):

src/ui/src/configurables/base/theme-colors.ts
ddelnano: MUI is new to me and I wasn't sure how to best structure this. Updated to reflect that :+1:

NickLanam (approved):

PR #1961, opened 2024-07-03
Move all terms and privacy links to tos-privacy configurables module

ddelnano:
Summary: Move all terms and privacy links to tos-privacy configurables module

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Verified that the links on `/auth/signup` are correct with ui local dev workflow

### DIFF
--- a/src/ui/src/configurables/base/tos-privacy.tsx
+++ b/src/ui/src/configurables/base/tos-privacy.tsx
@@ -17,15 +17,15 @@
  */
 
 import * as React from 'react';
-export const termsUri = 'https://pixielabs.ai/terms';
-export const privacyUri = 'https://pixielabs.ai/privacy';
+export const termsUri = 'https://www.linuxfoundation.org/terms';
+export const privacyUri = 'https://www.linuxfoundation.org/privacy';
 
 export const TermsAndPrivacy = React.memo<{ classes: Record<'text', string> }>(({ classes }) => {
   return (<>
-    <a href='https://www.linuxfoundation.org/terms' className={classes.text}>
+    <a href={termsUri} className={classes.text}>
       Terms & Conditions
     </a>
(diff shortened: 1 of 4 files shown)


src/ui/src/configurables/base/tos-privacy.tsx
```diff
@@ -17,6 +17,8 @@
  */
 
 import * as React from 'react';
+export const termsUri = 'https://pixielabs.ai/terms';
+export const privacyUri = 'https://pixielabs.ai/privacy';
```
ddelnano, line 21: Is it intentional that these are different from the ones in the footer (defined [below](https://github.com/pixie-io/pixie/blob/e0aa3e2d820927ac8834d36c134643c7ad748f59/src/ui/src/configurables/base/tos-privacy.tsx#L23-L26))? Is it because signing up for CC4P is outside of CNCF/LF?

NickLanam: That's a question for @zasgar, I think. Note that we use the same Linux Foundation footers on `px.dev`.

Also, the only reason these first two are still pointing to `pixielabs.ai` is that we never bothered to put them up on `px.dev`, otherwise I'd say the base config should only be pointing to `px.dev`.

src/ui/src/pages/auth/signup.tsx
```diff
@@ -64,8 +65,7 @@ export const SignupPage = React.memo(() => {
             toggleURL={`/auth/login${window.location.search}`}
             title='Get Started'
             // Need to encapsulate so that newline is properly escaped.
```
NickLanam, line 67: **Nit**: This comment should move to where the string is defined, rather than where it's used.

ddelnano: Good call 👍

PR #1962, opened 2024-07-03
Refactor vendor specific signup message to configurables

ddelnano:
Summary: Refactor vendor specific signup message to configurables

Relevant Issues: N/A

Type of change: /kind clenaup

Test Plan:  Verified `/auth/signup`'s text is displayed as it was before

### DIFF
--- a/src/ui/src/configurables/base/signup.ts
+++ b/src/ui/src/configurables/base/signup.ts
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2018- The Pixie Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const signupMessage = `Pixie Community is Free Forever.
+No Credit Card Needed.`;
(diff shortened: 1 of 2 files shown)


src/ui/src/configurables/base/signup.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const signupMessage = `Pixie Community is Free Forever.
```
NickLanam, line 19: **Nit**: This should be a single-quoted string, not a backtick interpolated one, as there are no variables in it. I thought we had a lint rule enforcing this, but I might have turned it off in `configurables` a long time ago and forgotten about it.

NickLanam (approved):

src/ui/src/configurables/base/signup.ts
ddelnano: I believe the use of backticks was intentional because it's meant to be a multi line string. From my research, it seems that a template literal is the only way to do a muli line string.

If I make the string a single line, the linter does complain about it.

```
$ git diff
diff --git a/src/ui/src/configurables/base/signup.ts b/src/ui/src/configurables/base/signup.ts
index 0c3790746..0e8920738 100644
--- a/src/ui/src/configurables/base/signup.ts
+++ b/src/ui/src/configurables/base/signup.ts
@@ -16,5 +16,4 @@
  * SPDX-License-Identifier: Apache-2.0
  */

-export const signupMessage = `Pixie Community is Free Forever.
-No Credit Card Needed.`;
+export const signupMessage = `Pixie Community is Free Forever.`;

$ yarn lint
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating

/home/ddelnano/code/pixie/src/ui/src/configurables/base/signup.ts
  19:30  error  Strings must use singlequote  @typescript-eslint/quotes

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

ddelnano: Going to resolve since this is expected behavior from the linter given the new line.

screen02

PR #1965, opened 2024-07-12
Add gke PodMonitoring compatible labels for prometheus scrape-able services

ddelnano:
Summary: Add gke PodMonitoring compatible labels for prometheus scrape-able services

GKE's managed prometheus works by creating [PodMonitoring](https://github.com/GoogleCloudPlatform/prometheus-engine/blob/v0.10.0/doc/api.md#podmonitoring) resources that identify what targets to scrape. This resource only supports matching by labels and does not work with the existing annotations present in the Pixie cloud manifests. This PR adds PodMonitoring compatible labels that match the existing annotations

Relevant Issues: N/A

Type of change: /kind compatibility

Test Plan: Skaffold'ed to a cluster and verified the following:
- [x] Annotations and labels are set on the modified Deployments
- [x] GKE PodMonitoring resource correctly identifies backends now

### DIFF
--- a/k8s/cloud/base/indexer_deployment.yaml
+++ b/k8s/cloud/base/indexer_deployment.yaml
@@ -14,6 +14,9 @@ spec:
       labels:
         name: indexer-server
         monitoring.gke.io/scrape: 'true'
+        prometheus.io/scrape: 'true'
+        prometheus.io/port: '51801'
+        prometheus.io/scheme: 'http'
       annotations:
         prometheus.io/scrape: 'true'
         prometheus.io/port: '51801'
--- a/k8s/cloud/base/metrics_deployment.yaml
+++ b/k8s/cloud/base/metrics_deployment.yaml
@@ -12,6 +12,9 @@ spec:
       labels:
         name: metrics-server
         monitoring.gke.io/scrape: 'true'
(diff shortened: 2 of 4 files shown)


k8s/cloud/base/indexer_deployment.yaml
```diff
@@ -14,6 +14,9 @@ spec:
       labels:
         name: indexer-server
         monitoring.gke.io/scrape: 'true'
+        prometheus.io/scrape: 'true'
+        prometheus.io/port: '51801'
+        prometheus.io/scheme: 'http'
```
ddelnano, line 19: I tried to use a yaml anchor to avoid duplicating these, but skaffold wasn't compatible with it:

vihangm:
The podMonitoring should only need labels to identify workloads to scrape. It then uses prometheus to scrape those targets which AFAIK still relies on the annotations.
If you use `monitoring.gke.io/scrape: true` as your labelSelector, I believe GKE PodMonitoring should work with the existing annotations for port discovery etc.

ddelnano:
ah, I see. I can confirm that using the `monitoring.gke.io/scrape: true` label selector with the existing annotation has the same effect.

Issue #1646, opened 2023-08-03
Go binary version detection results in `std::bad_alloc` exception


ddelnano:
We recently revamped how Pixie detects a binary's Go version (#1605). This aligned it with how the `go version` command implements its functionality as well as addressed other bugs in our previous implementation (as described in the linked PR) .

We received a report of a `std::bad_alloc` exception that occurs within this new logic, which causes the PEM to crash.
```
libc++abi: terminating due to uncaught exception of type std::bad_alloc: std::bad_alloc
E20230730 23:53:00.815058 10815 signal_action.cc:63] Caught Aborted, suspect faulting address 0x262c. Trace:
**************************
PC: @     0x7f019b04c472  (unknown)  abort
    @     0x562574f7f976  (unknown)  abort_message
    @     0x562574f6981b  (unknown)  demangling_terminate_handler()
    @     0x562574f7f5f3  (unknown)  std::__terminate()
    @     0x562574f81316  (unknown)  __cxxabiv1::failed_throw()
    @     0x562574f812af  (unknown)  __cxa_throw
    @     0x562574f16384  (unknown)  (anonymous namespace)::handle_oom()
    @     0x562574f868dd  (unknown)  tcmalloc::allocate_full_cpp_throw_oom()
    @     0x56257081c2ca  (unknown)  px::stirling::obj_tools::ElfReader::BinaryByteCode<>()
    @     0x56257081badb  (unknown)  px::stirling::obj_tools::ReadGoString()
    @     0x56257081cdde  (unknown)  px::stirling::obj_tools::ReadGoBuildVersion()
    @     0x562570749126  (unknown)  px::stirling::GoTLSSymAddrs()
    @     0x56257073421f  (unknown)  px::stirling::UProbeManager::UpdateGoTLSSymAddrs()
    @     0x562570738cbf  (unknown)  px::stirling::UProbeManager::AttachGoTLSUProbes()
    @     0x56257073d5aa  (unknown)  px::stirling::UProbeManager::DeployGoUProbes()
    @     0x56257073ecec  (unknown)  px::stirling::UProbeManager::DeployUProbes()
[...]

Issue #1318, opened 2023-05-09
Go uprobes attachment fails for go 1.20.4 and later


ddelnano:
@vihangm tried to upgrade our go version to 1.20.4 (from 1.20.2) in #1266. After investigating this more, I determined that our go version detection fails for these newer binaries and prevents stirling from attaching uprobes to the application. Detecting an application's go version is necessary because applications built with go 1.18 and later must be handled different with our uprobes ([source](https://github.com/pixie-io/pixie/blob/5b9e5e1f419e3d8738a2e84a003422d10c48319c/src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.cc#L469-L479)).

In addition to the go 1.20.4 issues, we recently discovered the same go version detection logic also doesn't handle 32 bit binaries properly (#1300). The `go` cli provides a mechanism for reading the go version of an application and its implementation is very different from what we currently do. We are proceeding with a go 1.20.4 specific fix to unblock #1266, but we need to rehaul our version detection to be more similar to Go's implementation so we can fix #1300 and issues with newer binaries.

ddelnano:
This was fixed in #1605.

Issue #1300, opened 2023-05-08
Parsing 32 bit Go binaries causes large memory allocations


ddelnano:
In order to attach uprobes to Go binaries, Pixie needs to detect what version of go the application was compiled with ([source](https://github.com/pixie-io/pixie/blob/456d9e18f9b3fa6a4887882dc1e0e2e1b2f92395/src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.cc#L469)). There was a recent bug where certain Go binaries caused memory allocation errors (#1111).

While that bug was fixed by enforcing an upper bound on its memory allocations, the underlying cause is due to our inability to properly parse 32 bit go binaries. We use a [hard coded size](https://github.com/pixie-io/pixie/blob/456d9e18f9b3fa6a4887882dc1e0e2e1b2f92395/src/stirling/obj_tools/go_syms.cc#L52) for the `runtime.buildVersion` ELF symbol. This value is correct for 64 bit binaries but is too large for 32 bit binaries and results in Pixie reading garbage memory. The `objdump` commands below show that this symbol is 8 bytes for 32 bit binaries and 16 bytes for 64 bit binaries:

```
$ objdump -Tt main \| grep runtime.buildVersion
objdump: main: not a dynamic object
0000000000525f50 g     O .data  0000000000000010 runtime.buildVersion      <------ 0x10 in size
00000000004b72d8 g     O .rodata        0000000000000009 runtime.buildVersion.str
$ objdump -Tt main-i386 \| grep runtime.buildVersion
objdump: main-i386: not a dynamic object
08166628 g     O .data  00000008 runtime.buildVersion                      <------ 0x8 in size
080f8160 g     O .rodata        00000009 runtime.buildVersion.str
```

We need to update our ELF parsing and tests to verify that we can properly detect Go's build version for 32 bit binaries.

PR #1976, opened 2024-07-29
Address PEM crash caused by parsing certain older Go application binaries

ddelnano:
Summary: Address PEM crash caused by parsing certain older Go application binaries

This PR fixes a crash caused by certain older Go application binaries. In addition, this change includes the `//src/stirling/binaries:go_binary_parse_profiling` cli tool. This tool was helpful for debugging the previous 32 bit issue and aided in debugging this problem (see Background section for more details). This change is best reviewed commit by commit.

**Background**

Our Golang binary parsing was revamped in https://github.com/pixie-io/pixie/pull/1605 to support Go 1.20.4 applications and later (https://github.com/pixie-io/pixie/issues/1318) in addition to fixing a PEM crash caused by 32 bit go binaries (https://github.com/pixie-io/pixie/issues/1300). While this solved the aforementioned issues, it resulted in a new crash that we weren't able to reproduce (#1646).

I was able to work with a Pixie Community Slack user to track down where one of these issues originate from. The overview is that Go embeds virtual addresses within the `.go.buildinfo` ELF section. These virtual addresses are used in certain cases to read the build settings used when the binary was created (toolchain version, go experiments, etc). In order to properly read these strings, these virtual addresses need to be converted into file offsets (binary addresses).

This bug presents itself when the `LOAD` ELF segments in the binary are not contiguous or ordered by increasing virtual memory address. Meaning if there are LOAD segments for segments 1, 2 and 3, this bug occurs if those segments aren't adjacent to each other or don't have increasing virtual memory addresses (vaddr of segment 1 < vaddr of segment 2 < vaddr of segment 3). Instead the virtual address that needs to be looked up, should be matched against the relevant segment and that segment's virtual address offset should be used.

Relevant Issues: Partially addresses #1646 -- there is one more known case, which must be investigated further

Type of change: /kind bug

Test Plan: Verified this change through the following
- [x] User from the community slack [verified](https://pixie-community.slack.com/archives/CQ63KEVFY/p1722271309767939?thread_ts=1721315312.198319&cid=CQ63KEVFY) that the issue was fixed.
- [x] New ElfReader function is covered with a test
- [x] go 1.17 test case ("[little endian](https://github.com/pixie-io/pixie/blob/50ddcd32eb217e1aa5e87124883ee284a36052a1/src/stirling/obj_tools/go_syms_test.cc#L51)" case) still works despite it not triggering this bug
- I was unable to recreate a binary that had the segments in an unordered fashion.

Changelog Message: Fixed an issue with Go uprobe attachment that previously caused crashes for a subset of older Go applications (Go 1.17 and earlier)

### DIFF
--- a/src/stirling/binaries/go_binary_parse_profiling.cc
+++ b/src/stirling/binaries/go_binary_parse_profiling.cc
@@ -29,8 +29,8 @@ using px::stirling::obj_tools::ElfReader;
 // This utility is designed to isolate parsing the debug symbols of a Go binary. This
 // verifies that the go version detection code is functioning as well. This is useful
 // for debugging when the Go elf/DWARF parsing is not working correctly and has been the
-// source of a few PEM crashes (gh#1300, gh#1646). This makes it easy for asking end users to run against
-// their binaries when they are sensitive (proprietary) and we can't debug them ourselves.
+// source of a few PEM crashes (gh#1300, gh#1646). This makes it easy for asking end users to run
+// against their binaries when they are sensitive (proprietary) and we can't debug them ourselves.
 //-----------------------------------------------------------------------------
 
 int main(int argc, char** argv) {
--- a/src/stirling/binaries/go_binary_parse_profiling.cc
+++ b/src/stirling/binaries/go_binary_parse_profiling.cc
@@ -26,7 +26,11 @@ using px::stirling::obj_tools::DwarfReader;
 using px::stirling::obj_tools::ElfReader;
 
 //-----------------------------------------------------------------------------
-// Main
+// This utility is designed to isolate parsing the debug symbols of a Go binary. This
+// verifies that the go version detection code is functioning as well. This is useful
+// for debugging when the Go elf/DWARF parsing is not working correctly and has been the
+// source of a few PEM crashes (gh#1300, gh#1646). This makes it easy for asking end users to run against
+// their binaries when they are sensitive (proprietary) and we can't debug them ourselves.
 //-----------------------------------------------------------------------------
 
 int main(int argc, char** argv) {
--- a/src/stirling/binaries/go_binary_parse_profiling.cc
+++ b/src/stirling/binaries/go_binary_parse_profiling.cc
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2018- The Pixie Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "src/common/base/base.h"
+#include "src/common/base/env.h"
+#include "src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.h"
+
+using px::StatusOr;
+using px::stirling::PopulateGoTLSDebugSymbols;
+using px::stirling::obj_tools::DwarfReader;
+using px::stirling::obj_tools::ElfReader;
+
+//-----------------------------------------------------------------------------
+// Main
+//-----------------------------------------------------------------------------
+
+int main(int argc, char** argv) {
+  px::EnvironmentGuard env_guard(&argc, argv);
+
+  if (argc < 2) {
+    LOG(FATAL) << absl::Substitute("Expected binary argument to be provided. Instead received $0",
+                                   *argv);
+  }
+
+  std::string binary(argv[1]);
+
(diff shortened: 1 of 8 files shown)


src/stirling/obj_tools/elf_reader_test.cc
```diff
@@ -40,6 +40,54 @@ using ::testing::UnorderedElementsAre;
...
+  int64_t vma;
+  int64_t lma;
+  int64_t file_offset;
+};
+
+StatusOr<Section> ObjdumpSectionNameToAddr(const std::string& path,
+                                           const std::string& section_name) {
+  Section section;
+  std::string objdump_out =
+      px::Exec(absl::StrCat("objdump -h -j ", section_name, " ", path)).ValueOrDie();
```
oazizi000, line 68: Using `px::Exec` breaks the hermeticity of the test. Might be better to run this externally and put the required assets into `testdata`.

src/stirling/binaries/go_binary_parse_profiling.cc
```diff
@@ -0,0 +1,67 @@
...
+#include "src/stirling/source_connectors/socket_tracer/uprobe_symaddrs.h"
+
+using px::StatusOr;
+using px::stirling::PopulateGoTLSDebugSymbols;
+using px::stirling::obj_tools::DwarfReader;
+using px::stirling::obj_tools::ElfReader;
+
+//-----------------------------------------------------------------------------
+// Main
+//-----------------------------------------------------------------------------
```
oazizi000, line 30: leave a comment on the purpose of the utility. There's an explanation on the PR, but would be useful to encode an explanation into this file.

oazizi000 (approved):

src/stirling/obj_tools/elf_reader_test.cc
ddelnano: While there is a hermeticity risk, the expectations are based on the [test_exe target](https://github.com/pixie-io/pixie/blob/50ddcd32eb217e1aa5e87124883ee284a36052a1/src/stirling/obj_tools/testdata/cc/BUILD.bazel#L25), so it's possible that the `test_exe` binary and the objdump output may become out of sync. While the `test_exe` binary could be added as testdata as well, there are other occurrences of accessing the binary directly via the [test_exe_fixture.h header](https://github.com/pixie-io/pixie/blob/50ddcd32eb217e1aa5e87124883ee284a36052a1/src/stirling/obj_tools/testdata/cc/test_exe_fixture.h#L33).
```
$ git grep test_exe_fixture.h
src/stirling/bpf_tools/bcc_wrapper_bpf_test.cc:#include "src/stirling/obj_tools/testdata/cc/test_exe_fixture.h"
src/stirling/obj_tools/elf_reader_test.cc:#include "src/stirling/obj_tools/testdata/cc/test_exe_fixture.h"
src/stirling/obj_tools/testdata/cc/BUILD.bazel:    hdrs = ["test_exe_fixture.h"],
src/stirling/source_connectors/dynamic_bpftrace/dynamic_bpftrace_connector_bpf_test.cc:#include "src/stirling/obj_tools/testdata/cc/test_exe_fixture.h"
src/stirling/source_connectors/dynamic_tracer/dynamic_trace_bpf_test.cc:#include "src/stirling/obj_tools/testdata/cc/test_exe_fixture.h"
```

We also follow this pattern for `nm` [cli output](https://github.com/pixie-io/pixie/blob/50ddcd32eb217e1aa5e87124883ee284a36052a1/src/stirling/obj_tools/elf_reader_test.cc#L46) and haven't had issues with these tests to my knowledge.

My preference would be to keep it as is, but happy to proceed with your suggestion if you feel strongly about it.

oazizi000: Just a few thoughts in response:

> "it's possible that the test_exe binary and the objdump output may become out of sync":

If we do check it in, I don't think the risk of them getting out of sync is a concern, because the tests would catch the mismatch.

> "While the test_exe binary could be added as testdata as well, there are other occurrences of accessing the binary directly via the [test_exe_fixture.h header](https://github.com/pixie-io/pixie/blob/50ddcd32eb217e1aa5e87124883ee284a36052a1/src/stirling/obj_tools/testdata/cc/test_exe_fixture.h#L33)."

Yeah, I'm not suggesting to check in `test_exe`. Wouldn't really make sense because it's already in the code-base.

> We also follow this pattern for nm [cli output](https://github.com/pixie-io/pixie/blob/50ddcd32eb217e1aa5e87124883ee284a36052a1/src/stirling/obj_tools/elf_reader_test.cc#L46) and haven't had issues with these tests to my knowledge.

That one's probably not great either :)

Another option is to have bazel generate the testdata by running objdump. That way it stays in sync. But not sure that's worth the hassle.

Having said all that, I don't feel that strongly that you should consider this a blocker. So it's okay by me if you want to put a `TODO` to revisit these two cases in the future, and land as is.

ddelnano: I've opened https://github.com/pixie-io/pixie/pull/1980 to update the nm usage to be hermetic. I'm not sure if doing the same for `objdump` will be as easy. (since nm has [built-in custom makefile variable](https://bazel.build/reference/be/make-variables#custom_variables) support, while `objdump` doesn't).

If that ends up being easy to do, I'll fix this otherwise I'll add a `TODO` for later.

ddelnano: I've added the TODO and will address this once I hear from James and Vihang regarding how to access `objdump` from the bazel toolchain.

PR #1980, opened 2024-08-13
Update `ElfReaderTest`'s nm usage to be hermetic

ddelnano:
Summary: Update `ElfReaderTest`'s nm usage to be hermetic

This was feedback from #1976. See [this comment](https://github.com/pixie-io/pixie/pull/1976#discussion_r1707595841) for the previous discussion.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Elf reader tests pass with updates

### DIFF
--- a/bazel/cc_toolchains/clang/toolchain_files.BUILD
+++ b/bazel/cc_toolchains/clang/toolchain_files.BUILD
@@ -60,6 +60,12 @@ filegroup(
 )
 
 filegroup(
+    name = "toolchain_nm_files",
+    srcs = [":nm"],
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
     name = "toolchain_strip_files",
     srcs = [":strip"],
     visibility = ["//visibility:public"],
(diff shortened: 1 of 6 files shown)


bazel/cc_toolchains/clang/toolchain_files.BUILD
```diff
@@ -110,6 +116,7 @@ filegroup(
         "as",
         "cov",
         "objcopy",
+        "nm",
```
ddelnano, line 119: @JamesMBartlett @vihangm is it possible to get access to a toolchain's other tools? Ideally I'd use the toolchain version of `objdump` for the test added in #1976 (see discussion linked in this PR's description). Unlike `nm` and `objcopy`, which have [custom Makefile variables](https://bazel.build/reference/be/make-variables#custom_variables) built-in, `objdump` doesn't have an existing variable.

It seems that [TemplateVariableInfo](https://www.stevenengelhardt.com/2023/03/03/practical-bazel-custom-bazel-make-variables/) could work for this, but it feels a little strange that it wouldn't be toolchain specific the way `@bazel_tools//tools/cpp:current_cc_toolchain` is.

oazizi000 (approved):

vihangm (approved):

Issue #1829, opened 2024-01-24
Populate local IP address from BPF space


oazizi000:
**Is your feature request related to a problem? Please describe.**
With #1808, we collect local IP addresses from eBPF space when doing server-side tracing. For consistency, we should investigate adding support for collecting the local IP address with client-side tracing. This will provide more consistency in understanding of our tables for all users. 

**Describe the solution you'd like**
Collect and populate local IP address in any tables that contain the local IP address.

**Describe alternatives you've considered**
Alternative solutions could try to populate it from user-space, but BPF-based collection is preferred for performance and reliability reasons.

**Additional context**
For background see #1807 #1808 and #1809

benkilimnik:
I believe we can merge `http_events` with `tcp-stats-events` using the [tcp stats connector](https://github.com/pixie-io/pixie/blob/fdce3fd44f55a3079ebc282a806e77f3b1de7656/src/stirling/stirling.cc#L131).

Ran a quick test in PxL and this seems to do the trick.
```py
import px
df = px.DataFrame(table='http_events', select=['time_', 'local_port', 'local_addr', 'remote_port', 'remote_addr', 'trace_role'], start_time='-100s')
df = df[df.trace_role == 1] # client-captured
df = df[df.local_port == -1] # couldn't parse port (and addr)
df2 = px.DataFrame(table="tcp_stats_events", start_time='-100s', select=["time_","local_addr", "local_port", "remote_addr", "remote_port"])
merged = df.merge(df2, how='inner', left_on=["remote_addr", "remote_port"], right_on=["remote_addr", "remote_port"], suffixes=['_http', '_tcp'])
px.display(df, "http events")
px.display(merged, "merged with tcp")
```

For this to work, the TCP stats source connector needs to enabled in stirling. Perhaps it would be worth adding it to `kProd`?

PR #1989, opened 2024-08-23
Populate client side trace's local address via tcp kprobes

ddelnano:
Summary: Populate client side trace's local address via tcp kprobes

This change populates client side trace's `local_addr` and `local_port` columns for the following use cases:
1. To provide more consistency for the protocol data tables. Having columns that are empty make it difficult for end users to understand what is being traced and make them less useful
2. To facilitate addressing a portion of the short lived process problems (#1638)

For 2, the root of the issue is that `df.ctx["pod"]` syntax relies on the [px.upid_to_pod_name](https://docs.px.dev/reference/pxl/udf/upid_to_pod_name/) function. If a PEM misses the short lived process during its metadata update, this function fails to resolve the pod name. For client side traces where the pod is making an outbound connection (non localhost), the `local_addr` column provides an alternative pod name lookup for short lived processes when the pod is long lived. This means the following would be equivalent to the `df.ctx["pod"]` lookup:  `px.pod_id_to_pod_name(px.ip_to_pod_id(df.local_addr))`.

I intend to follow this PR with a compiler change that will make `df.ctx["pod"]` try both methods should `px.upid_to_pod_name` fail to resolve. This will allow the existing pxl scripts to display the previously missed short lived processes.

**Alternatives**

Another approach I considered was expanding our use of the `sock_alloc` kprobe. I used ftrace on a simple curl command to see what other options could be used (`sudo trace-cmd record -F -p function_graph http://google.com`). The `socket` syscall calls `sock_alloc`, which would be another mechanism for accessing the `struct sock`. I decided against this approach because I don't think its viable to assume that the same thread/process that calls `socket` will be the one that does the later syscalls (how our BPF maps are set up). It's common to have a forking web server model, which means a different process/thread can call `socket` than the ones that later read/write to it.

**Probe stability**

These probes appear to be stable from our oldest and newest supported kernel. These functions exist in the [tcp_prot](https://elixir.bootlin.com/linux/v4.14.336/source/net/ipv4/tcp_ipv4.c#L2422), [tcpv6_prot](https://elixir.bootlin.com/linux/v4.14.336/source/net/ipv6/tcp_ipv6.c#L1941) structs and I've seen that other projects and bcc tools use these probes. This makes me believe that these functions have a pretty well defined interface.

Relevant Issues: #1829, #1638

Type of change: /kind feature

Test Plan: New tests verify that ipv4 and ipv6 cases work
- [x] Ran `for i in $(seq 0 1000); do curl http://google.com/$i; sleep 2; done` within a pod and verified that `local_addr` is populated with this change and `px.pod_id_to_pod_name(px.ip_to_pod_id(df.local_addr))` works for pod name resolution.

- [x] Verified the above curl test results in traces without `local_addr` without this change
local-addr-testing
- Tested on the following k8s offerings and machine images
- [x] GKE COS and Ubuntu
- [x] EKS Amazon Linux 2

Changelog Message: Populate socket tracer data table `local_addr` and `local_port` column for client side traces.


src/stirling/source_connectors/socket_tracer/bcc_bpf/socket_trace.c
```diff
@@ -585,6 +585,57 @@ int conn_cleanup_uprobe(struct pt_regs* ctx) {
...
+// Using the struct sock* for capturing a socket's local address only works for TCP sockets.
+// The equivalent UDP functions (udp_v4_connect, udp_v6_connect and upd_sendmsg) always receive a
+// sock struct with a 0.0.0.0 or ::1 local address. This is deemed acceptable since our local
+// address population for server side tracing relies on accept/accept4, which only applies for TCP.
+//
+// TODO(ddelnano): The current implementation works for mid stream TCP connections despite
+// my intuition that tcp_v4_connect and tcp_v6_connect should not be called mid stream.
+// If mid stream connections have a missing local address in the future, we should probe
+// tcp_sendmsg in addition to the current probes. This will require discerning between the
+// UDP and TCP case in process_implicit_conn.
```
ddelnano, line 606: I tried to test this mid stream case by creating a psql shell prior to the PEM starting and issuing queries after the socket tracer was initialized. I verified that a tcp connection was made and only `sendto` syscalls were issued when queries were executed (no `connect` syscalls) to check that this would simulate a mid stream case.

Surprisingly, I saw that the `local_addr` columns were populated despite my thinking that it should be missed. I intend to add a small addition to this change to cover the mid stream case even though I can't seem to reproduce it failing. This would work by grabbing the `sock*` struct from one additional kprobe `tcp_sendmsg` and handling it in the process_implict_conn. The slight complication here is that the UDP case mentioned above needs to be filtered out. 

I wanted to get feedback on the current scope of this before doing that part and to see there were ideas on why the testing above seemed to work.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -342,6 +342,18 @@ const auto kProbeSpecs = MakeArray<bpf_tools::KProbeSpec>({
...
+    {"tcp_v4_connect", ProbeType::kReturn, "probe_ret_populate_active_connect_sock",
+     /*is_syscall*/ false},
+    {"tcp_v6_connect", ProbeType::kEntry, "probe_entry_populate_active_connect_sock",
+     /*is_syscall*/ false},
+    {"tcp_v6_connect", ProbeType::kReturn, "probe_ret_populate_active_connect_sock",
+     /*is_syscall*/ false},
+    {"tcp_sendmsg", ProbeType::kEntry, "probe_entry_populate_active_connect_sock",
+     /*is_syscall*/ false},
+    {"tcp_sendmsg", ProbeType::kReturn, "probe_ret_populate_active_connect_sock",
+     /*is_syscall*/ false},
```
ddelnano, line 356: Since we don't have an easy mechanism for testing this, I tried to test this mid stream case by creating a psql shell prior to the PEM starting and issuing queries after the socket tracer was initialized. I verified that a tcp connection was made and only `sendto` syscalls were issued when queries were executed (no `connect` syscalls) to check that this would simulate a mid stream case.

Surprisingly, I saw that the `local_addr` column was populated before I added the probe on `tcp_sendmsg` despite my thinking that it should be empty. I went ahead and added this probe since I verified with `ftrace` on the same psql process that `tcp_v4_connect` isn't called. Just raising this since I wasn't able to verify that this additional probe adds extra coverage from the testing I did.

ddelnano:
@oazizi000 would appreciate if you could give this a review. I still need to fix the bpf instruction issue for the 4.x kernels and the incorrect storage of the local port (it is stored in network byte order and is an existing issue), but that shouldn't impact the core details in the PR.

src/stirling/source_connectors/socket_tracer/socket_trace_connector.cc
```diff
@@ -176,7 +176,7 @@ DEFINE_bool(
     stirling_debug_tls_sources, gflags::BoolFromEnv("PX_DEBUG_TLS_SOURCES", false),
     "If true, stirling will add additional prometheus metrics regarding the traced tls sources");
 
-DEFINE_uint32(stirling_bpf_loop_limit, 42,
+DEFINE_uint32(stirling_bpf_loop_limit, 41,
```
ddelnano, line 179: The iovec syscall return probes that rely on process_implicit_conn (recvmsg, recvmmsg, sendmsg, sendmmsg) had their instruction count increase beyond the 4.14 limit. This was my solution to getting the BPF instruction count on 4.14 kernels to work. The rationale is that cases that are right on the edge of this limit are likely to have problems already.

The other things I considered were the following:
- Remove match_trace_tgid logic on 4.14 kernels -- did not lower the instruction count enough
- Make the iovec ret syscalls provide a NULL sock for its `submit_new_conn` call (e.g. `submit_new_conn(ctx, tgid, fd, addr, /*sock*/ NULL, role, source_fn)`) -- not viable since it partially handles mid stream connections

I'm open to other suggestions if you have any, but this was the set of things I came up with.

src/stirling/source_connectors/socket_tracer/socket_trace_bpf_test.cc
```diff
@@ -747,6 +750,163 @@ TEST_F(NullRemoteAddrTest, IPv6Accept4WithNullRemoteAddr) {
...
+  uint16_t port = ntohs(client_sockaddr.sin6_port);
+  for (auto lport : GetLocalPorts(records, kHTTPLocalPortIdx, indices)) {
+    // TODO(ddelnano): Determine why the local_addr column is storing the port
+    // in network byte order.
+    LOG(INFO) << "Local port: " << lport << " and pre ntohs " << client_sockaddr.sin6_port
+              << " and ntohs " << port;
+    if (lport == client_sockaddr.sin6_port) {
+      found_port = true;
+      break;
+    }
```
ddelnano, line 905: My suspicion for why this is happening is that we are calling `ntohs` more than once and so we end up reversing the endianess back to network byte order. I'm still in the process of tracking it down.

oazizi000:
Approach looks good. We should just manually test on a few more endpoints to gain more confidence. If that all looks good, then this should be good to go.

src/stirling/source_connectors/socket_tracer/socket_trace_bpf_test.cc
ddelnano: The kernel's skc_num (also referred to as [sk_num](https://elixir.bootlin.com/linux/v6.10.7/source/include/net/sock.h#L359)) is stored in host byte order. I verified this through `bpf_trace_printk` calls and this example from the kernel [source](https://elixir.bootlin.com/linux/v6.10.7/source/net/ipv4/inet_diag.c#L79).

ddelnano:
@oazizi000 I've tested this change on EKS (Amazon Linux 2), GKE Ubuntu and GKE COS in addition to tracking down the endianness problem (kernel data structure stores the local port in host byte order).

I've also created #2002 to make that byte order more clear even though that case doesn't have a bug.

oazizi000 (approved):

PR #1992, opened 2024-08-23
Update kernel builder to work with kernels 6.6 and later. Add 6.8.x to kernel build list

ddelnano:
Summary: Update kernel builder to work with kernels 6.6 and later. Add 6.8.x to kernel build list

Sicne Ubuntu's current LTS release runs on a 6.8.x kernel, I wanted to keep our newest kernel version in line with that. If/when this is approved, I'll proceed with running the [upload](https://github.com/pixie-io/pixie/blob/33244cd28d79058792a6cc06fe2637f491cd51a7/tools/docker/Makefile#L288) (side stepping the GCS part) and then add this to the [all_kernel_versions](https://github.com/pixie-io/pixie/blob/33244cd28d79058792a6cc06fe2637f491cd51a7/ci/github/matrix.sh#L31) ci matrix.

Relevant Issues: N/A

Type of change: /kind feature

Test Plan: Ran the makefile target for an older and newer kernel to verify it works now


tools/docker/Makefile
```diff
@@ -131,7 +131,8 @@ KERNEL_BUILD_VERSIONS := 4.14.254 \
 	5.4.235 \
 	5.10.173 \
 	5.15.101 \
-	6.1.18
+	6.1.18 \
```
vihangm, line 134: I think we try to keep these on the newest patch versions. Consider bumping to `5.4.254`, `5.10.224`, `5.15.165`, `6.1.106`

vihangm (approved):

Issue #2013, opened 2024-09-09
Provide mechanism for opting cloud proxy out of listening on an ipv6 address


ddelnano:
A slack community member reported that the cloud proxy service will fail to start if it's deployed to an instance that has ipv6 disabled. This manifests through the following error and happens because of this line.

```
2024-09-05T09:35:25.025152669Z 2024/09/05 09:35:24 [emerg] 8#8: socket() [::]:56000 failed (97: Address family not supported by protocol) 2024-09-05T09:35:25.025172354Z nginx: [emerg] socket() [::]:56000 failed (97: Address family not supported by protocol)
```

Ideally the cloud proxy service could be configured to avoid trying to listen on this address. This could be driven via an environment variable, so that end users that need this behavior can opt into it.

ddelnano:
This issue and #2016 are covered by #2018. This will be available in the next cloud release.

PR #2016, opened 2024-09-10
Provide mechanism for config based kube dns service

JerrinAndrei:
Currently the nginx.conf expects the a kubernetes dns service named kube-dns.kube-system.svc.cluster.local to be existing in the cluster. If this dns service does not exist the cloud-proxy pod gets terminated on startup. This issue has been addressed in this PR. Here, we are moving the kubeernetes dns service name to a config and updating the nginx conf with this name at startup. This will prevent the container failing to be up on clusters that do not have a standard naming convention for the dns service.

### DIFF
--- a/k8s/cloud/base/domain_config.yaml
+++ b/k8s/cloud/base/domain_config.yaml
@@ -4,5 +4,6 @@ kind: ConfigMap
 metadata:
   name: pl-domain-config
 data:
+  PL_KUBE_DNS_SERVICE: kube-dns.kube-system.svc.cluster.local
   PL_DOMAIN_NAME: dev.withpixie.dev
   PASSTHROUGH_PROXY_PORT: "4444"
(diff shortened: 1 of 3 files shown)


ddelnano:
@JerrinAndrei thanks for opening this PR and the GitHub issue! After thinking about this and the IPV6 fix (#2013) more, I think the existing `sed` approach isn't flexible enough to accommodate additional use cases. If/when more one-off changes need to be made, it will require additional changes and cloud releases.

The ideal the solution would make it possible to facilitate additional nginx configuration tweaks should another use case arise. After discussing this with the team today, we believe #2018 will serve both of these use cases and future ones well. Would appreciate any feedback you have on that approach and we can proceed with that instead of my earlier IPv6 PR and this one.

ddelnano:
Thanks for raising this issue @JerrinAndrei. Since I didn't hear back on my earlier comment, I decided to proceed with #2018. Please let me know if you have any trouble using that and I'm happy to help.

JerrinAndrei:
This looks good to me @ddelnano  . Most of the components in nginx conf would vary from cluster to cluster, so moving this to a configmap would give the entire configuration control to the user.
Eagerly awaiting for the release this week. :)

Issue #2017, opened 2024-09-10
Cloud-proxy container fails to startup and throws an error. May be fixed by https://github.com/pixie-io/pixie/pull/2016


JerrinAndrei:
**Describe the bug**
Currently the nginx.conf expects the a kubernetes dns service named kube-dns.kube-system.svc.cluster.local to be existing in the cluster. If this dns service does not exist the cloud-proxy pod gets terminated on startup.

**To Reproduce**
1. Go to https://docs.px.dev/installing-pixie/install-guides/self-hosted-pixie/
2. On executing step 12 to "Deploy Pixie Cloud",
kustomize build k8s/cloud/public/ | kubectl apply -f -
The cloud proxy pod fails to start with error
"host not found in resolver kube-dns.kube-system.svc.cluster.local in /usr/local/openresty/nginx/conf/nginx.conf:43"

**Expected behavior**
Cloud proxy Pod should startup successfully with existing kube dns if any.
Every kubernetes cluster can have different kube dns service names. Also some clusters can have core-dns. So the cloud proxy service should look up for the dns service specific to the cluster in which its being setup. The control for defining this this should be given to the user. 

This might be fixed in https://github.com/pixie-io/pixie/pull/2016

ddelnano:
This is covered by https://github.com/pixie-io/pixie/pull/2018 and will be available in the next cloud release.

PR #2014, opened 2024-09-09
Provide mechanism for cloud proxy to opt out of listening on ipv6 address

ddelnano:
Summary: Provide mechanism for cloud proxy to opt out of listening on ipv6 address

This was a request from a community member in slack. They have a cluster that has ipv6 disabled and their cloud proxy service is unable to start.

Relevant Issues: Closes #2013

Type of change: /kind feature

Test Plan: Deployed to a cloud cluster and verified that setting `PL_IPV6_DISABLED ` templates the nginx file correctly

Changelog Message: Provide mechanism for opting the cloud proxy service out of listening on an ipv6 address

PR #2018, opened 2024-09-10
Remove nginx config files from cloud proxy container in favor of Configmaps for easier runtime overrides

ddelnano:
Summary: Remove nginx config files from cloud proxy container in favor of Configmaps for easier runtime overrides

This is an alternative approach to #2014 and #2016. While this doesn't provide an environment variable for configuring the intended behavior, this approach is more flexible since many Nginx directives don't work with variables (`server_name`, `resolver`, among others ).

Because nginx prohibits variables in these directives, it makes it very difficult to provide environment variable based settings without our previous `sed` approach. The `sed` approach also has its problems since it requires [hacks](https://github.com/pixie-io/pixie/pull/2014/files#diff-5ec7ca8d0f624fe1f4eb3778cc96dcee2f999bf39bad422807b67b15ce2f8e7bR27) to support configuration removals. Rather than trying to solve all potential use cases, this PR opts to make the configuration easy to swap out via the `pl-proxy-nginx-config` Configmap.

I plan to update the self hosted cloud docs to call out that this Configmap exists and should be used if custom nginx configuration is needed outside of the upstream defaults.

Relevant Issues: #2017

Type of change: /kind feature

Test Plan: Deployed to a cloud environment and verified that the upstream defaults and `PL_DOMAIN_NAME` apply as expected

Changelog Message: Removed nginx configuration from the container image into `pl-proxy-nginx-config` Configmap for easier runtime overrides


k8s/cloud/base/proxy_deployment.yaml
```diff
@@ -78,6 +78,36 @@ spec:
         volumeMounts:
         - name: certs
           mountPath: /certs
+        - name: nginx-config
```
vihangm, line 81: I think you should be able to mount the configmap as a directory of files without having to specify each file.

ddelnano: Good call. Verified that mounting the entire directory worked 👍

aimichelle (approved):

PR #2039, opened 2024-10-08
Add split UDF func to ease certain string handling

ddelnano:
Summary: Add split UDF func to ease certain string handling

This is functionality that I'm planning to use for my upcoming Kubecon demo/talk. For this use case, I want to access the individual fields of the XFCC header just like the example in the UDF docstring.

Relevant Issues: N/A

Type of change: /kind feature

Test Plan: New tests pass

Changelog Message: Added `px.split` function to support parsing strings that contain delimiters

### DIFF
--- a/src/carnot/funcs/builtins/json_ops.h
+++ b/src/carnot/funcs/builtins/json_ops.h
@@ -235,22 +235,19 @@ class SplitUDF : public udf::ScalarUDF {
     rapidjson::Writer<rapidjson::StringBuffer> writer(sb);
     writer.StartArray();
 
-    std::string_view s(in.data(), in.size());
-    size_t idx = 0;
-    while (idx < s.size()) {
-      auto next_idx = s.find(delimiter.data(), idx);
-      if (next_idx == std::string::npos) {
-        if (idx > 0) {
-          writer.String(s.substr(idx).data(), s.substr(idx).size());
-        }
-        break;
+    std::vector<absl::string_view> parts = absl::StrSplit(in.data(), delimiter.data());
+
+    // Verify that delimiter was found in the input string before
+    // writing the array entries.
+    if (parts.size() > 1 || parts[0] != in.data()) {
+      for (auto part : parts) {
+        writer.String(part.data(), part.size());
       }
-      writer.String(s.substr(idx, next_idx - idx).data(), next_idx - idx);
-      idx = next_idx + delimiter.size();
     }
     writer.EndArray();
     return sb.GetString();
   }
+
   static udf::ScalarUDFDocBuilder Doc() {
     return udf::ScalarUDFDocBuilder(
                "Splits a string by a delimiter and a returns JSON encoded array of strings.")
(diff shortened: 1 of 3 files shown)


src/carnot/funcs/builtins/json_ops.h
```diff
@@ -228,6 +228,49 @@ class PluckArrayUDF : public udf::ScalarUDF {
...
+class SplitUDF : public udf::ScalarUDF {
+ public:
+  StringValue Exec(FunctionContext*, StringValue in, StringValue delimiter) {
+    rapidjson::StringBuffer sb;
+    rapidjson::Writer<rapidjson::StringBuffer> writer(sb);
+    writer.StartArray();
+
+    std::string_view s(in.data(), in.size());
+    size_t idx = 0;
+    while (idx < s.size()) {
```
JamesMBartlett, line 240: Can we just use `absl::StrSplit` instead? Then write the results to the json array.

src/carnot/funcs/builtins/json_ops.h
```diff
@@ -228,6 +228,49 @@ class PluckArrayUDF : public udf::ScalarUDF {
...
+        .Details(
+            "This function splits a string by a delimiter and returns a JSON encoded array of "
+            "strings. The function is useful for splitting strings and then passing the result to "
+            "px.pluck_array in order to access individual values of a delimited string.")
+        .Example(R"doc(
+          | df = px.DataFrame('http_events', start_time='-5m')
+          | # Returns By=http://frontend.px.dev;URI=http://testclient.px.dev
+          | df.xfcc_hdr = px.pluck(df.req_headers, 'X-Forwarded-Client-Cert')
+          | df.by= px.pluck_array(df.xfcc_hdr, 0) # Returns "By=http://frontend.px.dev"
+          | df.uri = px.pluck_array(df.xfcc_hdr, 1) # Returns "URI=http://testclient.px.dev"
```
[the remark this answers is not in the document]
ddelnano: Whoops :sweat_smile:. Included `px.split` and made sure the spacing between the assignment operators is consistent as well.

src/carnot/funcs/builtins/json_ops.h
ddelnano: Yep, didn't realize it handled using `string_view`s out of the box. Updated.

src/carnot/funcs/builtins/json_ops.h
```diff
@@ -228,6 +228,47 @@ class PluckArrayUDF : public udf::ScalarUDF {
...
+  StringValue Exec(FunctionContext*, StringValue in, StringValue delimiter) {
+    rapidjson::StringBuffer sb;
+    rapidjson::Writer<rapidjson::StringBuffer> writer(sb);
+    writer.StartArray();
+
+    std::vector<absl::string_view> parts = absl::StrSplit(in.data(), delimiter.data());
+
+    // Verify that delimiter was found in the input string before
+    // writing the array entries.
+    if (parts.size() > 1 || parts[0] != in.data()) {
```
JamesMBartlett, line 242: I think typically, Split would return the whole string as the first element of the array if the delimiter is not in the string.

ddelnano: Not sure why I went this direction, but staying consistent with the API of other split functions is definitely the right thing. Updated.

JamesMBartlett (approved):

PR #2043, opened 2024-11-04
[1/2] Support rendering differential flamegraphs in the `StackTraceFlameGraph` display spec

ddelnano:
Summary: Support rendering differential flamegraphs in the `StackTraceFlameGraph` display spec

This change makes it possible to render a [differential flamegraphs](https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html). These are useful for comparing flamegraph profiles against each other in order to see what code paths are called more or less than the baseline profile. When a Vis spec specifies a `differenceColumn`, the widget will render this differential flamegraph. The difference column computes the sample delta between the baseline profile and the new profile.

Relevant Issues: N/A

Type of change: /kind feature

Test Plan: New unit tests pass and pxl script relying on this works as expected (see https://github.com/pixie-io/pixie/pull/2044)

differential

Changelog Message: Add support for rendering differential flamegraphs in the `StackTraceFlameGraph` display spec


NickLanam (approved):
Cool feature!

aimichelle (approved):

PR #2044, opened 2024-11-04
[2/2] Add differential flamegraph script

ddelnano:
Summary: Add differential flamegraph script

This depends on #2043.

Relevant Issues: N/A

Type of change: /kind new-pxl-script

Test Plan: Ran the local UI and verified that the non negated version works as expected
differential

Changelog Message: Add `px/differential_flamegraph` script for comparing flamegraphs profiles between pods. See [this post](https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html) for more background on how to use a differential flamegraph.

### DIFF
--- a/src/pxl_scripts/px/differential_flamegraph/differential.pxl
+++ b/src/pxl_scripts/px/differential_flamegraph/differential.pxl
@@ -1,0 +1,50 @@
+# Copyright 2018- The Pixie Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import px
+
+def export_flame_graph(start_time: str, namespace: str, pod: str, baseline_pod: str):
+  stack_traces = px.DataFrame(table='stack_traces.beta', start_time=start_time)
+  stack_traces.namespace = stack_traces.ctx['namespace']
+  stack_traces = stack_traces[px.contains(stack_traces.namespace, namespace)]
+  stack_traces.node = px.Node(px._exec_hostname())
+  stack_traces.pod = stack_traces.ctx['pod']
+  stack_traces.keep_row = px.select(px.contains(stack_traces.pod, baseline_pod), True, False)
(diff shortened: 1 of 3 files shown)


src/pxl_scripts/px/differential_flamegraph/differential.pxl
```diff
@@ -0,0 +1,77 @@
...
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import px
+
+negate = False
+# TODO(ddelnano): negation requires switching the type of join from right to left
+# or returning a different DataFrame. This might not be possible with pxl's current
+# functionality, but this should be implemented once it's possible.
```
ddelnano, line 22: Ideally the `negate` parameter would be a script variable. In order to compute the negated flamegraph properly, the `how` on line 26 must change from right to left in addition to changing the `groupby` on line 35 to be against the `pod1` dataframe.

I tried to implement this by calling `merge_and_compute_delta` with swapped parameters or to use `px.select` to pick the correct join type (right vs left). Neither of these seemed to work. I think we might need some additional pxl support in order to accomplish this, but please let me know if you have any ideas.

NickLanam:
If you make this PR target the other one instead of `main`, then it'll only show this PR's changes. GitHub will magically change the target back when the other one merges.

ddelnano:
@NickLanam thanks for the tip and I'll give that a try on future PRs to avoid recreating this one.

src/pxl_scripts/px/differential_flamegraph/vis.json
```diff
@@ -0,0 +1,66 @@
...
+                        "variable": "baseline_pod"
+                    }
+                ]
+            },
+            "displaySpec": {
+                "@type": "types.px.dev/px.vispb.StackTraceFlameGraph",
+                "stacktraceColumn": "stack_trace",
+                "countColumn": "count",
+                "percentageColumn": "percent",
+                "podColumn": "pod",
```
ddelnano, line 61: a54f66098bc25560770fb1575f91621900a44083 added the `podColumn` column. From testing the original change (everything before a54f66098bc25560770fb1575f91621900a44083), I would very infrequently receive crashes on [this line](https://github.com/pixie-io/pixie/blob/4778ba1f1f28f0e7019e33747dfc016d4eaaae6c/src/ui/src/containers/live/convert-to-vega-spec/flamegraph.ts#L1019).

I unfortunately don't have the stack trace handy, but the only thing I can attribute it to is that it's possible for the `stacktraceColumn` field to result in an undefined or empty value that causes the crash. My suspicion is this doesn't happen for the `px/perf_flamegraph` script because it forces the scopedStackTrace variable to contain semicolons and `(k8s)` ([source](https://github.com/pixie-io/pixie/blob/4778ba1f1f28f0e7019e33747dfc016d4eaaae6c/src/ui/src/containers/live/convert-to-vega-spec/flamegraph.ts#L994-L1016)).

aimichelle (approved):

Issue #2051, opened 2024-12-02
Warn users that kernel headers are missing during the Pixie install process


ddelnano:
While Pixie has invested in its prepackaged linux headers and working without upstream headers, it is highly recommended to install the given distro's kernel header package. Upstream distros patch and backport many changes, which make the prepackaged option susceptible issues that are hard to anticipate and work around. Examples of these inconsistencies can be seen in #1863, #252 and the recent openSUSE issues ([2037](https://github.com/pixie-io/pixie/issues/2037)).

Some of these issues get reported, but my suspicion is that this poor experience causes people to fail to fully evaluate Pixie since their initial impression shows that the socket tracer isn't functional (the most common way this problem manifests). For example, the openSUSE case mentioned above was only determined through my outreach and was in a position where the end user had moved on from evaluating Pixie.

If Pixie had the ability to detect when kernel headers aren't installed, we could warn the user that it is recommended to do so and link to common problems caused by the lack of headers. This will provide the end user with quick feedback on an area that's currently arcane to debug and hopefully prevent people from having a poor experience in these cases.

ddelnano:
https://github.com/pixie-io/pixie/issues/1986 is another great case of the need for this warning/tooling. That was a bug that lasted from August until now and was due to the fact that Amazon linux headers needed to be installed since pixie's pre-packaged headers resulted in broken Go TLS tracing.

ddelnano:
Users running Vizier v0.14.14 and Cloud v0.1.9 will now see these kernel warning checks when using the `px` cli to deploy. This warning is also included in `px collect-logs` since that is a prerequisite to opening a GitHub issue.

We need to add this warning check to Helm installs, but I'll probably be revisiting that at a later time.

PR #2052, opened 2024-12-02
Add UDTF that detects linux kernel header installation and add column to `GetAgentStatus`

ddelnano:
Summary: Add UDTF that detects linux kernel header installation and add column to `GetAgentStatus`

This is a prerequisite to accomplish #2051. The `px deploy` command uses the GetAgentStatus UDTF in its final [healthcheck step](https://github.com/pixie-io/pixie/blob/854062111cf4b91a40649a2e2647c88c0a68b0db/src/pixie_cli/pkg/cmd/deploy.go#L607-L613). With this kernel header detection in place, the `px` cli can use the results from the `px/agent_status` script to print a warning message if kernel headers aren't detected.

The helm install flow needs to be covered as well. My hope is that this UDTF could be used for that use case as well, but I need to further investigate the details of that.

Relevant Issues: #2051

Type of change: /kind feature

Test Plan: Skaffolded to a Ubuntu GKE cluster and tested the following
- [x] Kelvin always reports `false` as it doesn't bind mount `/` to `/host`
- [x] PEM running on host without `linux-headers-$(uname -r)` package reports `false`
- [x] PEM running on host with `linux-headers-$(uname -r)` package reports `true`
```
$ gcloud compute ssh gke-dev-cluster-ddelnano-default-pool-a27c1ac2-x5k2 --internal-ip -- 'ls -alh /lib/modules/$(uname -r)/build'

lrwxrwxrwx 1 root root 38 Aug  9 15:25 /lib/modules/5.15.0-1065-gke/build -> /usr/src/linux-headers-5.15.0-1065-gke

$ gcloud compute ssh gke-dev-cluster-ddelnano-default-pool-a27c1ac2-j6pg --internal-ip -- 'ls -alh /lib/modules/$(uname -r)/build'

ls: cannot access '/lib/modules/5.15.0-1065-gke/build': No such file or directory

```
Screen Shot 2024-12-02 at 9 30 29 AM

Changelog Message: Add `GetLinuxHeadersStatus` UDTF and add `kernel_headers_installed` column to `GetAgentStatus`


src/vizier/services/agent/pem/pem_main.cc
```diff
@@ -68,9 +71,19 @@ int main(int argc, char** argv) {
...
+
+  auto kernel_headers_installed = false;
+  auto uname = px::system::GetUname();
+  if (uname.ok()) {
+    const auto host_path = px::system::Config::GetInstance().ToHostPath(absl::Substitute("$0/$1/$2", kLinuxHeadersPath, uname.ConsumeValueOrDie(), "build"));
+
+    const auto resolved_host_path = px::system::ResolvePossibleSymlinkToHostPath(host_path);
+    kernel_headers_installed = resolved_host_path.ok();
+  }
+
```
ddelnano, line 83: This duplicates a portion of the work done by BCCWrapper. Another option would be to move that header installation here rather than checking for the presence of the headers. That wouldn't suffer from the test drawbacks outlined [here](https://github.com/pixie-io/pixie/blob/854062111cf4b91a40649a2e2647c88c0a68b0db/src/stirling/bpf_tools/bcc_wrapper.cc#L114-L117).

ddelnano:
@pixie-io/maintainers this is ready for review when you have the chance!

aimichelle (approved):

src/vizier/services/agent/pem/pem_manager.h
```diff
@@ -53,18 +53,19 @@ class PEMManager : public Manager {
  protected:
   PEMManager() = delete;
   PEMManager(sole::uuid agent_id, std::string_view pod_name, std::string_view host_ip,
-             std::string_view nats_url, px::system::KernelVersion kernel_version)
+             std::string_view nats_url, px::system::KernelVersion kernel_version,
+             bool kernel_headers_installed)
```
oazizi000, line 57: optional: Might be worth a `KernelInfo` struct that includes `KernelVersion kernel_version` and the `bool kernel_headers_installed`

src/vizier/services/agent/pem/pem_main.cc
```diff
@@ -41,6 +42,8 @@ using ::px::vizier::agent::DefaultDeathHandler;
 using ::px::vizier::agent::PEMManager;
 using ::px::vizier::agent::TerminationHandler;
 
+constexpr std::string_view kLinuxHeadersPath = "/lib/modules";
```
oazizi000, line 45: A little worried that there's an implicit relationship between this location and code we have in `pixie/src/stirling/utils/linux_headers.cc`. I guess one could argue that it's really BCC that's deciding to standardize on that location.

A comment about this being connected to BCC's search path might be warranted.

oazizi000 (approved):

src/vizier/services/agent/pem/pem_main.cc
ddelnano: Refactored this to a shared variable in fe6ead97447e2d5ef34eb1f73ffc72026b0f9a6a

src/vizier/services/agent/pem/pem_manager.h
ddelnano: Updated in 18853b052cb2be07eefbc1375e82b4d7683d9982

PR #2062, opened 2024-12-17
[cloud] Provide ability to disable executing modified pxl scripts

ddelnano:
Summary: [cloud] Provide ability to disable executing modified pxl scripts

Certain security conscious users are hesitant to use Pixie because without RBAC anyone with Pixie UI access can write arbitrary BPF code (bpftrace integration), access or export arbitrary data (modifying pxl scripts, writing export scripts). This change aims to address this concern with a global setting to prevent the ability to execute modified scripts. When an adhoc script is executed, the cloud will hash the contents of the script and check it against the scripts known to the scriptmgr service. If it is contained in the scriptmgr service, the script will be allowed to execute.

Note: this does not prevent users from writing new export scripts. Since the query broker can source its scripts from a configmap as of #1326, this is deemed as an appropriate mitigation for cluster admins and I'll follow up with UI support to reflect that a vizier is in "configmap mode".

Relevant Issues: N/A

Type of change: /kind feature

Test Plan: The following checks were performed
- [x] New tests verify the scriptmgr and api service changes work
- [x] Skaffold'ed to a testing cluster and verified script modification is blocked and unmodified scripts are allowed to run. In addition to this, the code editor in the UI is made read only and shows an explanation

<details><summary>Screenshots</summary>

Screen Shot 2025-01-07 at 8 58 34 AM
Screen Shot 2025-01-07 at 8 58 37 AM
Screen Shot 2025-01-07 at 8 58 59 AM
</details>

Changelog Message: Pixie Cloud can now disable executing modified pxl scripts via the `PL_SCRIPT_MODIFICATION_DISABLED` key in the `pl-script-bundle-config` ConfigMap. See reference manifests for more details.

### DIFF
--- a/k8s/cloud/base/proxy_nginx_config.yaml
+++ b/k8s/cloud/base/proxy_nginx_config.yaml
@@ -35,7 +35,7 @@ data:
     sub_filter '__CONFIG_DOMAIN_NAME__' "'${domain_name}'";
     sub_filter '__CONFIG_SCRIPT_BUNDLE_URLS__' "'${script_bundle_urls}'";
     sub_filter '__CONFIG_SCRIPT_BUNDLE_DEV__' "'${script_bundle_dev}'";
-    sub_filter '__CONFIG_SCRIPT_MODIFICATION_DISABLED__' "'${script_modification_disabled}'";
+    sub_filter '__CONFIG_SCRIPT_MODIFICATION_DISABLED__' "${script_modification_disabled}";
     sub_filter '__SEGMENT_UI_WRITE_KEY__' "'${segment_ui_write_key}'";
     sub_filter '__SEGMENT_ANALYTICS_JS_DOMAIN__' "'segment.${domain_name}'";
     sub_filter '__CONFIG_LD_CLIENT_ID__' "'${ld_client_id}'";
@@ -135,7 +135,7 @@ data:
     set_by_lua_block $segment_cli_write_key  { return os.getenv("PL_SEGMENT_CLI_WRITE_KEY") }
     set_by_lua_block $script_bundle_urls  { return os.getenv("SCRIPT_BUNDLE_URLS") }
     set_by_lua_block $script_bundle_dev  { return os.getenv("SCRIPT_BUNDLE_DEV") }
-    set_by_lua_block $script_modification_disabled  { return os.getenv("SCRIPT_MODIFICATION_DISABLED") }
+    set_by_lua_block $script_modification_disabled  { return os.getenv("PL_SCRIPT_MODIFICATION_DISABLED") }
     set_by_lua_block $analytics_enabled  { return os.getenv("ANALYTICS_ENABLED") }
     set_by_lua_block $announcement_enabled  { return os.getenv("ANNOUNCEMENT_ENABLED") }
     set_by_lua_block $announce_widget_url  { return os.getenv("ANNOUNCE_WIDGET_URL") }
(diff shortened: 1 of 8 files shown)


src/cloud/api/ptproxy/vizier_pt_proxy.go
```diff
@@ -20,32 +20,75 @@ package ptproxy
...
+
 // VizierPassThroughProxy implements the VizierAPI and allows proxying the data to the actual
 // vizier cluster.
 type VizierPassThroughProxy struct {
-	nc *nats.Conn
-	vc vzmgrClient
+	nc                       *nats.Conn
+	vc                       vzmgrClient
+	sm                       scriptmgrClient
+	disableScriptModifiation bool
```
aimichelle, line 58: Nit: typo

src/cloud/scriptmgr/scriptmgrpb/service.proto
```diff
@@ -36,6 +36,9 @@ service ScriptMgrService {
   rpc GetScripts(GetScriptsReq) returns (GetScriptsResp);
   // GetScriptContents returns the pxl string of the script.
   rpc GetScriptContents(GetScriptContentsReq) returns (GetScriptContentsResp);
+  // Checks to see if script with hash exists. This is used to check if a script
+  // has been modified or not.
+  rpc GetScriptByHash(GetScriptByHashReq) returns (GetScriptByHashResp);
```
aimichelle, line 41: The naming of this function makes me think that you'll be able to get the contents of the script given the hash, or atleast a script ID or something. You could make this more specific... Maybe `CheckScriptExists`?

aimichelle (commented):
For a slightly better experience, should we also update the UI so that the editor doesn't allow the user to make any modifications? That way they don't have to wait to edit the script, then try to run the script, in order to know it's not going to work.

src/ui/src/components/code-editor/code-editor.tsx
```diff
@@ -85,7 +86,7 @@ export class CodeEditor extends React.PureComponent<CodeEditorProps, CodeEditorS
         scrollBeyondLastColumn: 3, // Prevents hiding text behind the minimap or the scrollbar. Expands the scroll area.
         scrollBeyondLastLine: false,
         fontFamily: COMMON_THEME.typography.monospace.fontFamily,
-        readOnly: this.props.isReadOnly === true,
+        readOnly: this.props.isReadOnly === true || SCRIPT_MODIFICATION_DISABLED,
```
NickLanam, line 89: Notably, `CodeEditor` doesn't show the user _why_ it's read-only, and assumes the caller will tell them why (I don't remember if any existing caller does so, though - they might not).

Having `readOnly` set directly inside of `CodeEditor` this way breaks that assumption further, showing nothing and giving no indication that it should. Might be worth consideration in a future change!

NickLanam (approved):
UI adjustments LGTM, with the comment below.

k8s/cloud/base/proxy_nginx_config.yaml
```diff
@@ -169,7 +171,8 @@ data:
     env PL_HYDRA_SERVICE;
     env PL_KRATOS_SERVICE;
     env SCRIPT_BUNDLE_URLS;
-    env SCRIPT_BUNDE_DEV;
+    env SCRIPT_BUNDLE_DEV;
```
ddelnano, line 174: This is a typo but something that's only used fro development.

src/ui/src/components/code-editor/code-editor.tsx
ddelnano: I've changed this in favor of having each instance of the component to provide a read only reason. See testing done screenshots for more details.

src/cloud/scriptmgr/scriptmgrpb/service.proto
ddelnano: Updated this in e90686e10e5a7d9bef59995c0f119dad306c82a7

ddelnano:
@aimichelle please see the latest testing done screenshots. The UI now enables the read only setting for the `CodeEditor` component and shows a tool tip explaining why it's read only.

aimichelle (approved):

PR #2077, opened 2025-01-14
Allow clang and gcc builds for `src/e2e_test/vizier/planner:planner_test`

ddelnano:
Summary: Allow clang and gcc builds for `src/e2e_test/vizier/planner:planner_test`

As far as I can tell, there isn't a bazel config that matches the tags of the `src/e2e_test/vizier/planner:planner_test` target. This change removes the tag preventing clang builds since they appear to run without problems. In addition to this, a `vispb` proto change was missing from the earlier differential flamegraph work. This prevented the `planner_test` from running properly (see comment for more details).

Relevant Issues: N/A

Type of change: /kind bugfix

Test Plan: Verified gcc and clang builds work
- [x] `bazel test --config clang src/e2e_test/vizier/planner:planner_test` is successful
- [x] `bazel test -c opt --config gcc src/e2e_test/vizier/planner:planner_test` is successful


src/api/proto/vispb/vis.proto
```diff
@@ -458,4 +458,7 @@ message StackTraceFlameGraph {
   string node_column = 8;
   // The label for the percentage.
   string percentage_label = 9;
+  // The column containing the stacktrace's delta. This is only relevant for differential
+  // flamegraphs.
+  string difference_column = 10;
```
ddelnano, line 463: The recent differential flamegraph changes weren't made to this file. Without this change, the test target fails with the following error:

```
$ bazel test -c dbg src/e2e_test/vizier/planner:planner_test --test_output=all

INFO: From Testing //src/e2e_test/vizier/planner:planner_test:
==================== Test output for //src/e2e_test/vizier/planner:planner_test:
--- FAIL: TestAllScriptsCompile (0.04s)
    all_scripts_test.go:322:
                Error Trace:    src/e2e_test/vizier/planner/all_scripts_test.go:322
                Error:          Received unexpected error:
                                can't unmarshal Any nested proto *vispb.StackTraceFlameGraph: unknown field "differenceColumn" in vispb.StackTraceFlameGraph
                Test:           TestAllScriptsCompile
FAIL
```

The flamegraph script has been in a cloud release for a while now and I've used it extensively, so it appears that only the e2e tests here use this proto. Typescript does match on the [proto package name](https://github.com/pixie-io/pixie/blob/534b8926b22f6cbf3b531e5175383a0dc9e03928/src/ui/src/containers/live/convert-to-vega-spec/convert-to-vega-spec.ts#L246-L248), but it has its own [data structure](https://github.com/pixie-io/pixie/blob/534b8926b22f6cbf3b531e5175383a0dc9e03928/src/ui/src/containers/live/convert-to-vega-spec/flamegraph.ts#L43) for working with it.

NickLanam (approved):
Oof, good catch

src/e2e_test/vizier/planner/BUILD.bazel
```diff
@@ -24,8 +24,6 @@ pl_go_test(
     ],
     tags = [
         "no_asan",
-        "no_gcc",
-        "no_libcpp",
```
JamesMBartlett, line 28: Did you check if it works with libcpp?

JamesMBartlett (approved):
lgtm if it works with libcpp

src/e2e_test/vizier/planner/BUILD.bazel
ddelnano: `bazel build` (without specifying a `--config`) couldn't find this test target until `the no_libcpp` was removed. Therefore I thought using libcpp was the default, however, I've tested that `--config clang` results in a successful build as well.

Issue #666, opened 2022-12-10
Request for Feedback: Dynamic Logging's future (Should we deprecate Dynamic Logging?)


yzhao1012:
**Summary**

https://docs.pixielabs.ai/tutorials/custom-data/dynamic-go-logging/
Pixie Dynamic Logging let users write eBPF uprobes in PXL script, Pixie then deploys the uprobes onto user-specified Pods, and collect the data into Pixie's own data store.

https://docs.pixielabs.ai/tutorials/custom-data/distributed-bpftrace-deployment
Pixie also has Distributed Bpftrace Deployment, which lets users deploy kprobes and uprobes on Kubernetes cluster.

Given the limited capability of the PXL language, Dynamic Logging is used much less than Distributed Bpftrace Deployment. We are looking for feedbacks from the community regarding the future directions of Dynamic Logging.

**Suggestions**

1. One option is to simply remove Dynamic Logging from Pixie. That greatly reduces our maintenance burden.
2. Another option is to further expand PXL language support for Dynamic Logging to make it more powerful and useful.
3. A hybrid option is to extra the useful part of Dynamic Logging and contributed to BPFtrace, for example, the part of navigating struct variables with DWARF information (already part of BPFTrace https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#4-uprobeuretprobe-dynamic-tracing-user-level-arguments) could get improved with Pixie's current implementation.
4. Any other suggestions?

oazizi000:
Given the lack of feedback, we are now proposing to deprecate the Dynamic Logging feature.

Rationale:
- The feature is alpha and does not seem to be popular (given the lack of github issues related to it).
- It is only supported on Go 1.16, which has not been supported for 1 year.

If anyone believes that Dynamic Logging should be maintained, we would need to identify someone to further develop and maintain it. If there is not enough support for keeping Dynamic Logging over the next 30 days, we intend to start the deprecation process.

mayanksingh2298:
I think it's an interesting feature to be able to probe production deployments in user space. I see possible use-cases around live debugging of variables. As far as I remember Delve recently introduced ebpf backend. So I find it a little hard to believe that there's not enough need for it.

I think expanding PXL language support makes much more sense. (Maybe using the distributed bpftrace deployment).

Anyways, I'll be happy to contribute to this. Is there some more documentation around using pxtrace - something more than https://docs.px.dev/reference/pxl/tracepoint-decorator/pxtrace.probe/

[...]

PR #2093, opened 2025-01-21
Fix `ExtractGolangInterface` to handle go 1.20 binaries and later

ddelnano:
Summary: Fix `ExtractGolangInterface` to handle go 1.20 binaries and later

This is part of the work to deprecate or fix dynamic logging

Relevant Issues: #666

Type of change: /kind bugfix

Test Plan: New tests pass


src/stirling/obj_tools/go_syms.cc
```diff
@@ -148,13 +148,20 @@ StatusOr<absl::flat_hash_map<std::string, std::vector<IntfImplTypeInfo>>> Extrac
...
   PX_ASSIGN_OR_RETURN(std::vector<ElfReader::SymbolInfo> itable_symbols,
                       elf_reader->SearchSymbols(kITablePrefix, SymbolMatchType::kPrefix,
                                                 /*symbol_type*/ ELFIO::STT_OBJECT));
-
+  if (itable_symbols.empty()) {
+    kITablePrefix = "go.itab.";
+    PX_ASSIGN_OR_RETURN(itable_symbols,
+                        elf_reader->SearchSymbols(kITablePrefix, SymbolMatchType::kPrefix,
+                                                  /*symbol_type*/ ELFIO::STT_OBJECT));
+  }
```
oazizi000, line 164: Can we put the candidates in a list, apply the same logic in a loop, and then break on the first success? Reduces the copied code.

ddelnano: Updated in f3e1b76

oazizi000 (approved):

PR #2094, opened 2025-01-21
Make metadata pod lookups more resilient to short lived processes

ddelnano:
Summary: Make metadata pod lookups more resilient to short lived processes

This is a continuation of the work started from #1989. Since the `local_addr` column is populated for client side traces, it can be used as a fallback lookup for these traces. This doesn't solve all of the permutations of  missing short lived processes (#1638), but provides more coverage than before.

Relevant Issues: #1638

Type of change: /kind bugfix

Test Plan: Verified the following
- [x] Compared the performance with and without this change with `src/e2e_test/vizier/exectime:exectime`. This change has a minor performance impact, but it closes the gap on certain situations that previously caused users to distrust Pixie's instrumentation
```
# Performance baseline
$ ./exectime benchmark -a testing.getcosmic.ai:443 -c <cluster_id> 2>&1 | tee baseline_for_simple_udf_swap_e20880ffd.txt
# Performance of this change
./exectime benchmark -a testing.getcosmic.ai:443 -c <cluster_id> 2>&1 | tee simple_udf_swap_cd217c05c.txt
```
[simple_udf_swap_cd217c05c.txt](https://github.com/user-attachments/files/18497709/simple_udf_swap_cd217c05c.txt)
[baseline_for_simple_udf_swap_e20880ffd.txt](https://github.com/user-attachments/files/18497710/baseline_for_simple_udf_swap_e20880ffd.txt)
- [x] Ran `for i in $(seq 0 1000); do curl http://google.com/$i; sleep 2; done` within a pod and verified that with this change all traces are shown, without this change a significant number of traces are missed. See before and after screenshots below:

vizier-0 14 14-curl-with-missing-data
traces-with-new-fallback

Changelog Message: Fix a certain class of cases where Pixie previously missed protocol traces from short lived connections

### DIFF
--- a/src/carnot/planner/compiler/analyzer/convert_metadata_rule.cc
+++ b/src/carnot/planner/compiler/analyzer/convert_metadata_rule.cc
@@ -70,10 +70,27 @@ StatusOr<std::string> ConvertMetadataRule::FindKeyColumn(std::shared_ptr<TableTy
       absl::StrJoin(parent_type->ColumnNames(), ","));
 }
 
-bool CheckBackupConversionAvailable(std::shared_ptr<TableType> parent_type,
-                                    const std::string& func_name) {
-  return parent_type->HasColumn("time_") && parent_type->HasColumn("local_addr") &&
-         func_name == "upid_to_pod_name";
+StatusOr<FuncIR*> AddUPIDToPodNameFallback(std::shared_ptr<TableType> parent_type, IRNode* ir_node,
+                                           IR* graph) {
+  if (!parent_type->HasColumn("local_addr") || !parent_type->HasColumn("time_")) {
+    return error::NotFound("Parent type does not have required columns for fallback conversion.");
+  }
+  PX_ASSIGN_OR_RETURN(auto upid_column, graph->CreateNode<ColumnIR>(ir_node->ast(), "upid", 0));
+  PX_ASSIGN_OR_RETURN(auto local_addr_column,
+                      graph->CreateNode<ColumnIR>(ir_node->ast(), "local_addr", 0));
+  PX_ASSIGN_OR_RETURN(auto time_column, graph->CreateNode<ColumnIR>(ir_node->ast(), "time_", 0));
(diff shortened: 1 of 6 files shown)


src/carnot/planner/compiler/analyzer/convert_metadata_rule.cc
```diff
@@ -85,17 +91,38 @@ StatusOr<bool> ConvertMetadataRule::Apply(IRNode* ir_node) {
...
-                                std::vector<ExpressionIR*>{key_column}));
+  auto backup_conversion_available = CheckBackupConversionAvailable(resolved_table_type, func_name);
+
+  FuncIR* conversion_func;
+
+  // TODO(ddelnano): Until the short lived process issue (gh#1638) is resolved, use a
+  // conversion function that uses local_addr for pod lookups when the upid based default
+  // (upid_to_pod_name) fails.
+  if (backup_conversion_available) {
+    func_name = "_upid_to_podname_local_addr_fallback";
```
aimichelle, line 110: Nit: Feels a little weird to put something so specific to the local_addr stuff in here.
To make it a little cleaner, can we pull this into a `AddBackupConversions` function and put this logic into there instead?

ddelnano: I took a stab at this in 6f8ee0e41afc2e32eff246a6f1341022e9a72100.

If we were to continue to expand on this pattern, I think we should update the MetadataProperty class to work with a composite lookup key (`time_`, `local_addr` and `upid` => pod name as well as `upid` => pod name). That would allow this rule's logic to stay the same and keep the metadata lookup mappings all in the [metadata_handler.cc file](https://github.com/pixie-io/pixie/blob/1f6d18bfba7bd3cedf99e3d7eec5245b0e018f21/src/carnot/planner/metadata/metadata_handler.cc#L57-L58).

I did investigate what I described above, but I opted for this approach since that "composite lookup" hasn't seemed like a more common pattern at this time.

aimichelle (approved):

PR #2116, opened 2025-02-10
Add go 1.22 and 1.23 to socket tracer tests. Bump minor go versions

ddelnano:
Summary: Add go 1.22 and 1.23 to socket tracer tests. Bump minor go versions

This PR adds Go 1.22 and 1.23 to the socket tracer test suite in anticipation of removing Go 1.16 and 1.17 from our repo once #2108 is merged. The plan is to follow this up with a change that removes Go 1.16 and 1.17 in addition to upgrading a portion of our Go dependencies. The bulk of our dependencies still support Go 1.18, so this will allow us to upgrade to the latest versions.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Test suite passes

PR #2120, opened 2025-02-13
Remove go 1.16 and 1.17. Add Go 1.22 and 1.23 into dynamic tracer test matrix

ddelnano:
Summary: Remove go 1.16 and 1.17. Add Go 1.22 and 1.23 into dynamic tracer test matrix

Once Go 1.16 and 1.17 are removed from the repo, we can pursue go dependency upgrades like #2115. I also added Go 1.22 and 1.23 (the latest versions) to dynamic tracer tests that ideally should cover multiple go versions.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Existing tests


src/stirling/source_connectors/socket_tracer/BUILD.bazel
```diff
@@ -546,7 +540,7 @@ pl_cc_bpf_test(
 
 pl_cc_bpf_test(
     name = "go_tls_trace_bpf_test",
-    timeout = "moderate",
+    timeout = "long",
```
ddelnano, line 543: While running builds against #2115, I noticed that this test is flaky after merging #2116. This is because we added 2 go versions and only removed one ([diff](https://github.com/pixie-io/pixie/pull/2116/files#diff-4c54c2ecc1c0bf68208d05a35c85cafc82eec10c800f0d7e0c60397cbaf665cb)). Increasing the timeout should solve that problem.

vihangm: Given that `go1.24` was recently released and the golang policy is to old support upto two versions too old, I think. you can drop `1.18`. I'd probably also suggest dropping `1.19` too.
IMO increasing the test matrix too much doesn't seem worth it.

vihangm (approved):

src/stirling/source_connectors/socket_tracer/BUILD.bazel
ddelnano: Yea, I agree that we should deprecate those soon.

Omid had concerns that there might be older Go versions in use by end users despite Go's policy. I know of a few end users in this situation, so I might consider how to keep that test coverage without having our own dependencies tied to it.

If I decide to abandon keeping that coverage, at least its much easier to drop old support now with the dynamic logging stuff figured out. So a subsequent removal will be much easier.

ddelnano:
As for the @pixie-io/stirling review, Omid and I discussed this change and was aligned with removing these Go versions (https://github.com/pixie-io/pixie/pull/2116#pullrequestreview-2609679100). So I'm going to move forward with this change.

PR #2180, opened 2025-04-21
Fixup codeql runs

vihangm:
Summary: There were two issues on main related to codeql. This attempts to address both of them.
1) since go 1.21, the go.mod file is expected to have a version of the form v1.X.Y instead of v1.X codeql warns about this, so this fixes the same.
2) given the recent change in GH action runners, the runners don't seem to have nodejs installed by default, this ensures that nodejs is indeed installed.

Relevant Issues: N/A

Type of change: /kind bug

Test Plan: Tested e2e.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -27,21 +27,34 @@ jobs:
     - uses: github/codeql-action/analyze@1b549b9259bda1cb5ddde3b41741a82a2d15a841  # v3.28.13
       with:
         category: "/language:go"
-  analyze:
+  analyze-python:
     runs-on: oracle-8cpu-32gb-x86-64
     permissions:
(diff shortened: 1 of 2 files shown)


koksay:
I was preparing a PR for that, thanks for taking care of it!

.github/workflows/codeql.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
     - main
+    - vihang/codeql
```
ddelnano, line 7: You probably are already on this, but just a reminder to clean this up before merging.

ddelnano (approved):

JamesMBartlett (approved):

JamesMBartlett (approved):

PR #2191, opened 2025-05-03
Adding the Build-VM setup to Development.md

entlein:
On our GCP setup, we tested three times how to configure a VM as remote workstation to run Pixie on minikube using our registry. We wrote down the commands and hope others may find them useful.
Please review if this is in line with your documentation style.

I m opening the PR now, to see if the git commit signatures (if squashed) will be accepted


DEVELOPMENT.md
```diff
@@ -10,6 +10,104 @@ This document outlines the process for setting up the development environment fo
...
+make dev-env-start
+```
+Deploy vanilla pixie (remote cloud)
+```sh
+sudo bash -c "$(curl -fsSL https://getcosmic.ai/install.sh)"
+export PX_CLOUD_ADDR=getcosmic.ai
+px auth
+px deploy -p=1Gi
+```
+For reference and further information https://docs.px.dev/installing-pixie/install-guides/hosted-pixie/cosmic-cloud
```
ddelnano, line 87: Can we link to the documentation at https://docs.px.dev/installing-pixie/install-schemes/cli/ instead? Since our docs website is in a [separate repository](https://github.com/pixie-io/docs.px.dev), I worry that this could be missed if/when we make changes to our documentation.

In my opinion the important part to capture is that it requires deploying the Vizier first before a custom build can be deployed. I was thinking something along these lines:

> Before a custom build of Pixie can be deployed, you must install Pixie to your Kubernetes cluster first. Follow the steps [here](https://docs.px.dev/installing-pixie/install-schemes/cli/) and ensure Pixie is fully working before proceeding.

DEVELOPMENT.md
```diff
@@ -10,6 +10,104 @@ This document outlines the process for setting up the development environment fo
...
+sudo chmod 2775 <directory-path>
+```
+
+2) Create/Use a registry you control and login
+   
+```sh
+docker login ghcr.io/<myregistry>
+```
+
+3) Make Minikube run and deploy a vanilla pixie
```
ddelnano, line 74:
```suggestion
3) Prepare your Kubernetes cluster and deploy an existing Pixie release
```

Could we adjust the wording to emphasize that development can be done against any Kubernetes cluster? That way, users with an existing cluster won’t feel the need to set up an additional one.

Maybe something like this:

> If you have an existing kubernetes cluster to develop against, ensure your kubectl config is set properly. Otherwise, you can create a minikube cluster .... (existing instructions here).

DEVELOPMENT.md
```diff
@@ -10,6 +10,104 @@ This document outlines the process for setting up the development environment fo
...
+Uses a Ubuntu 24.04 as base to run `chef` to setup all dependencies.
+The initial compilation is CPU intense and `16vcpu` are recommended.
+This was tested on GCP: a balanced disk of 500 GB and a VM type that supports nested virtualization should be chosen, as of writing (May 2025)
+`n2-standard-16` works well. 
+
+> [!Warning]
+>  The first build takes several hours and at least 160 Gb of space
+
+Turn on nested virtualization and avoid the use of `spot` VMs for the first build to avoid the very long first
+build interrupting. If you create the VMs as templates from an image, you can later switch to more cost-effective `spot` instances.
```
ddelnano, line 25: Can we specify that nested virtualization is only needed if you intend to develop against Minikube?

Also I think the details about spot instances is a little niche as it's possible that people are running non cloud VMs. Could we remove those details?

skaffold/skaffold_vizier.yaml
```diff
@@ -144,3 +144,4 @@ profiles:
     path: /build/artifacts/context=./bazel/args
     value:
     - --config=x86_64_sysroot
+    - --compilation_mode=opt
```
ddelnano, line 147: I'm not sure we want to change this default. During development it's helpful to have debug checks trigger or to generate core dumps with usable stack traces (symbolized). For the file source changes you are working with, I think using `opt` is sensible, but most of the time its a case by case basis.

What if we added a brief note in 4) on  the `--compilation_mode` setting? We could add something like this:

> Note: You will want to stick with a sysroot based build (`-p x86_64_sysroot` or `-p aarch64_sysroot`), but you may want to change the `--complication_mode` setting based on your needs. `opt` builds remove assert/debug checks, while `dbg` builds work with debuggers (gdb). See the [bazel docs](https://bazel.build/docs/user-manual#compilation-mode) for more details.

I do think we might want to consider making the `x86_64_sysroot` profile the default (without any compilation_mode changes). Running a host build (VM glibc version) on a Kubernetes host usually results in cryptic glibc crashes. This occurs when the host's/VM's glibc version is newer than what the Kubernetes host supports.

ddelnano (commented):
Thanks for putting this together @entlein! Setting up our development environment has always come with a steep learning curve due to some undocumented aspects, but these updates will definitely make the process much smoother!

DEVELOPMENT.md
```diff
@@ -10,6 +10,104 @@ This document outlines the process for setting up the development environment fo
...
+
+```yaml
+advancedMachineFeatures:
+  enableNestedVirtualization: true
+```
+
+1) Install chef and some dependencies
+ 
+```
+sudo apt update sudo apt install -y git coreutils mkcert libnss3-tools screen libvirt-daemon-system libvirt-clients qemu-kvm virt-manager
```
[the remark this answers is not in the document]
entlein: I m moving that into 'chef', but we need to retest, so left a comment for now so ppl can install it manually. Most of the dependencies were for minikube, mkcert and coreutils were for running the vanilla pixie install (in a later step).

You re absolutely correct about screen: I separated out the tmux/screen install, as it is indeed a personal preference, but annoying if you forget it (IMO).

tools/chef/cookbooks/px_dev/recipes/linux.rb
```diff
@@ -56,6 +56,17 @@
...
+  # Minikube dependencies for kvm
+  'libnss3-tools',
+  'libvirt-daemon-system',
+  'libvirt-clients',
+  'qemu-kvm', 
+  'virt-manager',
+
+  # Pixie dependencies
+  'mkcert',
+  #'coreutils' not sure about that one, need to test
```
ddelnano, line 69:
```suggestion
```

This line looks like it was a remnant from an earlier change.

DEVELOPMENT.md
```diff
@@ -10,6 +10,133 @@ This document outlines the process for setting up the development environment fo
...
+```
+
+4) Make Minikube run and deploy a vanilla pixie
+
+If you added your user to the libvirt group (`sudo usermod -aG libvirt $USER`), starting the development environment on this VM will now work (if you did this interactively: you need to refresh your group membership, e.g. by logout/login). The following command will, amongst other things, start minikube
+```sh
+make dev-env-start
+```
+
+Onto this minikube, we first deploy the upstream pixie (`vizier`, `kelvin` and `pem`) using the remote cloud  `export PX_CLOUD_ADDR=getcosmic.ai` . Follow https://docs.px.dev/installing-pixie/install-schemes/cli , to install the  `px` command line interface and login:
```
ddelnano, line 100:
```suggestion
Onto this minikube, we first deploy the upstream pixie (`vizier`, `kelvin` and `pem`) using the hosted cloud. Follow [these instructions](https://docs.px.dev/installing-pixie/install-schemes/cli) to install the `px` command line interface and Pixie:
```

We refer to the Cosmic and New Relic clouds as "hosted". This suggestion keeps the terminology consistent with the rest of our docs. I also wanted to avoid any reference to any specific cloud. Our docs are supposed to be 'vendor neutral' and the linked documentation is set up according to the CNCF guidelines.

skaffold/skaffold_vizier.yaml
ddelnano: @entlein did you see my earlier comment above?

I'm happy to leave the instructions as is and not expound on `--compilation_mode`, but I do think this file should be left unedited as the `dbg` default is usually appropriate.

DEVELOPMENT.md
```diff
@@ -10,6 +10,133 @@ This document outlines the process for setting up the development environment fo
...
 ## Setting up the Environment
 
+Decide first if you'd like a full buildsystem (on a VM) or a containerized dev environment.
+
+### VM as buildsystem
+
+This utilizes `chef` to setup all dependencies and is based on `ubuntu`. The VM type must support nested virtualization for `minikube` to work.
+
+
+The following specifics were tested on GCP on a Ubuntu 24.04 (May 2025): The initial compilation is CPU intense and `16vcpu` were a good trade-off, a balanced disk of 500 GB seems convienent and overall `n2-standard-16` works well.
```
ddelnano, line 20:
```suggestion
The following specifics were tested on GCP on a Ubuntu 24.04 (May 2025). Please see the latest [packer file](https://github.com/pixie-io/pixie/blob/main/tools/chef/Makefile#L56) for the current supported Ubuntu version: The initial compilation is CPU intense and `16vcpu` were a good trade-off, a balanced disk of 500 GB seems convenient and overall `n2-standard-16` works well. 
```

People often ask what OS to build their development environment on, so I added a slight modification to link to the current Ubuntu base image. Also fixed a misspelling.

DEVELOPMENT.md
```diff
@@ -10,6 +10,133 @@ This document outlines the process for setting up the development environment fo
...
+px auth login
+```
+
+Once logged in to pixie, we found that limiting the memory is useful, thus after login, set the deploy option like so:
+```sh
+px deploy -p=1Gi
+```
+For reference and further information https://docs.px.dev/installing-pixie/install-guides/hosted-pixie/cosmic-cloud.
+
+You may encounter the following WARNING, which is related to the kernel headers missing on the minikube node (this is not your VM node). Usually, for development purposes this is safe to ignore. Please see [pixie-issue2051](https://github.com/pixie-io/pixie/issues/2051) for further details.
```
ddelnano, line 111:
```suggestion
You may encounter the following WARNING, which is related to the kernel headers missing on the minikube node (this is not your VM node). This is safe to ignore if Pixie starts up properly and your cluster is queryable from Pixie's [Live UI](https://docs.px.dev/using-pixie/using-live-ui). Please see [pixie-issue2051](https://github.com/pixie-io/pixie/issues/2051) for further details.
```

The kernel header issue can affect development and "real" clusters. This suggestion modifies the wording slightly to provide a hint at what to look for to give people confidence that their cluster is working properly.

ddelnano (changes requested):
Thanks again for the work on this! I have another round of minor comments, but this is shaping up nicely.

Also please see the errors on the [lint-pr-description build](https://github.com/pixie-io/pixie/actions/runs/14886988962/job/41818731172). The PR description needs to match our [PR template](https://github.com/pixie-io/pixie/blob/main/.github/PULL_REQUEST_TEMPLATE/blank_template.md).

DEVELOPMENT.md
entlein: Very good point, yes. will add. 

From a maintainer PoV: wouldnt you want ppl to use a golden path, though? To limit the range of potential issues?

entlein: I added the destinction in two WARNING boxes

skaffold/skaffold_vizier.yaml
entlein: Totally missed this comment yesterday. Addressed it now by adding exactly as you suggested and the option commented out so people have the correct syntax available

DEVELOPMENT.md
entlein: Suggestion was applied

DEVELOPMENT.md
entlein: This should now be addressed and vendor neutral

DEVELOPMENT.md
entlein: I removed and added explanation

DEVELOPMENT.md
```diff
@@ -10,6 +10,166 @@ This document outlines the process for setting up the development environment fo
...
+Once you make changes to the source code, or switch to another source code version, use Skaffold to deploy (after you have the vanilla setup working on minikube)
+
+Ensure that you have commented in the bazelcache-directory into the bazel config (see Step 2).
+
+
+Review the compilation-mode suits your purposes:
+```
+cat skaffold/skaffold_vizier.yaml
+# Note: You will want to stick with a sysroot based build (-p x86_64_sysroot or -p aarch64_sysroot),
+# but you may want to change the --complication_mode setting based on your needs.
```
ddelnano, line 138:
```suggestion
# but you may want to change the --compilation_mode setting based on your needs.
```

skaffold/skaffold_vizier.yaml
```diff
@@ -138,9 +138,15 @@ profiles:
...
+# opt builds remove assert/debug checks, while dbg builds work with debuggers (gdb).
+# See the bazel docs for more details https://bazel.build/docs/user-manual#compilation-mode
 - name: x86_64_sysroot
   patches:
   - op: add
     path: /build/artifacts/context=./bazel/args
     value:
     - --config=x86_64_sysroot
+    - --compilation_mode=dbg
+#    - --compilation_mode=opt
```
ddelnano, line 152: Since this is already in the DEVELOPMENT.md, I think it would best to omit any changes to this file. The comments about compilation_mode go for any skaffold file in the repo (of which there are a few), so I think the DEVELOPMENT.md location covers our bases.

ddelnano (commented):
Two more minor comments and then this is good to go.

skaffold/skaffold_vizier.yaml
entlein: reverted to main

ddelnano (approved):

ddelnano (approved):

PR #2194, opened 2025-05-06
Add PEM command line flag for enforcing a max file size for ElfReader

ddelnano:
Summary: Add PEM command line flag for enforcing a max file size for ElfReader

While memory profiling the PEM, I noticed that large binaries caused the PEM's ElfReader code to make > 100 MiB allocations. This PR adds a cli flag (`--elf_reader_max_file_size`) that allows memory conscious users to prevent the PEM from parsing binaries that would cause these large memory allocations.

Relevant Issues: N/A

Type of change: /kind feature

Test Plan: Existing tests and skaffold

Changelog Message: Added `--elf_reader_max_file_size` flag to PEM. This flag allows memory conscious users to prevent the PEM from parsing large binaries in order to avoid large memory allocations.

### DIFF
--- a/src/stirling/obj_tools/elf_reader.cc
+++ b/src/stirling/obj_tools/elf_reader.cc
@@ -200,15 +200,11 @@ Status ElfReader::LocateDebugSymbols(const std::filesystem::path& debug_file_dir
 StatusOr<std::unique_ptr<ElfReader>> ElfReader::Create(
     const std::string& binary_path, const std::filesystem::path& debug_file_dir) {
   if (FLAGS_elf_reader_max_file_size != 0) {
-    PX_ASSIGN_OR_RETURN(const auto stat, fs::Stat(binary_path));
-    int64_t file_size = stat.st_size;
+    PX_ASSIGN_OR_RETURN(int64_t file_size, fs::GetFileSize(binary_path));
     if (file_size > FLAGS_elf_reader_max_file_size) {
       return error::Internal(
           "File size $0 exceeds ElfReader's max file size $1. Refusing to process file", file_size,
           FLAGS_elf_reader_max_file_size);
-    } else if (file_size < 0) {
-      return error::Internal("stat() returned negative file size $0 for file $1", file_size,
-                             binary_path);
     }
   }
   return CreateImpl(binary_path, debug_file_dir);
--- a/src/stirling/obj_tools/elf_reader.cc
+++ b/src/stirling/obj_tools/elf_reader.cc
@@ -32,6 +32,10 @@
 #include "src/common/fs/fs_wrapper.h"
 #include "src/stirling/obj_tools/init.h"
 
+DEFINE_int64(elf_reader_max_file_size, 0,
+             "Maximum file size in bytes for ELF files. Default value of 0 means all files will "
+             "be parsed regardless of size.");
+
 namespace px {
 namespace stirling {
 namespace obj_tools {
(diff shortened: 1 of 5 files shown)


src/stirling/obj_tools/elf_reader.h
```diff
@@ -50,8 +52,14 @@ class ElfReader {
...
-      const std::string& binary_path,
-      const std::filesystem::path& debug_file_dir = "/usr/lib/debug");
+      const std::string& binary_path, const std::filesystem::path& debug_file_dir = kDebugFileDir);
+
+  /**
+   * Creates an ElfReader that does not enforce the max file size limit. This is useful for cases
+   * where the binary size is known in advance or the binary must be loaded regardless of size.
+   */
+  static StatusOr<std::unique_ptr<ElfReader>> CreateUncapped(
+      const std::string& binary_path, const std::filesystem::path& debug_file_dir = kDebugFileDir);
```
oazizi000, line 62: From an API perspective, it might be simpler to have a single Create() where the size is an optional parameter that defaults to FLAGS_elf_reader_max_file_size. What do you think?

src/stirling/obj_tools/elf_reader.cc
```diff
@@ -158,40 +197,26 @@ Status ElfReader::LocateDebugSymbols(const std::filesystem::path& debug_file_dir
...
+  if (FLAGS_elf_reader_max_file_size != 0) {
+    PX_ASSIGN_OR_RETURN(const auto stat, fs::Stat(binary_path));
+    int64_t file_size = stat.st_size;
+    if (file_size > FLAGS_elf_reader_max_file_size) {
+      return error::Internal(
+          "File size $0 exceeds ElfReader's max file size $1. Refusing to process file", file_size,
+          FLAGS_elf_reader_max_file_size);
+    } else if (file_size < 0) {
+      return error::Internal("stat() returned negative file size $0 for file $1", file_size,
+                             binary_path);
```
oazizi000, line 211: I'd put much of this in a helper function StatusOr<int> GetFileSize(), which can return an error in the case of file_size < 0. Then you can do:
```
 PX_ASSIGN_OR_RETURN(int64_t file_size, GetFileSize(binary_path));
 if (file_size > FLAGS_elf_reader_max_file_size) {
    return error::Internal(
          "File size $0 exceeds ElfReader's max file size $1. Refusing to process file", file_size,
          FLAGS_elf_reader_max_file_size);
  }
```

which is a bit more focused.

ddelnano: Makes sense. I've refactored this and relocated to the fs_wrapper code in 9d48a38d7cb3202542432964ce39f5b8517b0def.

src/stirling/obj_tools/elf_reader.h
ddelnano: I'm open to changing this, but I wanted to make the uncapped behavior an explicit opt-in for consumers of ElfReader. Since it's used in many places, I could see a future bug where the max file size check is unintentionally bypassed when most consumers should retain the restriction.

oazizi000: We talked about this offline, but the suggestion is that the optional parameter defaults to the capped version, so I don't think it exposes any risk. The goal is just to keep the API simpler. But this is a bit subjective, so doesn't need to be a blocker if you prefer the two separate API calls.

oazizi000 (approved):

PR #2207, opened 2025-06-04
Add GoOffsetLocator that fulfills DwarfReader API. Use as shim for Go Uprobes

ddelnano:
Summary: Add GoOffsetLocator that fulfills DwarfReader APIs. Use as shim for Go Uprobes

Our current Go uprobe implementation relies on parsing DWARF information. This is memory intensive and is not ideal if end users want to keep the PEM's memory usage low (results in 100-150MB memory spikes).

In order to support lower memory uprobes and to expand our Go TLS support for binaries without DWARF, we can introduce a shim for the existing DwarfReader API that can read the offsets from a file (via openteletrmy-go-instrumentation generated offsets) or through our existing DwarfReader interface.

The next set of changes will use the `offset_results.json` file from https://github.com/pixie-io/opentelemetry-go-instrumentation/pull/1.

Relevant Issues: N/A

Type of change: /kind feature

Test Plan: Existing Go bpf trace tests should pass

### DIFF
--- a/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
+++ b/src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
@@ -174,10 +174,10 @@ Status UProbeManager::UpdateOpenSSLSymAddrs(obj_tools::RawFptrManager* fptr_mana
   return Status::OK();
 }
 
-Status UProbeManager::UpdateGoCommonSymAddrs(ElfReader* elf_reader, DwarfReader* dwarf_reader,
+Status UProbeManager::UpdateGoCommonSymAddrs(ElfReader* elf_reader, GoOffsetLocator* offset_locator,
                                              const std::vector<int32_t>& pids) {
   PX_ASSIGN_OR_RETURN(struct go_common_symaddrs_t symaddrs,
-                      GoCommonSymAddrs(elf_reader, dwarf_reader));
+                      GoCommonSymAddrs(elf_reader, offset_locator));
 
   for (auto& pid : pids) {
     PX_RETURN_IF_ERROR(go_common_symaddrs_map_->SetValue(pid, symaddrs));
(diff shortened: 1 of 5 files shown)


src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
```diff
@@ -886,8 +886,28 @@ int UProbeManager::DeployGoUProbes(const absl::flat_hash_set<md::UPID>& pids) {
...
+    if (build_info_s.ok()) {
+      auto& build_info_pair = build_info_s.ValueOrDie();
+      go_version = build_info_pair.first;
+      build_info = std::move(build_info_pair.second);
+    } else {
+      VLOG(1) << absl::Substitute(
+          "Failed to read build info from binary $0. Likely a Go binary built with an older "
+          "version of Go without module support (pre 1.12)."
+          "Message = $1",
+          binary, build_info_s.status().msg());
```
oazizi000, line 902: Should we be explicitly skipping/returning in this case?

src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
```diff
@@ -886,8 +886,28 @@ int UProbeManager::DeployGoUProbes(const absl::flat_hash_set<md::UPID>& pids) {
...
+    }
+
     std::unique_ptr<DwarfReader> dwarf_reader = dwarf_reader_status.ConsumeValueOrDie();
-    Status s = UpdateGoCommonSymAddrs(elf_reader.get(), dwarf_reader.get(), pid_vec);
+    StructOffsetMap struct_offsets;
+    FunctionArgMap function_offsets;
```
oazizi000, line 907: Comment that these are for future use.

oazizi000 (approved):

src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
ddelnano: Updated in f68e5f158.

src/stirling/source_connectors/socket_tracer/uprobe_manager.cc
ddelnano: I updated this in https://github.com/pixie-io/pixie/commit/f68e5f158450d3b17627a68b771193dd711c0c92. However, this breaks a test case that traces a Go <= 1.11 binary and why I filed #2210.

My plan is to fix that issue and then we can rebase this change once it's done.

ddelnano: https://github.com/pixie-io/pixie/pull/2225 will address this.

PR #2261, opened 2025-08-25
update oracle runners with vm ones

koksay:
Summary: Updating container-based GitHub action runners to the VM-based ones.

Type of change: /kind chore

Test Plan: Updated GH actions should not fail due to the runners.


ddelnano:
This work will be picked up in #2277. I got the build to pass as of today, so this should be merged soon! Thanks for all the help @koksay!

PR #2311, opened 2026-01-07
Consolidate `all_scripts_test.go` to use a single CGO target to fix duplicate symbol errors introduced by rules_go v0.58.1

ddelnano:
Summary: Consolidate `all_scripts_test.go` to use a single CGO target to fix duplicate symbol errors introduced by rules_go v0.58.1

https://github.com/bazel-contrib/rules_go/pull/4438, included in `rules_go` v0.58.1, causes certain statically linked CGO binaries to fail with duplicate symbol errors. This occurs when a binary depends on more than one CGO library that transitively depends on a common set of object files.

`all_scripts_test.go` previously depended on two CGO targets:
- `//src/carnot/planner`
- `//src/e2e_test/vizier/planner/dump_schemas/godumpschemas`

This PR solves this issue by removing the src/e2e_test/vizier/planner/dump_schemas/godumpschemas CGO library and instead generate the protobuf export directly in C++, loading it in the main application. This approach mirrors the existing pattern used in [src/vizier/funcs](https://github.com/pixie-io/pixie/blob/a6349a90b1e4b30f0bb13872ad03dff83a53f363/src/vizier/funcs/BUILD.bazel#L50-L66).

**Why not fix `rules_go`?**

The `rules_go` change that causes the issue explains that it doesn't include the necessary deduplication logic to avoid these duplicate symbol errors (https://github.com/bazel-contrib/rules_go/pull/4438#discussion_r2319892905). This tradeoff was [deemed acceptable](https://github.com/bazel-contrib/rules_go/pull/4438#issuecomment-3258120536) since it solved the c++ initialization problem with minimal complexity.

Relevant Issues: N/A

Type of change: /kind cleanup

Test Plan: Build should pass


src/e2e_test/vizier/planner/dump_schemas/BUILD.bazel
```diff
@@ -14,20 +14,23 @@
...
-    name = "dump_schemas",
-    srcs = [
-        "dump_schemas.cc",
-        "dump_schemas.h",
-    ],
-    hdrs = ["dump_schemas.h"],
+pl_cc_binary(
+    name = "export_schemas",
+    srcs = ["export_schemas.cc"],
+    stamp = -1,
```
vihangm: why do we need to disable stamping here?

vihangm (approved):

PR #2401, opened 2026-08-10
feat(pem): add authenticated direct-query gRPC endpoint

entlein:
Exposes VizierService.ExecuteScript directly on the normal PEM over a dedicated port (:50305 default), so node-local clients can query Carnot without a broker hop or cloud dependency.

The feature is opt-in (--direct_query_enabled=false by default) so existing PEM deployments are byte-for-byte unchanged until opted in. A compile-time kill switch (--//src/vizier/services/agent/pem:direct_query=disabled) removes the feature entirely from the binary.

Changes:

direct_query_server.{h,cc} (new)
- DirectQueryServer implements api::vizierpb::VizierService::ExecuteScript against the live PEM Carnot (reusing its table_store + metadata callback).
- AuthenticateRequest: HS256 JWT verifier via BoringSSL HMAC + rapidjson. Requires aud=vizier, iss=PL, Scopes=service, valid exp. Defends against alg:none, wrong-key, expired, tampered-payload attacks.
- Fail-soft startup: init failure never brings the PEM data plane down.
- ExecuteScript only; mutations return UNIMPLEMENTED.

direct_query_server_test.cc (new)
- In-process gRPC fixture exercises the full auth metadata flow.
- Auth-negative cases: no token, wrong key, expired, alg:none, truncated, tampered header/payload/sig, wrong aud/iss/scope.
- Execution cases: trivial query, projection, time-range, join, concurrent.

BUILD.bazel (modified)
- config_setting(:direct_query_disabled) for compile-time kill switch.
- New deps on //src/carnot, @boringssl//:crypto, @rapidjson.
- New pl_cc_test(:direct_query_server_test).

pem_manager.{h,cc} (modified)
- MaybeStartDirectQueryServer() builds and starts the gRPC server after broker registration; StopImpl tears it down.
- Flags: --direct_query_enabled (default false), --direct_query_port (default 50305), --direct_query_jwt_signing_key.

shared/manager/manager.cc (modified)
- Empty-key guard at Manager::Init: refuses an empty PL_JWT_SIGNING_KEY rather than crashing mid-stream on the first service call.

DIRECT_QUERY_CONTRACT.md, DIRECT_QUERY_SECURITY.md (new)
- Behavioral spec, auth requirements, threat model, key-flow diagram, tampering test matrix, TLS transport details, disable instructions.

### DIFF
--- a/src/carnot/exec/BUILD.bazel
+++ b/src/carnot/exec/BUILD.bazel
@@ -16,7 +16,10 @@
 
 load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test", "pl_cc_test_library")
 
-package(default_visibility = ["//src/carnot:__subpackages__"])
+package(default_visibility = [
+    "//src/carnot:__subpackages__",
+    "//src/vizier/services/agent:__subpackages__",
+])
 
 pl_cc_library(
     name = "cc_library",
--- a/src/vizier/services/agent/pem/BUILD.bazel
+++ b/src/vizier/services/agent/pem/BUILD.bazel
@@ -20,6 +20,18 @@ load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test
 
 package(default_visibility = ["//src/vizier:__subpackages__"])
 
+# Compile-time kill switch for the direct-query endpoint.
+# Operators who do not want the feature available even as a disabled-by-default
+# option build with `bazel build … --//src/vizier/services/agent/pem:direct_query=disabled`
+# (or `--define=PX_PEM_DIRECT_QUERY=disabled`). This propagates
+# `-DPX_PEM_DIRECT_QUERY_DISABLED` into cc_library; the entire feature body
+# in direct_query_server.cc and pem_manager.cc is `#ifndef`'d out and
+# stub methods return UNIMPLEMENTED. See DIRECT_QUERY_SECURITY.md.
+config_setting(
+    name = "direct_query_disabled",
+    define_values = {"PX_PEM_DIRECT_QUERY": "disabled"},
+)
+
 pl_cc_library(
     name = "cc_library",
     srcs = glob(
@@ -30,7 +42,27 @@ pl_cc_library(
         ],
     ),
     hdrs = glob(["*.h"]),
+    defines = select({
+        ":direct_query_disabled": ["PX_PEM_DIRECT_QUERY_DISABLED"],
+        "//conditions:default": [],
+    }),
     deps = [
+        # direct-query server deps (direct_query_server.{h,cc}):
+        "//src/api/proto/vizierpb:vizier_pl_cc_proto",
+        "//src/carnot",
+        "//src/carnot:cc_library",
+        "//src/carnot/carnotpb:carnot_pl_cc_proto",
+        "//src/carnot/exec:cc_library",
+        "//src/carnot/funcs:cc_library",
+        "//src/carnot/planner/compiler:cc_library",
+        "//src/carnot/planpb:plan_pl_cc_proto",
+        "//src/carnot/udf:cc_library",
+        # HS256 verify uses BoringSSL HMAC directly and parses claims via rapidjson.
+        "@boringssl//:crypto",
+        "@com_github_grpc_grpc//:grpc++",
+        "@com_github_rlyeh_sole//:sole",
+        "@com_github_tencent_rapidjson//:rapidjson",
+        # existing PEM deps:
         "//src/carnot/planner/dynamic_tracing/ir/logicalpb:logical_pl_cc_proto",
         "//src/integrations/grpc_clocksync:cc_library",
         "//src/shared/tracepoint_translation:cc_library",
@@ -48,6 +80,26 @@ pl_cc_test(
         "//src/common/testing/event:cc_library",
         "//src/stirling/testing:stirling_mock",
         "//src/vizier/services/agent/shared/manager:test_utils",
+    ],
+)
+
+# TDD contract for the PEM direct-query endpoint.
+# See DIRECT_QUERY_CONTRACT.md for the behavioral spec.
+pl_cc_test(
+    name = "direct_query_server_test",
+    srcs = ["direct_query_server_test.cc"],
+    deps = [
+        ":cc_library",
+        "//src/api/proto/vizierpb:vizier_pl_cc_proto",
+        "//src/carnot",
+        "//src/carnot:cc_library",
+        "//src/carnot/exec:cc_library",
+        "//src/carnot/funcs:cc_library",
+        "//src/carnot/udf:cc_library",
+        "//src/table_store:cc_library",
+        "@com_github_arun11299_cpp_jwt//:cpp_jwt",
+        "@com_github_grpc_grpc//:grpc++",
+        "@com_github_rlyeh_sole//:sole",
     ],
 )
 
--- a/src/vizier/services/agent/pem/BUILD.bazel
+++ b/src/vizier/services/agent/pem/BUILD.bazel
@@ -14,22 +14,34 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 load("@io_bazel_rules_docker//cc:image.bzl", "cc_image")
 load("@io_bazel_rules_docker//container:container.bzl", "container_image")
 load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test")
 
 package(default_visibility = ["//src/vizier:__subpackages__"])
 
-# Compile-time kill switch for the direct-query endpoint.
-# Operators who do not want the feature available even as a disabled-by-default
-# option build with `bazel build … --//src/vizier/services/agent/pem:direct_query=disabled`
-# (or `--define=PX_PEM_DIRECT_QUERY=disabled`). This propagates
-# `-DPX_PEM_DIRECT_QUERY_DISABLED` into cc_library; the entire feature body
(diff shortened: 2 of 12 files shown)


src/vizier/services/agent/pem/direct_query_server.cc
```diff
@@ -0,0 +1,524 @@
...
+  if (computed_mac.empty()) {
+    return ::grpc::Status(::grpc::StatusCode::UNAUTHENTICATED, "direct-query: HMAC compute failed");
+  }
+  std::string signature;
+  if (!base64UrlDecode(parts[2], &signature)) {
+    return ::grpc::Status(::grpc::StatusCode::UNAUTHENTICATED, "direct-query: bad signature b64");
+  }
+  if (!constantTimeEquals(signature, computed_mac)) {
+    return ::grpc::Status(::grpc::StatusCode::UNAUTHENTICATED, "direct-query: signature mismatch");
+  }
```
ddelnano: Do you know how this constant time equals implementation compares to other library implementations?

src/vizier/services/agent/pem/direct_query_server.cc
```diff
@@ -0,0 +1,524 @@
...
+constexpr char kExpectedAudience[] = "vizier";
+constexpr char kExpectedIssuer[] = "PL";
+// Service tokens carry "service" in the Scopes claim (NOT the subject — sub is
+// the serviceID, e.g. "dx"). See GenerateJWTForService (claims.go) + jwt.go:56.
+constexpr char kServiceScope[] = "service";
+
+// We don't link cpp_jwt's HMAC verifier here because its impl calls
+// BIO_f_base64() which lives in BoringSSL's decrepit/ tree — not exposed as a
+// bazel target on this fork. Instead we parse the JWT envelope manually and
+// HMAC with BoringSSL natively. ~50 lines vs. carrying a boringssl patch.
```
ddelnano: Is this statement true? I think the cpp_jwt's HMAC path will work with our version of boringssl. If not, it seems like that's true of newer versions of cpp_jwt (I checked the latest main).

src/vizier/services/agent/pem/BUILD.bazel
```diff
@@ -30,7 +42,27 @@ pl_cc_library(
         ],
     ),
     hdrs = glob(["*.h"]),
+    defines = select({
+        ":direct_query_disabled": ["PX_PEM_DIRECT_QUERY_DISABLED"],
+        "//conditions:default": [],
+    }),
```
ddelnano: This should use `local_defines` otherwise it will propagate to every transitive dependency.

src/vizier/services/agent/pem/BUILD.bazel
```diff
@@ -20,6 +20,18 @@ load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test
...
+# Operators who do not want the feature available even as a disabled-by-default
+# option build with `bazel build … --//src/vizier/services/agent/pem:direct_query=disabled`
+# (or `--define=PX_PEM_DIRECT_QUERY=disabled`). This propagates
+# `-DPX_PEM_DIRECT_QUERY_DISABLED` into cc_library; the entire feature body
+# in direct_query_server.cc and pem_manager.cc is `#ifndef`'d out and
+# stub methods return UNIMPLEMENTED. See DIRECT_QUERY_SECURITY.md.
+config_setting(
+    name = "direct_query_disabled",
+    define_values = {"PX_PEM_DIRECT_QUERY": "disabled"},
+)
```
ddelnano, line 45: Did you try the `bazel build … --//src/vizier/services/agent/pem:direct_query=disabled` build and verify it works? I believe you need a `bool_flag` associated with this. I think it would be better as a bool_flag so it can work with `--//src/vizier/services/agent/pem:direct_query=false/true`

src/vizier/services/agent/pem/direct_query_server.cc
```diff
@@ -0,0 +1,524 @@
...
+// trivially differ but we still walk the shorter to keep timing predictable
+// across malformed lengths.
+bool constantTimeEquals(absl::string_view a, absl::string_view b) {
+  if (a.size() != b.size()) return false;
+  uint8_t acc = 0;
+  for (size_t i = 0; i < a.size(); ++i) {
+    acc |= static_cast<uint8_t>(a[i] ^ b[i]);
+  }
+  return acc == 0;
+}
```
ddelnano, line 117: I think we want to avoid handling constant time crypto ourselves. If we must do it, I think using [CRYPTO_memcp](https://github.com/google/boringssl/blob/179af5abdbf5aeba9c71a082126639b8d872af0a/include/openssl/mem.h#L68-L73) would be the way to go.

src/vizier/services/agent/pem/direct_query_server.cc
```diff
@@ -0,0 +1,524 @@
...
+// Service tokens carry "service" in the Scopes claim (NOT the subject — sub is
+// the serviceID, e.g. "dx"). See GenerateJWTForService (claims.go) + jwt.go:56.
+constexpr char kServiceScope[] = "service";
+
+// We don't link cpp_jwt's HMAC verifier here because its impl calls
+// BIO_f_base64() which lives in BoringSSL's decrepit/ tree — not exposed as a
+// bazel target on this fork. Instead we parse the JWT envelope manually and
+// HMAC with BoringSSL natively. ~50 lines vs. carrying a boringssl patch.
+
+// stripBearerPrefix returns the token slice after a case-insensitive "Bearer "
```
ddelnano, line 93: I think all of the jwt handling code should go in a new location (`/src/shared/services/jwt`). Since this has vizier specific information, I think it makes more sense there then a more general `src/common` or similar directory.

This new directory can hold something like this:
```
  // src/shared/services/jwt/service_token.h
  Status VerifyServiceJWT(std::string_view token, std::string_view signing_key,
                          const ServiceJWTOptions& opts);   // aud, iss, required scopes, leeway
```

src/vizier/services/agent/pem/direct_query_server.cc
```diff
@@ -0,0 +1,524 @@
...
+// ExecuteScriptResponse per GRPC_SINK_OPERATOR sink. The client uses these to
+// learn output table names and column types before the data chunks arrive.
+// Mirrors standalone_pem/vizier_server.h:132-173.
+void emitSchemaResponses(const ::px::carnot::planpb::Plan& plan, const std::string& query_id,
+                         ::grpc::ServerWriter<::px::api::vizierpb::ExecuteScriptResponse>* writer) {
+  for (const auto& f : plan.nodes()) {
+    for (const auto& n : f.nodes()) {
+      if (n.op().op_type() != ::px::carnot::planpb::OperatorType::GRPC_SINK_OPERATOR) continue;
+      const auto& sink = n.op().grpc_sink_op();
+      if (!sink.has_output_table()) continue;
```
ddelnano, line 324: Can we reuse all of the code in src/experimental/standalone_pem/sink_server.h and src/experimental/standalone_pem/vizier_server.h ? If the existing standalone pem implementation has slight differences, we should resolve those or adapt it to be general enough for both.

src/carnot/exec/BUILD.bazel
```diff
@@ -16,7 +16,10 @@
 
 load("//bazel:pl_build_system.bzl", "pl_cc_binary", "pl_cc_library", "pl_cc_test", "pl_cc_test_library")
 
-package(default_visibility = ["//src/carnot:__subpackages__"])
+package(default_visibility = [
+    "//src/carnot:__subpackages__",
+    "//src/vizier/services/agent:__subpackages__",
+])
```
ddelnano, line 22: I think it'd be best to avoid making this package broadly open to all `//src/vizier/services/agent` subpackages. I need to think about how to scope this more tightly, but we do have a similar pattern in [here](https://github.com/pixie-io/pixie/blob/830ff2ad3d5b6d4ff5a975ac5b21d43d42ce4bef/src/carnot/BUILD.bazel#L24-L32).

I think we need to consider this for the other package visibility changes as well. I'll have to think about this more before I have any proposals on how to structure it better.

src/vizier/services/agent/pem/direct_query_server.cc
```diff
@@ -0,0 +1,524 @@
...
+constexpr char kServiceScope[] = "service";
+
+// We don't link cpp_jwt's HMAC verifier here because its impl calls
+// BIO_f_base64() which lives in BoringSSL's decrepit/ tree — not exposed as a
+// bazel target on this fork. Instead we parse the JWT envelope manually and
+// HMAC with BoringSSL natively. ~50 lines vs. carrying a boringssl patch.
+
+// stripBearerPrefix returns the token slice after a case-insensitive "Bearer "
+// prefix, or an empty string if the prefix is missing. gRPC normalises metadata
+// keys to lowercase but does NOT touch values; manager.cc:440 mints with a
```
ddelnano, line 95: In general, I think we should avoid `file_name.cc:{line_num}` references. These are guaranteed to break when someone refactors the referenced code.

I think github permalinks are the way to go for these reference. Please analyze all of them in this PR.

src/vizier/services/agent/pem/direct_query_server.cc
```diff
@@ -0,0 +1,524 @@
...
+    } else if ((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') ||
+               c == '+' || c == '/') {
+      std_b64.push_back(c);
+    } else {
+      return false;
+    }
+  }
+  while (std_b64.size() % 4 != 0) std_b64.push_back('=');
+  return absl::Base64Unescape(std_b64, out);
+}
```
ddelnano, line 125: The jwt library we use has a header only implementation of [this](https://github.com/arun11299/cpp-jwt/blob/365dcc9b843ef04ffd398ef5adc62742dc4c6d27/include/jwt/base64.hpp). I'm hoping we can defer to it for the entire crypto part (which would include the base64 part), but if we need to handle the crypto we should use that header only library.

